PR #27, opened 2024-05-06
Enter prerelease mode

oskarhane:
We're now releasing `0.1.X` versions that might (will) have breaking changes, which is not desired from a developers point of view.
This PR takes the package into prerelease mode (`0.2.0a0`), to signal that breaking changes might happen at any time.


willtai (approved):
LGTM 🌳 Don't forget to edit the PR description

PR #28, opened 2024-05-06
Add debug logging in retrievers

oskarhane:
We do not set any basic config on the loggers, but leave that for application developers to keep the log format consistent with their existing ones.

This can be tested by logging `DEBUG` when running tests:

```bash
poetry run pytest tests/unit --log-cli-level=DEBUG
```

### DIFF
--- a/src/neo4j_genai/retrievers/hybrid.py
+++ b/src/neo4j_genai/retrievers/hybrid.py
@@ -86,8 +86,8 @@ class HybridRetriever(Retriever):
 
         search_query = get_search_query(SearchType.HYBRID, self.return_properties)
 
-        logger.debug("HybridRetriever Cypher parameters", parameters)
-        logger.debug("HybridRetriever Cypher query", search_query)
+        logger.debug("HybridRetriever Cypher parameters %s", parameters)
+        logger.debug("HybridRetriever Cypher query %s", search_query)
 
         records, _, _ = self.driver.execute_query(search_query, parameters)
         return records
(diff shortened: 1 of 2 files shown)


src/neo4j_genai/retrievers/hybrid.py
```diff
@@ -157,5 +164,11 @@ def search(
         search_query = get_search_query(
             SearchType.HYBRID, retrieval_query=self.retrieval_query
         )
+
+        logger.debug(
+            f"HybridCypherRetriever Cypher parameters: {json.dumps(parameters)}"
```
stellasia, line 169: Do we have to dump the parameters here? I'm asking because it may contain non JSON serializable values, like UUIDs (this particular example would fail at the query level later on, but it makes more sense to me than failing because of a log)

oskarhane: Interesting point. When debugging, you sure want to have the information needed to debug the issue.
How would you do this?

stellasia: I would pass the parameters dict to the logger, without formatting. The logger will take care of "stringifying" each item.

oskarhane: Thanks, updated

src/neo4j_genai/retrievers/hybrid.py
```diff
@@ -83,6 +86,9 @@ def search(
 
         search_query = get_search_query(SearchType.HYBRID, self.return_properties)
 
+        logger.debug("HybridRetriever Cypher parameters", parameters)
```
stellasia, line 89: Another point: we need to use placeholder in the first string for the other parameters to be used (sorry missed that during first read). I.e.:
``` 
logger.debug("HybridRetriever Cypher parameters %s", parameters)
```

willtai:
Currently the codebase is still small so this looks good. I think in time we can centralize logging more as the codebase grows larger

Non-blocking comment: I would prefer we use f-strings wherever possible for readability.

PR #29, opened 2024-05-06
Support for filters in Vector and VectorCypher retrievers

stellasia:
Add support for pre-filtering for the Vector and VectorCypher retrievers


src/neo4j_genai/neo4j_queries.py
```diff
@@ -12,51 +12,169 @@
...
+    elif search_type == SearchType.VECTOR:
+        query, params = _get_vector_query(
+            filters, node_label, embedding_node_property, embedding_dimension
+        )
+    else:
+        raise ValueError(f"Search type is not supported: {search_type}")
+    query_tail = _get_query_tail(
+        retrieval_query, return_properties, fallback_return="RETURN node, score"
+    )
+    return " ".join([query, query_tail]), params
```
willtai, line 158: I think f"{query} {query_tail}" is more readable here

oskarhane: I don't agree, I think it's better as it is :)

willtai: I'd personally do `" ".join([a, b])` if we already have a list to begin with and pass it in, which is why I'd prefer `f"{query} {query_tail}"`. 

However, this is subjective. I'm happy to go either way as long as we keep preferences consistent across the codebase

src/neo4j_genai/retrievers/filters.py
```diff
@@ -0,0 +1,345 @@
...
+        self.params = {}
+
+    def _get_params_name(self, key="param"):
+        """Find parameter name so that param names are unique.
+        This function adds a suffix to the key corresponding to the number
+        of times the key have been used in the query.
+        E.g.
+        node.age >= $param_0 AND node.age <= $param_1
+
+        :param p:
```
willtai, line 157: Wdyt of sticking to Google style docstrings? https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html we currently do this for the other docstrings

src/neo4j_genai/retrievers/filters.py
```diff
@@ -0,0 +1,345 @@
...
+    field: str,
+    native_operator_class: Type[Operator],
+    value: Any,
+    param_store: ParameterStore,
+    node_alias: str,
+) -> str:
+    """Return Cypher for field operator value.
+
+    Args:
+        field: the name of the field being filtered
```
willtai, line 185:
```suggestion
        field: The name of the field being filtered
```

Don't forget to capitalize the first letter of the descriptions

src/neo4j_genai/retrievers/filters.py
```diff
@@ -0,0 +1,345 @@
...
+    query = cypher_operator.join(
+        [
+            f"({ _construct_metadata_filter(el, param_store, node_alias)})"
+            for el in value
+        ]
+    )
+    return query
+
+
+def construct_metadata_filter(
```
willtai, line 329: Minor but perhaps rename to `get_metadata_filter` or something else for this function? Given there is a private function with the same name

src/neo4j_genai/retrievers/filters.py
```diff
@@ -0,0 +1,345 @@
...
+        node_alias: name of the node being filtered in the Cypher query
+
+    Returns:
+        str: the Cypher WHERE clause
+
+    NB: the param_store argument is mutable, it will be updated in this function
+    """
+
+    if not isinstance(filter, dict):
+        raise ValueError()
```
willtai, line 296: Perhaps add an error message here

src/neo4j_genai/neo4j_queries.py
stellasia: I do not have a strong opinion on this one, I kept the join that was there in the previous version. Now, f-string seems to be slightly faster, and I do not think we plan to add more items to this list (post-filtering can be performed in the retrieval query), so we can move to f-string.

stellasia: Actually, updated to f-string for consistency with other parts of the code (sorry Oskar :) )

src/neo4j_genai/filters.py
```diff
@@ -0,0 +1,349 @@
...
+        consistent with the operator (e.g. LIKE operator only works with string values)
+    """
+
+    CYPHER_OPERATOR = None
+
+    def __init__(self, node_alias=DEFAULT_NODE_ALIAS):
+        self.node_alias = node_alias
+
+    def lhs(self, field):
+        return f"{self.node_alias}.`{field}`"
```
oskarhane, line 38: 1. Prefix any field name backtick with an backtick
2. If you find non alphabetic characters in the field name, wrap the whole field name with backticks

```
name -> name
1name -> `1name`
na`me -> `na``me`
```

Example of valid Cypher:

```
MATCH (n:`Hell``o`) WHERE n.`prop``erty` RETURN n LIMIT 25
```

More info: https://neo4j.com/docs/cypher-manual/current/syntax/naming/#symbolic-names-escaping-rules

src/neo4j_genai/retrievers/base.py
```diff
@@ -57,3 +57,19 @@ def _verify_version(self) -> None:
...
     def search(self, *args, **kwargs) -> Any:
         pass
+
+    def _fetch_index_infos(self):
+        """Fetch the node label and embedding property from the index definition"""
+        query = """SHOW VECTOR INDEXES
+YIELD name, labelsOrTypes, properties, options
+WHERE name = $index_name
+RETURN labelsOrTypes as labels, properties, options.indexConfig.`vector.dimensions` as dimensions
+        """
```
oskarhane, line 67: Please format this query like we do with the others

oskarhane (commented):
This is great, so good to see this getting in.
I've got some comments initially, and then I'll test it out locally as well.

src/neo4j_genai/filters.py
```diff
@@ -0,0 +1,349 @@
...
+        of times the key have been used in the query.
+        E.g.
+        node.age >= $param_0 AND node.age <= $param_1
+
+        Args:
+            key (str): The prefix for the parameter name
+        Returns:
+            The full unique parameter name
+        """
+        # key = slugify(key.replace(".", "_"), separator="_")
```
willtai, line 160: Do we want to use slugify?

src/neo4j_genai/filters.py
```diff
@@ -0,0 +1,349 @@
...
+    """
+    Store parameters for a given query.
+    Determine the parameter name depending on a parameter counter
+    """
+
+    def __init__(self):
+        self._counter = Counter()
+        self.params = {}
+
+    def _get_params_name(self, key="param"):
```
willtai, line 148: Does `key` need a default?

src/neo4j_genai/filters.py
```diff
@@ -0,0 +1,349 @@
...
+        # key = slugify(key.replace(".", "_"), separator="_")
+        param_name = f"{key}_{self._counter[key]}"
+        self._counter[key] += 1
+        return param_name
+
+    def add(self, key, value):
+        """This function adds a new parameter to the param dict.
+        It returns the name of the parameter to be used as a placeholder
+        in the cypher query, e.g. $param_0"""
+        param_name = self._get_params_name()
```
willtai, line 169: Is this meant to be 

```python
param_name = self._get_params_name()
```
?

src/neo4j_genai/filters.py
stellasia: Interesting, thanks! Wdyt about always wrapping the field name in backticks, whether it contains special characters or not? In your example it means that 
``` 
name -> `name`
```

src/neo4j_genai/filters.py
stellasia: I used it in my initial implementation, where the parameter name was a slugified version of the field name. But now I'm not sure it's worth it? Right now all parameters are called `param_<index>`.

src/neo4j_genai/filters.py
stellasia: You mean we are missing the `key` argument? If we decide to use it yes, otherwise it should probably be deleted from everywhere.

src/neo4j_genai/filters.py
```diff
@@ -0,0 +1,349 @@
...
+        self.params[param_name] = value
+        return param_name
```
willtai, line 171: Should returning the `param_name` be in its own public method?

src/neo4j_genai/filters.py
willtai: Hmm I ask because the argument `key` is not used here in `add()`. When do we need to use `key` in general?

src/neo4j_genai/filters.py
willtai: If it's always going to be `param_<index>` I guess we don't `key` at all

stellasia: Yes, removed this argument

src/neo4j_genai/filters.py
stellasia: Argument removed as discussed above

src/neo4j_genai/filters.py
stellasia: TBH I don't see how, the parameter name is dependent on the order in which the parameters are added and since we have the increment each time the `add` method is called, how would you write a specific public method?

src/neo4j_genai/filters.py
oskarhane: I'd see that we'd rather not do that. It makes the queries more noisy when debugging.

willtai (changes requested):
The `ParameterStore` is not thread-safe. Is this something we should implement? I think it depends on whether users will run this code in multi-threaded environments.

Following [this post](https://stackoverflow.com/questions/35088139/how-to-make-a-thread-safe-global-counter-in-python), I think we can change it to something like:

```python
import threading
from collections import Counter

class ParameterStore:
    def __init__(self):
        self._counter = Counter()
        self.params = {}
        self.lock = threading.Lock()

    def _get_params_name(self):
        with self.lock:  # Lock around accessing and incrementing the counter
            key = "param"
            param_name = f"{key}_{self._counter[key]}"
            self._counter[key] += 1
            return param_name

    def add(self, value):
        with self.lock:  # Lock around the entire add operation
            param_name = self._get_params_name()
            self.params[param_name] = value
            return param_name

```

It could be interesting to test whether there is a change in performance

src/neo4j_genai/filters.py
stellasia: I understand your point, and at the same time I'm a bit scared that we miss some special characters and produce invalid query "just" to make it nice. Let me check if we can find something really robust.

stellasia: See here, if you think the test cases are covering enough use cases: https://github.com/neo4j/neo4j-genai-python/pull/29/commits/2b89aff3865e3987c89bf5de0947288559c21c70#diff-00b9f6996b286b97ed1b966f8596505071f2876cac256d97e31bb0de4449b198R53-R69

willtai (approved):

src/neo4j_genai/filters.py
oskarhane: Yeah, I see your concern. But if anyone should be able to do the right thing it should be us :) 
This should give us confidence: https://github.com/neo4j/cypher-builder/blob/main/src/utils/escape.ts#L54-L63

willtai (approved):
Great work! On thread-safety, we concluded that the usage of`ParameterStore` is fine as it is instantiated in `get_metadata_filter()`. We have no plans on using multithreading inside `get_metadata_filter()` so this shouldn't block this PR

src/neo4j_genai/filters.py
```diff
@@ -0,0 +1,362 @@
...
+            for el in value
+        ]
+    )
+    return query
+
+
+def get_metadata_filter(
+    filter: dict[str, Any], node_alias: str = DEFAULT_NODE_ALIAS
+) -> tuple[str, dict]:
+    """Construct the cypher filter snippet based on a filter dict
```
willtai, line 348: Suggestion: we can add a comment here saying this function is not thread-safe

src/neo4j_genai/filters.py
stellasia: Fair enough, used the same regex to test the field validity.

oskarhane (approved):
Awesome 🌊

PR #30, opened 2024-05-09
Handle failure cases for index handling methods

willtai:
# Summary
- Adds try/catch for `create_vector_index`, `create_fulltext_index`, and `drop_index_if_exists`
- Rename `drop_index` to `drop_index_if_exists` as this is more true to its functionality
- Adds tests to assert that `neo4j.exceptions.ClientError` is raised successfully
- Adds logging for index handling methods
- Rename imports of `neo4j.Record` and `neo4j.Driver` to be more explicit

### DIFF
--- a/src/neo4j_genai/indexes.py
+++ b/src/neo4j_genai/indexes.py
@@ -13,13 +13,17 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from neo4j import Driver
+import neo4j
 from pydantic import ValidationError
 from .types import VectorIndexModel, FulltextIndexModel
+import logging
+
+
+logger = logging.getLogger(__name__)
 
 
 def create_vector_index(
-    driver: Driver,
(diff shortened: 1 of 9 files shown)


src/neo4j_genai/indexes.py
```diff
@@ -97,26 +115,39 @@ def create_fulltext_index(
...
-    driver.execute_query(query, parameters)
+    try:
+        query = "DROP INDEX $name IF EXISTS"
+        parameters = {
+            "name": name,
+        }
+        logger.info(f"Dropping index named '{name}'")
+        driver.execute_query(query, parameters)
+    except neo4j.exceptions.ClientError as e:
+        logger.error(f"Neo4j fulltext index creation failed {e}")
```
jonbesga, line 152: It should be `Neo4j full-text index drop failed`

jonbesga (approved):
One minor comment on the logging of one error

PR #31, opened 2024-05-13
Add new types for validating inputs to retrievers init

willtai:
# Summary
This PR introduces data validation when instantiating the retriever classes using Pydantic, helping make the use of these retrievers more reliable and secure.

# Implementation Details
- Added 4 Pydantic data models for the 4 different retrievers.
- Added try/catch for data validation. When there is a validation error, `logger` logs a message and a `ValueError` is raised
- Added unit tests for incorrect data validation
- A "Neo4j driver model" and an "embedder model" is passed to the retriever data models as it's tricky to pass a custom class type to another Pydantic data model, [see the Pydantic docs](https://docs.pydantic.dev/latest/concepts/types/#handling-custom-generic-classes).

### DIFF
--- a/src/neo4j_genai/types.py
+++ b/src/neo4j_genai/types.py
@@ -14,7 +14,13 @@
 #  limitations under the License.
 from enum import Enum
 from typing import Any, Literal, Optional
-from pydantic import BaseModel, PositiveInt, model_validator, field_validator
+from pydantic import (
+    BaseModel,
+    PositiveInt,
+    model_validator,
+    field_validator,
+    ConfigDict,
+)
 import neo4j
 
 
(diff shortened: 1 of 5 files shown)


src/neo4j_genai/types.py
```diff
@@ -93,3 +99,59 @@ class SearchType(str, Enum):
...
+        return value
+
+
+class Neo4jDriverModel(BaseModel):
+    driver: neo4j.Driver
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    @field_validator("driver")
+    def check_driver(cls, value):
+        if not isinstance(value, neo4j.Driver):
```
stellasia, line 125: Isn't this validation covered by the type declaration above?

src/neo4j_genai/types.py
```diff
@@ -93,3 +99,59 @@ class SearchType(str, Enum):
 
     VECTOR = "vector"
     HYBRID = "hybrid"
+
+
+class EmbedderModel(BaseModel):
+    embedder: Optional[Any]
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
stellasia, line 106: I probably missed something but where are these `model_config` used?

tests/unit/retrievers/test_vector.py
```diff
@@ -28,6 +29,26 @@ def test_vector_retriever_initialization(driver):
         mock_verify.assert_called_once()
 
 
+def test_vector_retriever_bad_data_validation(driver):
+    with pytest.raises(ValueError):
+        with patch(
+            "neo4j_genai.retrievers.base.Retriever._verify_version"
+        ) as mock_verify:
+            VectorRetriever(driver=driver, index_name=42)
+            mock_verify.assert_called_once()
```
stellasia, line 38: Are we keeping this check for all tests on purpose? Also, do we want to verify version if the provided data is not valid? For instance if the driver is not a valid neo4j driver, we won't even be able to try, right?
Last but not least, I don't understand how this got called looking at the code actually, isn't the version check performed in the super method after the input validation?

src/neo4j_genai/types.py
willtai: Good question because it's subtle. Without `model_config` we will get errors when trying to pass arbitrary types like `Embedder`. I chose to pass this into the `EmbedderModel` instead of every retriever data model for better encapsulation

[In Pydantic V2 you can do this as seen here](https://docs.pydantic.dev/2.7/api/config/#pydantic.config.ConfigDict.arbitrary_types_allowed)

src/neo4j_genai/types.py
willtai: Replied in the question regarding `model_config`

tests/unit/retrievers/test_vector.py
willtai: Good catch!

src/neo4j_genai/types.py
stellasia: Got it, thanks

src/neo4j_genai/types.py
stellasia: I understand it in the case of embedder, where we can have different types (maybe not ours). But here we will always have a neo4j driver right? Or did I miss something?

willtai: Like `Embedder`, `__get_pydantic_core_schema__` is not implemented on `neo4j.Driver` as well which would result in the same error. So either we are more lenient on this type here or we try and implement the same on the python driver library. I think it is easier for now to do this here

stellasia: Oh ok, interesting! Maybe worth opening an issue for the driver to get it implemented, wdyt? 
But anyway in the meantime we can do it this way, LGTM.

stellasia (approved):

PR #34, opened 2024-05-17
Fixed tests to raise correct error for retriever instantiation data validation

willtai:
# Description
Previously the tests for testing incorrect data validation were wrong as they raised the wrong `ValueError`. They raised errors because `_verify_version` failed. This PR ensures the tests are set up correctly using `@patch` and additional assertions are added to the tests.

# Implementation Details
- Setup for `_verify_version` added for data validation tests. These would have failed and raised a `ValueError`, so more assertions are added to ensure the correct input argument fails the data validation
- Data validation test names made more descriptive
- `embedder` fixture added for unit tests


oskarhane (approved):
I think this makes sense 👍

PR #35, opened 2024-05-20
Add support for external vector integration via Weaviate

oskarhane:
## What is it?

With `WeaviateNeo4jRetriever` developers can use a remote Weaviate database to perform vector similarity search which then jumps over to Neo4j to continue the query to get tog et all the benefits a graph database gives in RAG systems.

```mermaid
sequenceDiagram
    participant Caller
    participant ER as WeaviateNeo4jRetriever
    participant ES as Weaviate
    Caller->>ER: Q = What do zebras eat?
    ER->>ES: Q or embedding of Q
    ES-->>ER: list[tuple[str, float]]
    Note over ER,ES: References to data in Neo4j
    ER->>Neo4j: Cypher, list[tuple[str, float]]
    Neo4j-->>ER: neo4j.Result
    ER-->>Caller: list[neo4j.Record]
```

Check out the README in the `src/neo4j_genai/retrievers/external/weaviate/` folder on how to test this out locally.

The idea:

```python
retriever = WeaviateNeo4jRetriever(
    driver=neo4j_driver,
    client=w_client,
    collection="Question",
    id_property_external="neo4j_id",
    id_property_neo4j="id",
)

res = retriever.search(query_text="What do zebras eat?", top_k=2)
```

<details>
  <summary><h3>Open fold for design considerations and info</h3></summary>

##### Original design
The original design was to re-use our `VectorRetreiver` and `VectorCypherRetriever` and inject an "index provider" that could be either Neo4j or external ones. By doing that we would limit the number of retrievers and the developers would have a smaller surface.

Something like this:

```mermaid
flowchart LR
    Input-->search[".search()"]
    EI[External Vector Index]
    IP[Similarity search]
    subgraph retriever [VectorRetriever]
        search-->Check{User external Index?}
        subgraph v1 [ExternalVectorIndexProvider]
        IP
        end
        subgraph g1 [GraphQuery]
        GQ[Graph Query]-- Cypher query -->Neo4j
        end
    end
    v1 -->GQ
    Check-- Yes --> IP
    Check -- No --> GQ
    IP -- query --> EI
    EI -- result --> IP
    Neo4j --> Output
```

But that had multiple downsides: The interface for external index providers will become bloated over time because providers have different feature sets and subtle differences in behavior. It also introduces an extra layer that makes it harder to get started. It was also a bit awkward to create a `VectorRetriever` but the actual search in the remote store could be a hybrid search and similar things.

##### Updated
Instead we made a change where each external retriever is a first class citizen and all operations happens in each external retriever.

This should make the integrations easier to maintain in the long term. We can also support all specific features for each vendor without have to change interfaces or make updates anywhere else.

```mermaid
flowchart LR
    Input-->search[".search()"]
    EI[External Vector Index]
    IP[Similarity search]
    subgraph retriever [WeaviateNeo4jRetriever]
        search-->IP
        IP-->GQ
        GQ[Graph Query]-- Cypher query -->Neo4j
    end
    IP -- query --> EI
    EI -- result --> IP
    Neo4j --> Output
```

One downside of the new approach is that we currently separate simple retrievers from retrievers + retrieval query by name (`VectorRetriever` vs. `VectorCypherRetreiver`) and I don't think we can keep that pattern with the combined external retrievers. So in this PR the arguments to the constructor includes `return_properties` and `retrieval_query` but the former will not be used if the latter is present.
</details>

For the E2E tests, we're using a local embedding model `all-MiniLM-L6-v2`.

## Todo

- [x] E2E tests
- [x] Add filter support


src/neo4j_genai/retrievers/external/weaviate/examples/text_search_remote_embedder.py
```diff
@@ -0,0 +1,40 @@
...
+from neo4j import GraphDatabase
+import weaviate
+
+NEO4J_URL = "neo4j://localhost:7687"
+NEO4J_AUTH = ("neo4j", "password")
+
+
+def main():
+    with GraphDatabase.driver(NEO4J_URL, auth=NEO4J_AUTH) as neo4j_driver:
+        with weaviate.connect_to_local() as w_client:
```
stellasia, line 26: The use of the remote embedder comes from a Weaviate configuration? If yes maybe add a comment here, might be obvious for users who come from weaviate but not for others.

src/neo4j_genai/retrievers/external/weaviate/weaviate.py
```diff
@@ -0,0 +1,148 @@
...
+        If no embedder is provided, then it will assume that the vectorizer is used in Weaviate.
+
+        See the following documentation for more details:
+        - [Query a vector index](https://neo4j.com/docs/cypher-manual/current/indexes-for-vector-search/#indexes-vector-query)
+        - [db.index.vector.queryNodes()](https://neo4j.com/docs/operations-manual/5/reference/procedures/#procedure_db_index_vector_queryNodes)
+        - [db.index.fulltext.queryNodes()](https://neo4j.com/docs/operations-manual/5/reference/procedures/#procedure_db_index_fulltext_querynodes)
+        Args:
+            query_text (str): The text to get the closest neighbors of.
+            query_vector (Optional[list[float]], optional): The vector embeddings to get the closest neighbors of. Defaults to None.
+            top_k (int, optional): The number of neighbors to return. Defaults to 5.
```
stellasia, line 73: Missing `weaviate_filters` arg :)

oskarhane: https://github.com/neo4j/neo4j-genai-python/pull/35/commits/c8b3aab5eb82a476425a7127903940320b860ab4

src/neo4j_genai/retrievers/external/weaviate/examples/text_search_remote_embedder.py
oskarhane: Yeah, good one. https://github.com/neo4j/neo4j-genai-python/pull/35/commits/d43f2210bd06da11fe26b01970bb8afcb284e1c5

src/neo4j_genai/retrievers/external/weaviate/examples/text_search_local_embedder.py
```diff
@@ -0,0 +1,43 @@
...
+import weaviate
+
+NEO4J_URL = "neo4j://localhost:7687"
+NEO4J_AUTH = ("neo4j", "password")
+
+
+def main():
+    with GraphDatabase.driver(NEO4J_URL, auth=NEO4J_AUTH) as neo4j_driver:
+        with weaviate.connect_to_local() as w_client:
+            embedder = HuggingFaceEmbeddings(model_name="all-MiniLM-L6-v2")
```
willtai, line 28: I see starting the container with `all-MiniLM-L6-v2` takes the most time in the E2E test. Maybe consider using `paraphrase-albert-small-v2` model instead as this is the smallest sentence-transformers model?

willtai: Based on https://www.sbert.net/docs/pretrained_models.html

oskarhane: I think we're good with the current one, if you're ok with that. Locally it's only slow the first time.

src/neo4j_genai/retrievers/external/weaviate/examples/vector_search.py
```diff
@@ -0,0 +1,46 @@
...
+#  limitations under the License.
+
+from neo4j_genai.retrievers.external.weaviate import WeaviateNeo4jRetriever
+from neo4j import GraphDatabase
+import weaviate
+import sys
+import os
+
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(f"{BASE_DIR}/../../../../../../tests/e2e/weaviate_e2e")
```
willtai, line 23: Why do we need to add this to the path?

oskarhane: So I can import the function. Maybe there are other ways?

willtai: I'm thinking if it make more sense to keep the example outside of `src/` and in `examples/` instead? Imo it should live outside of the source code

Is there a reason the vector needs to sit inside `tests/`? If not we can either duplicate it in both `examples/` and `tests/`. If you think we should still import this function from `tests/` then you can make it cleaner like this:

```python
root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
sys.path.insert(0, root_dir)
```

tests/e2e/weaviate_e2e/utils.py
```diff
@@ -0,0 +1,388 @@
+def get_query_vector():
```
willtai, line 1: I'm thinking whether this should be a constant instead as we don't expect to mutate this vector

src/neo4j_genai/retrievers/external/weaviate/examples/vector_search.py
oskarhane: The vector is used in the tests, that's why it's there. And I figured we'd reuse it when demonstrating vector search.
Happy to duplicate it though. And will move the examples to the examples folder.

Generally I'm not a fan of grouping files by type but rather by feature (I would move all tests, examples etc right next to the implementation) but I understand that's not very common in the python world.
Long term maintainability would suffer when grouping by type. But That shouldn't block this PR :)

tests/e2e/weaviate_e2e/utils.py
oskarhane: Yeah, makes sense.

src/neo4j_genai/retrievers/external/weaviate/examples/vector_search.py
oskarhane: Moved around a bit and duplicated the embedding: https://github.com/neo4j/neo4j-genai-python/pull/35/commits/140e610e0875401b145e0f2a58608060ef9a223b

stellasia (approved):

willtai (approved):
LGTM!

PR #36, opened 2024-05-21
E2E test for the DB schema functionality

leila-messallem:
E2E tests for the DB schema functionality.

### DIFF
--- a/.github/workflows/pr-e2e-tests.yaml
+++ b/.github/workflows/pr-e2e-tests.yaml
@@ -39,7 +39,7 @@ jobs:
         env:
           NEO4J_AUTH: neo4j/password
           NEO4J_ACCEPT_LICENSE_AGREEMENT: 'eval'
-          NEO4J_PLUGINS: fromJson('["apoc"]')
+          NEO4J_PLUGINS: "[\"apoc\"]"
         ports:
           - 7687:7687
           - 7474:7474
--- a/.github/workflows/pr-e2e-tests.yaml
+++ b/.github/workflows/pr-e2e-tests.yaml
@@ -39,6 +39,7 @@ jobs:
         env:
           NEO4J_AUTH: neo4j/password
           NEO4J_ACCEPT_LICENSE_AGREEMENT: 'eval'
+          NEO4J_PLUGINS: fromJson('["apoc"]')
         ports:
           - 7687:7687
           - 7474:7474
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -111,3 +111,13 @@ def setup_schema(driver):
         MERGE (la)-[:REL_TYPE {rel_prop: 'abc'}]-> (lc)
         """
     )
+
+
+@pytest.fixture(scope="module")
+def setup_excluded_labels(driver):
+    # Delete all nodes in the graph
+    driver.execute_query("MATCH (n) DETACH DELETE n")
+    # Create two labels and a relationship to be excluded
+    driver.execute_query(
+        "CREATE (:_Bloom_Scene_{property_a: 'a'})-[:_Bloom_HAS_SCENE_{property_b: 'b'}]->(:_Bloom_Perspective_)"
+    )
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -95,3 +95,19 @@ def setup_neo4j(driver):
             "authorName": random_str(1536),
         }
         driver.execute_query(insert_query, parameters)
+
+
+@pytest.fixture(scope="module")
+def setup_schema(driver):
+    # Delete all nodes in the graph
+    driver.execute_query("MATCH (n) DETACH DELETE n")
+    # Create two nodes and a relationship
+    driver.execute_query(
+        """
+        CREATE (la:LabelA {property_a: 'a'})
+        CREATE (lb:LabelB)
+        CREATE (lc:LabelC)
+        MERGE (la)-[:REL_TYPE]-> (lb)
+        MERGE (la)-[:REL_TYPE {rel_prop: 'abc'}]-> (lc)
+        """
+    )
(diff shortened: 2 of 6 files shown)


.github/workflows/pr-e2e-tests.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
         env:
           NEO4J_AUTH: neo4j/password
           NEO4J_ACCEPT_LICENSE_AGREEMENT: yes
+          NEO4J_PLUGINS: fromJson('["apoc"]')
```
leila-messallem, line 27: I think we also need to to update `neo4j.conf` to access the procedures:
`dbms.security.procedures.unrestricted=apoc.*`

jonbesga: AFAIK There's no need to do the `fromJson`? The Dockerfile / entrypoint defined in he image take care of everything, including the `dbms.security.procedures.unrestricted=apoc.*`

leila-messallem: Seems like it doesn't, because the we're getting `ProcedureNotFound` error in the test. Or am I missing something else?

jonbesga: What's the procedure being called? Does it work when doing the fromJson?

leila-messallem: Just checked the neo4j 5 community edition image locally, apoc is neither installed or allowed in the config by default, but providing the `NEO4J_PLUGINS` env var installs the plugin as well as adds the line to the neo4j.conf.

So something is not working with the env var in GitHub 🤔

jonbesga: Maybe the version we are using? What are the matrix values?

jonbesga: Are we sure we are loading the right image on GitHub? As in, the matrix variables are being used in the service?

leila-messallem: Yes, according to the logs.

jonbesga: I wonder if when doing Docker run Docker takes care of the conversion but GH doesn't do that

leila-messallem: Or not, I found another syntax that seems to be working! 🎉

leila-messallem: > I wonder if when doing docker run Docker takes care of the conversion but GH doesn't do that.

jonbesga: > Or not, I found another syntax that seems to be working! 🎉 [8855efa](https://github.com/neo4j/neo4j-genai-python/pull/36/commits/8855efa85c067f94d93e43e76aabb428644bb6de)

W-O-W. But basically GH at fault.

tests/e2e/test_schema_e2e.py
```diff
@@ -0,0 +1,104 @@
...
+    NODE_PROPERTIES_QUERY,
+    BASE_ENTITY_LABEL,
+    EXCLUDED_LABELS,
+    EXCLUDED_RELS,
+    REL_PROPERTIES_QUERY,
+    REL_QUERY,
+)
+
+
+def test_cypher_returns_correct_schema(driver):
```
willtai, line 27: I think we can split these into 3 separate unit tests so it's more granular. Wdyt?

leila-messallem: Sure, we could do that. And also move the setup cypher to some kind of decorator.

willtai: You can add and use a fixture like the ones in `conftest.py`

tests/e2e/test_schema_e2e.py
```diff
@@ -0,0 +1,104 @@
...
+    ]
+    assert node_properties == expected_node_properties
+    assert relationships_properties == expected_relationships_properties
+    assert (
+        sorted(relationships, key=lambda x: x["output"]["end"])
+        == expected_relationships
+    )
+
+
+def test_get_schema_filtering_labels(driver):
```
willtai, line 79: I think these can be split to 2 unit tests as well?

tests/e2e/test_schema_e2e.py
leila-messallem: [acbbe02](https://github.com/neo4j/neo4j-genai-python/pull/36/commits/acbbe0208b52bae330738c97b0934f9f01b1279e)

tests/e2e/conftest.py
```diff
@@ -95,3 +95,19 @@ def random_str(n: int) -> str:
             "authorName": random_str(1536),
         }
         driver.execute_query(insert_query, parameters)
+
+
+@pytest.fixture(scope="module")
+def setup_schema(driver):
```
leila-messallem, line 101: Naming suggestions are welcome.

willtai: How about `setup_neo4j_for_schema_query`? We can also rename the other `setup_neo4j` fixture to `setup_neo4j_for_retrieval`

tests/e2e/test_schema_e2e.py
leila-messallem: [8de0edd](https://github.com/neo4j/neo4j-genai-python/pull/36/commits/8de0eddbb6f0c1daabcc85de4082f705c737cf91)

tests/e2e/conftest.py
leila-messallem: Sounds good! 👍

leila-messallem: [8eacfb6](https://github.com/neo4j/neo4j-genai-python/pull/36/commits/8eacfb6020596ab2cbb58eb1c67bfa40894b45de)

willtai (approved):
Looks great!

PR #37, opened 2024-05-21
Adds a text to cypher retriever

alexthomas93:
Adds the `Text2CypherRetriever` class which allows for the retrieval of records from a Neo4j database using natural language. This retriever converts a user's natural language query to a Cypher query using an LLM and then retrieves records from a Neo4j database using the generated Cypher query. PR includes unit tests, an end-to-end test, and an example script detailing how to use the class.


src/neo4j_genai/retrievers/text_to_cypher.py
```diff
@@ -0,0 +1,104 @@
...
+            if validated_data.examples
+            else "",
+            input=validated_data.query_text,
+        )
+        logger.debug("TextToCypherRetriever prompt: %s", prompt)
+
+        t2c_query = self.llm.invoke(prompt)
+        logger.debug("TextToCypherRetriever Cypher query: %s", t2c_query)
+
+        records, _, _ = self.driver.execute_query(query_=t2c_query)
```
willtai, line 102: What would happen if the Cypher query generated by TextToCypherRetriever is invalid? Should we have a try/catch for the Cypher error?

src/neo4j_genai/retrievers/text_to_cypher.py
```diff
@@ -0,0 +1,104 @@
...
+    Neo4jDriverModel,
+    Neo4jSchemaModel,
+    TextToCypherRetrieverModel,
+    TextToCypherSearchModel,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class TextToCypherRetriever(Retriever):
```
oskarhane, line 36: Could you rename this to `Text2CypherRetriever`, it's a convention that's already established for Cypher but also in `Text2SQL` functions.

src/neo4j_genai/prompts.py
```diff
@@ -0,0 +1,30 @@
...
+{schema}
+
+Examples (optional):
+{examples}
+
+Input:
+{input}
+
+Do not use any properties or relationships not included in the schema.
+Do not include triple backticks ``` or any additional text except the generated Cypher statement in your response.
```
oskarhane, line 29: How did you come up with this prompt?

src/neo4j_genai/retrievers/text_to_cypher.py
alexthomas93: Done!

src/neo4j_genai/prompts.py
alexthomas93: I came up with it myself. It was based on the prompts [here](https://github.com/langchain-ai/langchain/blob/09919c2cd5398068c43662ff3acf2f5c21c35747/libs/community/langchain_community/chains/graph_qa/prompts.py#L98). None of those prompts seem quite right however as most request the LLM use the returned records from the Neo4j database to directly respond to the user's input. Given this is a retriever I assume we just want the raw records returned. They also don't allow for the addition of few-shot learning examples, which in LangChain is done through the `PromptTemplate` class. I considered implementing a version of this class here but thought it would be better to keep things simple and just use templated strings for now. I also tried LangChain hub but couldn't find anything useful there.

examples/text_2_cypher_search.py
```diff
@@ -1,6 +1,6 @@
 from langchain_openai import OpenAI
 from neo4j import GraphDatabase
-from neo4j_genai.retrievers.text_to_cypher import TextToCypherRetriever
+from neo4j_genai.retrievers.text_2_cypher import Text2CypherRetriever
```
willtai, line 3: Should we remove underscores to keep it consistent?
```suggestion
from neo4j_genai.retrievers.text2cypher import Text2CypherRetriever
```

src/neo4j_genai/retrievers/text_to_cypher.py
alexthomas93: I've added a try except block which logs and re-raises the error in the event the LLM generates incorrect Cypher. Let me know if you think we need anything else for this scenario.

willtai: We should think about the user experience here if the Cypher query isn't valid. Should we try generating a few more Cypher queries before throwing an error, at the expense of LLM calls? Should we throw an error immediately and advise the user to provide a more descriptive prompt or schema? I think we can try to make this nicer for the user

stellasia: At the moment, we could let the user decides what to do next IMHO. When we'll have agents or more complex retrieval strategies, we will be able to  deal with these failures.

src/neo4j_genai/prompts.py
alexthomas93: Done!

examples/text_2_cypher_search.py
alexthomas93: Done! I've renamed text_2_cypher text2cypher everywhere to make it consistent across the package

src/neo4j_genai/retrievers/text_to_cypher.py
alexthomas93: @willtai In the event Cypher query generation fails the user would need to either adjust the parameters of the LLM (temperature, etc.) or add in examples for the LLM to learn from. I can update the error message to be a bit more descriptive but I'm not sure that it'd be worth re-trying automatically as if query generation is failing, then there's probably something wrong with the prompt or the way the LLM is being used and the user will need to intervene.

Additionally, the idea currently is to not allow the users to edit the full prompt but to just to add examples the LLM can learn from. @oskarhane should we change this and allow users to edit the full prompt?

oskarhane: I think I agree with @stellasia on this. 
For now we should let the developer know it failed and give them the full error message. Not sure what the pythonic way of doing this is, if it's to raise an exception and let the developer catch it or return something with a status field in it.

alexthomas93: I've done a bit of research and based on the advice in [here](https://eli.thegreenplace.net/2008/08/21/robust-exception-handling/) we should prefer raising exceptions to returning status codes. Currently I'm just logging the exception then re-raising it as a `RuntimeError`. We might want to think of specifying our own Exceptions at some point though as I imagine an LLM failing to return the correct response will be a common occurrence when using this library.

willtai: I like that. I think we can improve our error handling using custom exceptions and that allows this application to be more scalable and maintainable to be used in other Python applications. I don't think that needs to block this PR though

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -0,0 +1,109 @@
...
+
+        try:
+            t2c_query = self.llm.invoke(prompt)
+            logger.debug("Text2CypherRetriever Cypher query: %s", t2c_query)
+            records, _, _ = self.driver.execute_query(query_=t2c_query)
+        except CypherSyntaxError as e:
+            logger.error("Cypher query generation failed: %s", e.message)
+            raise RuntimeError(f"Cypher query generation failed: {e.message}")
+
+        return records
```
oskarhane, line 109: It seems useful to return the generated query here, in addition to the records (would be useful to show in UI:s etc).

Right now we don't have a shared return type from our retrievers, but we would need to have that once we want to add the `Neo4jRAG` instance to handle retrieval.

We have alternatives though, maybe there's a different return type depending on a flag that's passed in `include_metadata` or something.
I don't know, please think a bit about this both for this local code but also for the larger story.

stellasia: In the RAG-eval code, we used this model which is returned by all retrievers:
```python
class RetrieverResult(BaseModel):
    documents: list[Document]
    execution_metadata: Optional[dict] = None
```
The generated cypher query (if any), is stored in the `execution_metadata` dict.

alexthomas93: What sort of metadata will need to be returned when using the other retrievers? LangChain retrievers return a list of `Document` objects (see [here](https://api.python.langchain.com/en/latest/documents/langchain_core.documents.base.Document.html#langchain_core.documents.base.Document)) which allow you to attach arbitrary pieces of metadata to each document. We could do something similar here? Alternatively we return one `Metadata` object with each search? I guess it depends if we expect different metadata for each record. The first approach would be more extensible but for this specific retriever would mean attaching the query to each record which would be a tad wasteful.

oskarhane (approved):
Looks good from my point of view, but wait for others to approve as well ⚡

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+        except ValidationError as e:
+            raise ValueError(f"Validation failed: {e.errors()}")
+
+        super().__init__(validated_data.driver_model.driver)
+        self.llm = validated_data.llm_model.llm
+        self.neo4j_schema = (
+            validated_data.neo4j_schema_model.neo4j_schema
+            if validated_data.neo4j_schema_model
+            else get_schema(validated_data.driver_model.driver)
```
willtai, line 66: I'm thinking if we should try/catch an error here too if get_schema() fails. Perhaps add some clarification of this in the docstring

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -0,0 +1,109 @@
...
+    ) -> list[neo4j.Record]:
+        """Converts query_text to a Cypher query using an LLM.
+           Retrieve records from a Neo4j database using the generated Cypher query.
+
+        Args:
+            query_text (str): The natural language query used to search the Neo4j database.
+            examples (Optional[list[str], optional): Optional user input/query pairs for the LLM to use as examples.
+
+        Raises:
+            ValidationError: If validation of the input arguments fail.
```
willtai, line 80: Should add `CypherSyntaxError` can also be raised. We should have this at least until we implement exceptions

src/neo4j_genai/retrievers/text2cypher.py
alexthomas93: Should handling database errors not be done in the `get_schema` function itself? If `get_schema` fails the only thing we can really do is log then re-raise the error, and it might even just be better to log the error in the `get_schema` function also

src/neo4j_genai/retrievers/text2cypher.py
alexthomas93: Done!

src/neo4j_genai/retrievers/text2cypher.py
willtai: I think you're right that we should raise exceptions/errors at `get_schema` function level but I think it will also be meaningful here to raise an error here too with a message like `"Failed to get schema for Text2CypherRetriever {error.message}"`. Of course when we refactor to exception handling this will be changed but I think we can raise at both levels

alexthomas93: Happy to add it in this case but I think we need to come up with a strategy for when to add try-except statements to our code. Adding them at all levels of a call stack just to do logging doesn't seem like best practice.

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -0,0 +1,116 @@
...
+            )
+            validated_data = Text2CypherRetrieverModel(
+                driver_model=driver_model,
+                llm_model=llm_model,
+                neo4j_schema_model=neo4j_schema_model,
+            )
+        except ValidationError as e:
+            raise ValueError(f"Validation failed: {e.errors()}")
+
+        super().__init__(validated_data.driver_model.driver)
```
stellasia, line 61: Not for this PR, but since we are passing the driver to the parent class, I think the validation of the driver type should happen here, wdyt?

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -0,0 +1,116 @@
...
+        self.llm = validated_data.llm_model.llm
+        try:
+            self.neo4j_schema = (
+                validated_data.neo4j_schema_model.neo4j_schema
+                if validated_data.neo4j_schema_model
+                else get_schema(validated_data.driver_model.driver)
+            )
+        except (Neo4jError, DriverError) as e:
```
stellasia, line 69: I think `DriverError` has no `message` attribute. To me we can just print the exception itself. Or we can use [`logger.exception(e)`](https://docs.python.org/3/library/logging.html#logging.exception)

```python
try:
    raise DriverError()
except Exception as e:
    print(e.message)
```

==>
```
Traceback (most recent call last):
  File "<ipython-input-5-2a34558e7899>", line 2, in <module>
    raise DriverError()
neo4j.exceptions.DriverError
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/estelle/.cache/pypoetry/virtualenvs/rag-evaluation-njKKJ5sT-py3.12/lib/python3.12/site-packages/IPython/core/interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-5-2a34558e7899>", line 4, in <module>
    print(e.message)
          ^^^^^^^^^
AttributeError: 'DriverError' object has no attribute 'message'
```

alexthomas93: Updated!

stellasia (approved):
Nice one!

willtai (approved):
Nice work 💯

PR #43, opened 2024-05-27
Get structured DB schema

leila-messallem:
# Description

Added a method for providing the database schema in a structured format.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_genai/schema.py
+++ b/src/neo4j_genai/schema.py
@@ -89,50 +89,27 @@ def get_schema(
     Returns:
         str: the graph schema information in a serialized format.
     """
-    node_properties = [
-        data["output"]
-        for data in query_database(
-            driver,
-            NODE_PROPERTIES_QUERY,
-            params={"EXCLUDED_LABELS": EXCLUDED_LABELS + [BASE_ENTITY_LABEL]},
-        )
-    ]
+    structured_schema = get_structured_schema(driver)
 
-    rel_properties = [
-        data["output"]
-        for data in query_database(
-            driver, REL_PROPERTIES_QUERY, params={"EXCLUDED_LABELS": EXCLUDED_RELS}
-        )
-    ]
-    relationships = [
-        data["output"]
-        for data in query_database(
-            driver,
-            REL_QUERY,
-            params={"EXCLUDED_LABELS": EXCLUDED_LABELS + [BASE_ENTITY_LABEL]},
-        )
-    ]
+    def _format_props(props):
+        return ", ".join([f"{prop['property']}: {prop['type']}" for prop in props])
 
     # Format node properties
-    formatted_node_props = []
-    for element in node_properties:
-        props_str = ", ".join(
-            [f"{prop['property']}: {prop['type']}" for prop in element["properties"]]
-        )
-        formatted_node_props.append(f"{element['labels']} {{{props_str}}}")
+    formatted_node_props = [
+        f"{label} {{{_format_props(props)}}}"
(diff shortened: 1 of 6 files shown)


src/neo4j_genai/schema.py
```diff
@@ -138,3 +145,54 @@ def get_schema(
...
+            REL_QUERY,
+            params={"EXCLUDED_LABELS": EXCLUDED_LABELS + [BASE_ENTITY_LABEL]},
+        )
+    ]
+
+    # Get constraints and indexes
+    try:
+        constraint = query_database(driver, "SHOW CONSTRAINTS")
+        index = query_database(driver, INDEX_QUERY)
+    except ClientError:
```
stellasia, line 189: In which case would these queries fail? Apoc missing?

src/neo4j_genai/schema.py
```diff
@@ -138,3 +145,54 @@ def get_schema(
             "\n".join(formatted_rels),
         ]
     )
+
+
+def get_structured_schema(driver: neo4j.Driver) -> dict[str, Any]:
```
stellasia, line 150: Can't this function be used in the `get_schema` function? Seems to be a lot of duplicated code.

src/neo4j_genai/schema.py
leila-messallem: Not all users are allowed to create constraint and indexes. Apparently it's not allowed at all in Community edition, because the E2E tests are failing because of it, so I will have to setup a different test depending on Neo4j edition 😅

src/neo4j_genai/schema.py
leila-messallem: I couldn't find a way of doing that without making the code unreadable. 🤷‍♀️

src/neo4j_genai/schema.py
leila-messallem: Sorry, we're not creating anything here (I was thinking about the tests). The problem here is read-only user. I can add a comment.

src/neo4j_genai/schema.py
leila-messallem: [1f2561e](https://github.com/neo4j/neo4j-genai-python/pull/43/commits/1f2561efa30cae676a8e2e31b7b09297c85a8bfc)

src/neo4j_genai/schema.py
```diff
@@ -82,6 +89,51 @@ def get_schema(
     Returns:
         str: the graph schema information in a serialized format.
     """
+    structured_schema = get_structured_schema(driver)
```
stellasia, line 92: Are we planning to add more functions to get the schema as a text but using different formats? (I think I read it somewhere but I'm not sure). In this case, shall we try and find a more descriptive name for the `get_schema` function?

stellasia (approved):

src/neo4j_genai/schema.py
leila-messallem: You're probably thinking about [this card](https://trello.com/c/Bm4TVI5d) about enhancing the schema with values? But that would be used internally, so I don't think we have to worry about it mixing up the naming.

stellasia: I think this function could be used by LangChain/LLamaIndex eventually. But we can also think about that when/if needed in the future.

PR #45, opened 2024-05-28
Retriever results

stellasia:
# Description

Add a `RetrieverResult` class. To deal with the retrievers with a `retrieval_query`, introduce a `format_record_function` that users can provide when instantiating the retriever (at the same time as the retrieval query). For retrievers with fixed outputs (VectorRetriever) or unconstrained output (Text2Cypher), a default formatting function is used.

**BREAKING CHANGE**
The method to be instantiated by all `Retriever` is `_get_search_results`, that must return a list of `neo4j.Record` and an optional metadata dictionnary (e.g.: the generated Cypher query for `Text2CypherRetriever`).

Also, the `ExternalRetriever` now inherits from `Retriever` to be able to use the new implementation of `search`, which means it requires a `neo4j.Driver` during instantiation. However, we do not have any constraint on the neo4j version there (no vector index), so retrievers classes have a `VERIFY_NEO4J_VERSION` class attribute, which is `True` by default but has been toggled to `False` for external retrievers.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_genai/retrievers/base.py
```diff
@@ -75,24 +73,63 @@ def _fetch_index_infos(self):
...
+        ]
+        metadata = raw_result.metadata or {}
+        metadata["__retriever"] = self.__class__.__name__
+        return RetrieverResult(
+            items=search_items,
+            metadata=metadata,
+        )
+
+    @abstractmethod
+    def get_search_results(self, *args, **kwargs) -> RetrieverRawResult:
```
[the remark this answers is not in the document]
stellasia: Yes, that's the idea. In this way, we can have a lot of common code in the `search` method. Probably more could be factorized but that would be too much changes for this PR (and need to be discussed internally :) )

willtai, line 95: If only `.search()` is exposed to users, should we make this a private method `_get_search_results()`?

src/neo4j_genai/types.py
```diff
@@ -27,17 +27,57 @@
...
+    """
+    Represents the raw result returned from the retriever get_search_result
+    method. It needs to be formatted further before being returned as a RetrieverResult.
+
+    Attributes:
+        records (list[neo4j.Record]): A list of records from neo4j.
+        metadata: Record-related metadata, such as score.
+    """
+
+    records: list[neo4j.Record]
```
mgozsoy-neo4j, line 54: Is it possible to merge RetrieverRawResult and RetrieverResult, such that RetrieverResultItem either contains str_content or record_content? 
As the result user wouldn't need to think/learn when to use either of them

src/neo4j_genai/types.py
```diff
@@ -27,17 +27,57 @@
...
+    metadata: Optional[dict] = None
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
mgozsoy-neo4j, line 57: Is model_config relevant to retrieved result? If not, it could be better to some other class.

stellasia: `model_config` is a Pydantic parameter. We use it to tell pydantic not to perform type checking (because `neo4j.*` do not implement the required methods for that)

src/neo4j_genai/types.py
stellasia: The `get_search_results` method returns `neo4j.Record`, but the results we need for RAG is a string. That why we introduced both models. It also makes sense to have a string returned here, since the way to extract the text content depends on the retriever.

willtai (commented):
I think you can also replace `VectorSearchRecord` with `RetrieverRawResult` in the documentation. It's located in `types.rst`

src/neo4j_genai/types.py
willtai: `RetrieverRawResult` does make me think that it's a parent of `RetrieverResult` though. Perhaps we can rename `RetrieverRawResult` to something like `RawSearchResult`? As it's a result for the retriever directly

stellasia: Thinking again about that : 
- The inconvenient is that users who may want to test different formatting will have to re-run the retriever.
- Advantage is that when we will have agents, we will be able to have one formatting per retriever for free, if we postpone the formatting it will be harder to have something other than `str(retriever_result)`.

.github/workflows/pr-e2e-tests.yaml
```diff
@@ -71,4 +71,11 @@ jobs:
         run: |
           count=0; until curl -s --fail localhost:8080/v1/.well-known/ready; do ((count++)); [ $count -ge 10 ] && echo "Reached maximum retry limit" && exit 1; sleep 15; done
       - name: Run tests
-        run: poetry run pytest ./tests/e2e
+        shell: bash
+        run: |
+          if [[ "${{ matrix.neo4j-edition }}" == "community" ]]; then
+              poetry run pytest -m 'not enterprise_only' ./tests/e2e
```
willtai, line 77: Why run with this pytest mark?

stellasia: This file is not supposed to be updated, let me check what happened (probably a wrong merge)

stellasia: Should be fixed after merge with main.

oskarhane (approved):
Nice ☂️

PR #49, opened 2024-06-03
Type everything. Add mypy checks to pre-commit and Github Actions

jonbesga:
# Description

> **Note**
>
>
>

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_genai/filters.py
```diff
@@ -55,11 +55,11 @@ def safe_field_cypher(field_name: str) -> str:
         escaped_field = field_name.replace("`", "``")
         return f"`{escaped_field}`"
 
-    def lhs(self, field):
+    def lhs(self, field: str) -> str:
         safe_field_cypher = self.safe_field_cypher(field)
         return f"{self.node_alias}.{safe_field_cypher}"
 
-    def cleaned_value(self, value):
+    def cleaned_value(self, value: str) -> str:
```
stellasia, line 62: Here `value` should be `Any`. We might want to compare stuff to float or anything, we can also have lists.

src/neo4j_genai/filters.py
```diff
@@ -90,7 +90,7 @@ class GteOperator(Operator):
 class InOperator(Operator):
     CYPHER_OPERATOR = "IN"
 
-    def cleaned_value(self, value):
+    def cleaned_value(self, value: str) -> str:
```
stellasia, line 93: `value` should be `list[float | int | str]` here.

src/neo4j_genai/filters.py
```diff
@@ -164,11 +164,11 @@ class ParameterStore:
     Determine the parameter name depending on a parameter counter
     """
 
-    def __init__(self):
-        self._counter = Counter()
-        self.params = {}
+    def __init__(self) -> None:
+        self._counter: Counter[str] = Counter()
+        self.params: dict[str, Union[str, int, list[Any]]] = {}
```
stellasia, line 169: I'd use `Any` as type for the dict value, it could potentially be a float or even a date (good question, we should test dates btw)

src/neo4j_genai/filters.py
```diff
@@ -185,7 +185,7 @@ def _get_params_name(self):
         self._counter[key] += 1
         return param_name
 
-    def add(self, value):
+    def add(self, value: Union[str, int, list[Any]]) -> Union[str, int, list[Any]]:
```
stellasia, line 188:
```suggestion
    def add(self, value: Any) -> str:
```

stellasia: It always returns the name of the parameter as string

src/neo4j_genai/types.py
```diff
@@ -48,7 +48,7 @@ class IndexModel(BaseModel):
     driver: Any
 
     @field_validator("driver")
-    def check_driver_is_valid(cls, v):
+    def check_driver_is_valid(cls, v: Any) -> neo4j.Driver:
```
stellasia, line 51: Is this type of check on the parameter type still useful when using mypy?

src/neo4j_genai/embeddings/sentence_transformers.py
```diff
@@ -1,8 +1,11 @@
 from neo4j_genai.embedder import Embedder
+from typing import Any
 
 
 class SentenceTransformerEmbeddings(Embedder):
-    def __init__(self, model="all-MiniLM-L6-v2", *args, **kwargs):
+    def __init__(
+        self, model: str = "all-MiniLM-L6-v2", *args: Any, **kwargs: Any
```
stellasia, line 7: Wow, that's a pretty ugly syntax :( Do you know the reason why we need to add types to args and kwargs? Do we have the ability to add constraints like 'any kwarg must be a float'?

src/neo4j_genai/neo4j_queries.py
```diff
@@ -56,9 +56,9 @@ def _get_hybrid_query() -> str:
 
 def _get_filtered_vector_query(
     filters: dict[str, Any],
-    node_label: str,
-    embedding_node_property: str,
-    embedding_dimension: int,
+    node_label: Optional[str],
```
stellasia, line 59: I think at this stage the parameters should not be optional anymore (None values are not managed inside this function). But if you had to add it, probably this is something we need to change in the code.

src/neo4j_genai/retrievers/external/weaviate/weaviate.py
```diff
@@ -130,6 +132,7 @@ def search(
                 logger.debug(
                     "No embedder provided, assuming vectorizer is used in Weaviate."
                 )
+        weaviate_filters = kwargs.get("weaviate_filters")
```
stellasia, line 135: Is this line needed?

src/neo4j_genai/types.py
willtai: I think it might be the only way to check for custom classes that don't have `_ get _pydantic_core schema _` set up unfortunately when I was looking into it https://docs.pydantic.dev/latest/concepts/types/#customizing-validation-with-__get_pydantic_core_schema__  Unless there is an alternative? Otherwise it's a bit of defensive programming here

src/neo4j_genai/embeddings/sentence_transformers.py
willtai: https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs Based on this it is possible, are we able to restrict the type of these further?

src/neo4j_genai/filters.py
willtai: If so should it be `list[float | int | str]` as you suggested below?

src/neo4j_genai/filters.py
willtai: If possible I'd avoid `Any` if we know it's a `float` or a date type

src/neo4j_genai/embeddings/sentence_transformers.py
stellasia: 🤯

No I don't think we can restrict it, but I was not aware of this feature, nice :)

src/neo4j_genai/types.py
jonbesga: > Is this type of check on the parameter type still useful when using mypy?

Yes because mypy is an static checker and this wil check dynamically on runtime

src/neo4j_genai/filters.py
jonbesga: c9081d159d46cdaddbf90285271128c4cf7dc91a

src/neo4j_genai/filters.py
jonbesga: c9081d159d46cdaddbf90285271128c4cf7dc91a

src/neo4j_genai/filters.py
jonbesga: c9081d159d46cdaddbf90285271128c4cf7dc91a

src/neo4j_genai/retrievers/external/weaviate/weaviate.py
jonbesga: c9081d159d46cdaddbf90285271128c4cf7dc91a

src/neo4j_genai/neo4j_queries.py
jonbesga: This one is the golden crown. If we don't make them optional then there are some parts of the code that need to be changed because the code sends `None` in some cases :D

src/neo4j_genai/filters.py
```diff
@@ -90,7 +90,7 @@ class GteOperator(Operator):
 class InOperator(Operator):
     CYPHER_OPERATOR = "IN"
 
-    def cleaned_value(self, value):
+    def cleaned_value(self, value: Any) -> Any:
```
willtai, line 93: I think value should be of type `list[Union[str, int, float]]` since we do this type check below anyways

src/neo4j_genai/types.py
willtai, line 51: Is it possible to make this type hint v: neo4j.Driver?

willtai (commented):
Very nice work! Left some comments where the type hints are `Any`. I'd pay more attention to those as they make the type checks redundant.

For example, in the function `_handle_field_filter`:

```python
def _handle_field_filter(
    field: str,
    value: Any,
    param_store: ParameterStore,
    node_alias: str = DEFAULT_NODE_ALIAS,
) -> str:
```

I think we can use `dict` for `value`. I understand this was already there before so I couldn't comment on the diff

src/neo4j_genai/filters.py
jonbesga: You mean a `Union[str, int, float]`

jonbesga:
> ```python
> _handle_field_filter
> ```

Value is actually not a dict, if you look at the code you put "str" in there

src/neo4j_genai/filters.py
jonbesga: 5ab1e8ca0179137388dc7b8386e5a7cbc0008d06

src/neo4j_genai/types.py
jonbesga: 5ab1e8ca0179137388dc7b8386e5a7cbc0008d06

jonbesga:
> Very nice work! Left some comments where the type hints are `Any`. I'd pay more attention to those as they make the type checks redundant.
> 
> For example, in the function `_handle_field_filter`:
> 
> ```python
> def _handle_field_filter(
>     field: str,
>     value: Any,
>     param_store: ParameterStore,
>     node_alias: str = DEFAULT_NODE_ALIAS,
> ) -> str:
> ```
> 
> I think we can use `dict` for `value`. I understand this was already there before so I couldn't comment on the diff

5ab1e8ca0179137388dc7b8386e5a7cbc0008d06

willtai:
> Very nice work! Left some comments where the type hints are `Any`. I'd pay more attention to those as they make the type checks redundant.
> 
> For example, in the function `_handle_field_filter`:
> 
> ```python
> def _handle_field_filter(
>     field: str,
>     value: Any,
>     param_store: ParameterStore,
>     node_alias: str = DEFAULT_NODE_ALIAS,
> ) -> str:
> ```
> 
> I think we can use `dict` for `value`. I understand this was already there before so I couldn't comment on the diff

Now looking closer I think I am wrong. Based on this docstring

```
If provided as is then this will be an equality filter
            If provided as a dictionary then this will be a filter, the key
            will be the operator and the value will be the value to filter by
```

 and the type checks belowther than a `dict` it can also be a `str`, is this correct @stellasia ?

jonbesga:
> > other than a dict it can also be a str,
> 
> Yes, dict is used to specify the comparison operator, but equality comparison can be implicit. When implicit, the value is passed directly (no dict). And in this case, the value can be of type `str` but also `int` or `float` or anything (depending on the field we are filtering on). So I think `Any` is not that bad here :)

All happy? Approve before someone else merge something and I have to add more types 🥲

willtai (approved):
Great work on this. I think there are some minor things to fix with ruff formatting but otherwise I think it's ready

stellasia (approved):
Go!!! 🔥

PR #55, opened 2024-06-05
Update all docstrings and exceptions to be used for documentation

willtai:
# Description

This PR prepares the repository for the first version of the official documentation. The changes were:

- Fixed and updated docstrings to be clearer
- Update custom exceptions where generic exceptions are used
- Update documentation

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

To test the documentation build locally. Following instructions under `docs/` i.e. run `make -c docs html` to generate a build.

# Checklist


src/neo4j_genai/retrievers/external/weaviate/weaviate.py
```diff
@@ -36,6 +36,44 @@
...
+          )
+
+          result = retriever.search(query_text="biology", top_k=2)
+
+    Args:
+        driver (neo4j.Driver): The Neo4j Python driver.
+        client (WeaviateClient): The Weaviate client object.
+        collection (str): Name of a set of Weaviate objects that share the same data structure.
+        id_property_external (str): Weaviate property identifier that refers to a corresponding Neo4j node id.
+        id_property_neo4j (str): Neo4j node id that refers to a corresponding `id_property_external`.
```
willtai, line 68: I'm unsure of the descriptions here. Open to suggestions

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -67,15 +81,10 @@ def __init__(
                 if validated_data.neo4j_schema_model
                 else get_schema(validated_data.driver_model.driver)
             )
-        except Neo4jError as e:
-            logger.error("Failed to get schema for Text2CypherRetriever: %s", e.message)
-            raise RuntimeError(
-                f"Failed to get schema for Text2CypherRetriever: {e.message}"
+        except (Neo4jError, DriverError) as e:
+            raise SchemaFetchError(
+                f"Failed to fetch schema for Text2CypherRetriever: {e}"
```
willtai, line 86: `mypy` tells me `e.message` doesn't exist

src/neo4j_genai/retrievers/external/weaviate/weaviate.py
oskarhane, line 68:
```suggestion
        id_property_neo4j (str): The name of the Neo4j node property that's used as the identifier for relating matches from Weaviate to Neo4j nodes.
```

src/neo4j_genai/retrievers/external/weaviate/weaviate.py
```diff
@@ -36,6 +36,44 @@
...
+              id_property_neo4j="id"
+          )
+
+          result = retriever.search(query_text="biology", top_k=2)
+
+    Args:
+        driver (neo4j.Driver): The Neo4j Python driver.
+        client (WeaviateClient): The Weaviate client object.
+        collection (str): Name of a set of Weaviate objects that share the same data structure.
+        id_property_external (str): Weaviate property identifier that refers to a corresponding Neo4j node id.
```
oskarhane, line 67:
```suggestion
        id_property_external (str): The name of the Weaviate property that has the identifier that refers to a corresponding Neo4j node id property.
```

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -115,8 +124,9 @@ def _get_search_results(
             logger.debug("Text2CypherRetriever Cypher query: %s", t2c_query)
             records, _, _ = self.driver.execute_query(query_=t2c_query)
         except CypherSyntaxError as e:
-            logger.error("Cypher query generation failed: %s", e.message)
-            raise RuntimeError(f"Cypher query generation failed: {e.message}")
+            raise Text2CypherGenerationError(
+                f"Cypher query generation failed: {e.message}"
```
stellasia, line 128: We don't really know if the generation failed, or if it generated something that's invalid, right? Ideally, we should validate the generated query before executing it (for syntax error), but that's probably another topic. For now, I'd just rely on the python driver errors actually in this case.

src/neo4j_genai/retrievers/text2cypher.py
stellasia: I think it's only defined for `Neo4jError` and not for `DriverError`. If I remember well, that was the reason for splitting the error handling in two.

willtai: https://github.com/neo4j/neo4j-genai-python/pull/55/commits/b46b32db96c901cbd79be0a91071782d9566d488 I do a check here. I think this is less verbose

src/neo4j_genai/retrievers/text2cypher.py
willtai: I was thinking whether `Text2CypherGenerationError` here is generic enough. Perhaps the naming is off, maybe something like `Text2CypherRetrievalError` is better?

alexthomas93: Another scenario could be that it generates a valid query but includes some additional junk (```, etc.) which causes the query to fail. I agree, without knowing why retrieval fails it might be best to stick to something more generic like `Text2CypherRetrievalError` as opposed to `Text2CypherGenerationError` might bias the user too much when they're trying to determine why their retriever isn't working

willtai: Changed to `Text2CypherRetrievalError` in https://github.com/neo4j/neo4j-genai-python/pull/55/commits/53c2d3c6f32dc41425ab3adb15620eb1934a3ee6

stellasia (approved):
Nice one!

oskarhane (approved):
🍡

PR #58, opened 2024-06-07
Stopped the embedding from being returned with VectorRetriever

alexthomas93:
# Description
Stops embeddings from being returned when searching with VectorRetriever

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Low

## How Has This Been Tested?
- [X] Unit tests
- [X] E2E tests
- [X] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_genai/neo4j_queries.py
+++ b/src/neo4j_genai/neo4j_queries.py
@@ -147,7 +147,9 @@ def get_search_query(
     else:
         raise ValueError(f"Search type is not supported: {search_type}")
     query_tail = get_query_tail(
-        retrieval_query, return_properties, fallback_return="RETURN node, score"
+        retrieval_query,
+        return_properties,
+        fallback_return=f"RETURN apoc.map.removeKeys(node, ['{embedding_node_property}']) AS node, score",
     )
     return f"{query} {query_tail}", params
 
--- a/src/neo4j_genai/neo4j_queries.py
+++ b/src/neo4j_genai/neo4j_queries.py
@@ -149,7 +149,7 @@ def get_search_query(
     query_tail = get_query_tail(
         retrieval_query,
         return_properties,
-        fallback_return=f"RETURN apoc.map.removeKeys(node, ['{embedding_node_property}']) AS node, score",
+        fallback_return=f"RETURN node {{ .*, `{embedding_node_property}`: null }} AS node, score",
     )
     return f"{query} {query_tail}", params
 
--- a/src/neo4j_genai/neo4j_queries.py
+++ b/src/neo4j_genai/neo4j_queries.py
@@ -149,7 +149,7 @@ def get_search_query(
     query_tail = get_query_tail(
         retrieval_query,
         return_properties,
-        fallback_return=f"RETURN node {{ .*, _nodeLabels: labels(node), `{embedding_node_property}`: null }} AS node, score",
+        fallback_return=f"RETURN node {{ .*, `{embedding_node_property}`: null }} AS node, labels(node) AS nodeLabels, elementId(node) AS id, score",
     )
     return f"{query} {query_tail}", params
 
--- a/src/neo4j_genai/neo4j_queries.py
+++ b/src/neo4j_genai/neo4j_queries.py
@@ -149,7 +149,7 @@ def get_search_query(
     query_tail = get_query_tail(
         retrieval_query,
         return_properties,
-        fallback_return=f"RETURN node {{ .*, `{embedding_node_property}`: null }} AS node, score",
+        fallback_return=f"RETURN node {{ .*, _nodeLabels: labels(node), `{embedding_node_property}`: null }} AS node, score",
     )
     return f"{query} {query_tail}", params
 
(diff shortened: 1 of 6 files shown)


src/neo4j_genai/neo4j_queries.py
```diff
@@ -147,7 +147,9 @@ def get_search_query(
     else:
         raise ValueError(f"Search type is not supported: {search_type}")
     query_tail = get_query_tail(
-        retrieval_query, return_properties, fallback_return="RETURN node, score"
+        retrieval_query,
+        return_properties,
+        fallback_return=f"RETURN apoc.map.removeKeys(node, ['{embedding_node_property}']) AS node, score",
```
willtai, line 152: Do we know which specific properties from `node` we want to return? I'm wondering we should use `apoc.maps.get` instead as it's cleaner and readable.

Unless I misunderstood, do we actually want `node.text_node_property`?

oskarhane: Ideally we don't rely on APOC at all (because the full APOC Core is not available in Aura).
Is there a way to do this without APOC?

But I think removing keys over adding keys is better in this case (as a response to @willtai's comment.

alexthomas93: So we could do it the same way its done in LangChain:
```python
fallback_return=f"RETURN node {{ .*, `{embedding_node_property}`: null }} AS node, score"
```
This does still return the `embedding_node_property` as a key in the `content` but with a value of `null` e.g.
```python
items=[RetrieverResultItem(content="{'released': 1997, 'taglineEmbedding': None, 'title': 'As Good as It Gets', 'tagline': 'A comedy from the heart that goes for the throat.'}", metadata={'score': 0.7144040465354919})] metadata={'__retriever': 'VectorRetriever'}
```

oskarhane: Yeah, that makes sense to me

stellasia: I'd keep the node label information as well, something like:
```
fallback_return=f"RETURN node {{ .*, _nodeLabels: labels(node), `{embedding_node_property}`: null }} AS node, score"
```
?

alexthomas93: Done!

oskarhane: I don't like that we're adding magic properties to the node object. I wonder if a better solution is to do what the langchain code does and also return a `metadata` object that holds node labels and elementId.
Open for other suggestions though @alexthomas93 @stellasia.

```python
fallback_return=f"RETURN node {{ .*, `{embedding_node_property}`: null }} AS node, {nodeLabels: labels(node), id: elementId(node)} AS metadata, score"
```

stellasia: LGTM, a nice use of the record metadata @alexthomas93 :)

alexthomas93: I've moved them to the metadata. Currently responses from the `VectorRetriever` look like this:
```python
items=[RetrieverResultItem(content="{'released': 1997, 'taglineEmbedding': None, 'title': 'As Good as It Gets', 'tagline': 'A comedy from the heart that goes for the throat.'}", metadata={'score': 0.7144040465354919, 'nodeLabels': ['Movie'], 'id': '4:2c4492dc-b2d3-45fb-b175-e1de34b609a7:52'})] metadata={'__retriever': 'VectorRetriever'}
```

oskarhane:

src/neo4j_genai/neo4j_queries.py
willtai: Should the keys in metadata be in `camelCase` or `snake_case`? I'm wondering if it should be in snake case as we do for other Python variables. I'm happy with either as long as we keep things consistent

alexthomas93: The convention for Neo4j is `camelCase` is it not? If we wanted to move to `snake_case` we'd need to manually update all of the node properties before returning them

alexthomas93:
@CodiumAI-Agent /update_changelog --pr_update_changelog.push_changelog_changes=true

oskarhane:
> @CodiumAI-Agent /update_changelog --pr_update_changelog.push_changelog_changes=true

You need to add it manually @alexthomas93, otherwise it ends up in the wrong section.  
See https://github.com/neo4j/neo4j-genai-python?tab=readme-ov-file#updating-the-changelogmd

oskarhane:
Can we alter the message the codium bot writes when it generates the CHANGELOG entry suggestion @willtai?
It's confusing right now, it'd be better if it could say that it's a manual task to update the CHANGELOG.

alexthomas93:
> > @CodiumAI-Agent /update_changelog --pr_update_changelog.push_changelog_changes=true
> 
> You need to add it manually @alexthomas93, otherwise it ends up in the wrong section. See https://github.com/neo4j/neo4j-genai-python?tab=readme-ov-file#updating-the-changelogmd

Done!

oskarhane:
@alexthomas93 The changelog entry is under the wrong section. It should be under "Next". Did you forget to push?

alexthomas93:
> @alexthomas93 The changelog entry is under the wrong section. It should be under "Next". Did you forget to push?

Apologies I put it under `## 0.2.0a3 / ## Added` I've moved it to `## Next / ### Added` now

oskarhane (approved):
LGTM 🖼️

PR #68, opened 2024-06-25
doc/search_methods

stellasia:
# Description

After this PR, users will be able to see the expected arguments and the docstring for the retrievers search method. 

In practise, the docstring and signatures are copied from the`_get_search_results` method

- Pros:
  - It's clearer for the user
  - No more work required for developper: just implement the `_get_search_results` method
- Cons:
  - It looks like Python black magic
  - The search method is actually *copied* into the derived class, but in an automatic way (well, not sure it's actually bad, it is what it is)
  - Since this is dynamic, the static type checker from PyCharm (for instance) is not able to see it, and auto suggestion won't work in this case.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Low in terms of usage, High in terms of implemetnation

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


tests/unit/retrievers/test_base.py
```diff
@@ -37,12 +41,62 @@ def test_retriever_version_support(
...
         MockRetriever(driver=driver)
+
+
+@patch("neo4j_genai.retrievers.base.Retriever._verify_version")
+def test_retriever_search_docstring_copied(
+    _verify_version_mock: MagicMock,
+    driver: MagicMock,
+) -> None:
+    class MockRetriever(Retriever):
+        def _get_search_results(self, query: str, top_k: int = 10) -> RawSearchResult:
```
willtai, line 61: Should `_get_search_results` be a public method since it's optionally customisable by users?

src/neo4j_genai/retrievers/base.py
```diff
@@ -13,15 +13,69 @@
...
 import neo4j
 
 from neo4j_genai.types import RawSearchResult, RetrieverResult, RetrieverResultItem
 from neo4j_genai.exceptions import Neo4jVersionError
 
+T = ParamSpec("T")
+P = TypeVar("P")
+
+
+def copy_function(f: Callable[T, P]) -> Callable[T, P]:
```
willtai, line 34: Maybe be more explicit and name it `deepcopy_function`

tests/unit/retrievers/test_base.py
stellasia: I think it used to be public and we updated it because it is not callable from outside the class :) I do not have a strong opinion on this one TBH, could be public for the reason you describe.

willtai: Now that I think about it, having it private makes it confusing for the user as to whether it should be implemented. It should probably be public like in the first place, wdyt?

stellasia: Yes makes sense to me, private also tends to be synonym of "do not touch". I'll update it to public and add a comment so that we try not to change it again in the future :)

stellasia: And actually, other methods in this class with the same goal, like `get_result_formatter` were not private, so it's better in terms of consistency.

Done in https://github.com/neo4j/neo4j-genai-python/pull/68/commits/f3a589b193b2c720ff9853344c08a6878cdc247e

src/neo4j_genai/retrievers/base.py
```diff
@@ -13,15 +13,70 @@
...
 
 from neo4j_genai.types import RawSearchResult, RetrieverResult, RetrieverResultItem
 from neo4j_genai.exceptions import Neo4jVersionError
 
+T = ParamSpec("T")
+P = TypeVar("P")
+
 
-class Retriever(ABC):
+def copy_function(f: Callable[T, P]) -> Callable[T, P]:
```
willtai, line 35: This is a cool approach but I'm wondering whether doing a deepcopy for every retriever instance just to get typehints and docstrings is a bit too complex, and whether it introduces any performance overhead from the deep copying. 

Did you consider other approaches that don't involve copying? Something like

```python
def propagate_docstring_signature(cls):
    get_search_results = getattr(cls, "get_search_results", None)
    if get_search_results and "search" not in cls.__dict__:

        def search(self, *args, **kwargs):
            return get_search_results(self, *args, **kwargs)

        search.__doc__ = get_search_results.__doc__
        search.__signature__ = inspect.signature(get_search_results)
        setattr(cls, "search", search)
    return cls


@propagate_docstring_signature
class Retriever(ABC):
    ...
```

Inspired by this post: https://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions

willtai: Unless there was a specific reason to go with this approach?

stellasia: I agree about the bad complexity/gain ratio, but couldn't find any other working solution that do not require more work for the developer. 
If not (deep) copying, the search method is always the one attached to the BaseRetriever, and hence gets the last documentation attached to it, for all retrievers. In your code snippet, we have to consider that the search method is more complex than just calling the `get_search_results` method, and it's also a bit weird to me to have this method "hanging" in the middle.

Regarding performances, this is done only once when the class is discovered (so on first import of the package), so it's not a concern IMHO.

The only other working solution I could think of, is to override the search method with a docstring and parameters. The advantage is that we would have autocompletion working, but then the developer needs to instantiate this method "just" for documentation purposes.

willtai: I see, in that case I'm happy to go with this approach 👍

src/neo4j_genai/retrievers/base.py
```diff
@@ -140,4 +195,8 @@ def _get_search_results(
                 list[neo4j.Record]: List of Neo4j Records
 
         """
+        # NB: even though this method is not intended to be
```
willtai, line 198: Should have this as a comment or a docstring for the method?

src/neo4j_genai/retrievers/base.py
```diff
@@ -93,7 +144,20 @@ def search(self, *args: Any, **kwargs: Any) -> RetrieverResult:
         )
 
     @abstractmethod
-    def _get_search_results(self, *args: Any, **kwargs: Any) -> RawSearchResult:
+    def get_search_results(self, *args: Any, **kwargs: Any) -> RawSearchResult:
```
willtai, line 147: I think we can add an example to the docs on how to implement this method, but doesn't have to be part of this PR, wdyt?

willtai (approved):
Nice work! Don't forget to the update the CHANGELOG before merging 💯

src/neo4j_genai/retrievers/base.py
stellasia: Yes sure, we can probably find an example to show implementation of a new retriever. For instance a retriever with a specific retrieval query as suggested by Oskar.

PR #69, opened 2024-06-26
Pydantic search model refactoring

alexthomas93:
# Description

This simplifies the Pydantic models used to validate the `search` methods of retrievers by removing fields that are already validated by the retrievers' `__init__` methods.

**NOTE** This PR does not merge the search models. Given the vector search models allow for only one input (either a `query_vector` or `query_text`) and the hybrid search models allow for both inputs it is not possible to merge them.

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [X] Unit tests
- [X] E2E tests
- [X] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_genai/retrievers/external/pinecone/types.py
+++ b/src/neo4j_genai/retrievers/external/pinecone/types.py
@@ -14,7 +14,7 @@
 #  limitations under the License.
 from __future__ import annotations
 
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Optional, Union
 
 import neo4j
 from pinecone import Pinecone
@@ -29,7 +29,7 @@ from neo4j_genai.types import EmbedderModel, Neo4jDriverModel, VectorSearchModel
 
 class PineconeSearchModel(VectorSearchModel):
     pinecone_filter: Optional[
-        Dict[str, Union[str, float, int, bool, List[Any], Dict[Any, Any]]]
+        dict[str, Union[str, float, int, bool, list[Any], dict[Any, Any]]]
     ] = None
 
 
--- a/src/neo4j_genai/retrievers/external/pinecone/types.py
+++ b/src/neo4j_genai/retrievers/external/pinecone/types.py
@@ -13,7 +13,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from __future__ import annotations
-from typing import Any, Callable, Optional
+
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import neo4j
 from pinecone import Pinecone
@@ -27,7 +28,9 @@ from neo4j_genai.types import EmbedderModel, Neo4jDriverModel, VectorSearchModel
 
 
 class PineconeSearchModel(VectorSearchModel):
-    pinecone_filter: Optional[dict[str, Any]] = None
+    pinecone_filter: Optional[
+        Dict[str, Union[str, float, int, bool, List[Any], Dict[Any, Any]]]
+    ] = None
 
 
 class PineconeClientModel(BaseModel):
--- a/src/neo4j_genai/retrievers/hybrid.py
+++ b/src/neo4j_genai/retrievers/hybrid.py
@@ -13,31 +13,31 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from __future__ import annotations
-from typing import Optional, Any, Callable
+
+import logging
+from typing import Any, Callable, Optional
 
 import neo4j
 from pydantic import ValidationError
 
 from neo4j_genai.embedder import Embedder
 from neo4j_genai.exceptions import (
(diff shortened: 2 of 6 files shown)


src/neo4j_genai/retrievers/external/pinecone/types.py
```diff
@@ -13,7 +13,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from __future__ import annotations
-from typing import Any, Callable, Optional
+
+from typing import Any, Callable, Dict, List, Optional, Union
```
stellasia, line 17: With the annotations import, I thought we do not have to use `Dict` and `List` from `typing`, builtin types are ok (`dict` and `list`). This is not the case?

src/neo4j_genai/retrievers/hybrid.py
```diff
@@ -13,31 +13,32 @@
...
 from pydantic import ValidationError
 
 from neo4j_genai.embedder import Embedder
 from neo4j_genai.exceptions import (
+    EmbeddingRequiredError,
     RetrieverInitializationError,
     SearchValidationError,
-    EmbeddingRequiredError,
 )
+from neo4j_genai.neo4j_queries import get_search_query
```
stellasia, line 29: Out of curiosity, which rule are you using to sort the imports? Is it something we should enforce for the whole team?

src/neo4j_genai/retrievers/vector.py
```diff
@@ -275,16 +276,17 @@ def _get_search_results(
...
                 query_text=query_text,
+                top_k=top_k,
                 query_params=query_params,
+                filters=filters,
             )
         except ValidationError as e:
             raise SearchValidationError(e.errors()) from e
 
         parameters = validated_data.model_dump(exclude_none=True)
+        parameters["vector_index_name"] = self.index_name
```
stellasia, line 289: I think `filters` can be excluded from these `parameters`, the only use of this variable is  to store the parameters for the Cypher query, right?

src/neo4j_genai/retrievers/hybrid.py
stellasia: I'm asking because it makes the review more difficult, if there is a convention there we can follow, happy to apply it.

src/neo4j_genai/retrievers/external/pinecone/types.py
alexthomas93: I believe we need to use the `typing` imports if we want to support Python 3.8 (https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc)

src/neo4j_genai/retrievers/hybrid.py
alexthomas93: I'm just using Ruff. I have it set to do formatting and import sorting through the Python VSCode extension. Maybe we could set up the CI/CD to check the imports are sorted correctly as well as the formatting is correct?

src/neo4j_genai/retrievers/vector.py
alexthomas93: Yeah it's used to store the params used to format the Cypher query. I'll add a `del` statement to get rid of the filters.

src/neo4j_genai/retrievers/external/pinecone/types.py
willtai: With `from __future__ import annotations` at the top of the file we shouldn't need to, does this cause an error for you?

alexthomas93: Have just ran the unit and E2E tests using 3.8 and the `list` and `dict` imports and seems there are no issues so I've updated them to use the builtin types

src/neo4j_genai/retrievers/hybrid.py
stellasia: Ok, thanks. Yes, makes sense to me to add it to CI, and maybe in README so that we can run the check locally before pushing - but not for this PR.

stellasia (approved):
LGTM!

willtai (approved):
✅ with a non-blocking question

src/neo4j_genai/retrievers/external/pinecone/types.py
```diff
@@ -27,7 +28,9 @@
 
 
 class PineconeSearchModel(VectorSearchModel):
-    pinecone_filter: Optional[dict[str, Any]] = None
+    pinecone_filter: Optional[
+        dict[str, Union[str, float, int, bool, list[Any], dict[Any, Any]]]
```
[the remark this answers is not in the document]
alexthomas93: I took them from the Pinecone Python library

PR #71, opened 2024-06-27
Allow specifying database name for retrievers and index handling methods

willtai:
# Description

Previously only the default database name is used because we didn't pass the database name to `driver.execute_query` calls. This PR allows users to optionally specify database names to the index handling methods and retrievers.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


alexthomas93:
The Weaviate and Pinecone retrievers will also need updating as they both user Neo4j databases too.

src/neo4j_genai/retrievers/base.py
```diff
@@ -185,11 +190,16 @@ class ExternalRetriever(Retriever, ABC):
...
 
     def __init__(
-        self, driver: neo4j.Driver, id_property_external: str, id_property_neo4j: str
+        self,
+        driver: neo4j.Driver,
+        id_property_external: str,
+        id_property_neo4j: str,
+        neo4j_database: Optional[str] = None,
     ):
         super().__init__(driver)
```
stellasia, line 199: The database can be passed to the parent class since ExternalRetriever inherits from Retriever.

src/neo4j_genai/retrievers/base.py
```diff
@@ -81,8 +81,9 @@ class Retriever(ABC, metaclass=RetrieverMetaclass):
     index_name: str
     VERIFY_NEO4J_VERSION = True
 
-    def __init__(self, driver: neo4j.Driver):
+    def __init__(self, driver: neo4j.Driver, database: Optional[str] = None):
```
stellasia, line 84: (Not related to this PR): In terms of data validation, what about validating driver and database here, instead of in each retriever' model?

I mean having something like:

```
class RetrieverInitModel(BaseModel):
   driver: neo4j.Driver # (or Any I don't rememberà
   database: str
   # usual validation of these

class Retriever:
   def __init__(driver, database):
      try:
         m = RetrieverInitModel(driver, database)
      except ValidationError:
         raise TheCorrectException

class VectorRetriever(Retriever):
    def __init__(driver, index_name, database):
       super().__init__(driver, database)  # driver and database validation happens here
       # then we can validate the rest of the inputs
```

src/neo4j_genai/retrievers/base.py
stellasia: If just a matter of naming, I'll go for `neo4j_database` everywhere, wdyt?

src/neo4j_genai/retrievers/base.py
willtai: That would help with not duplicating code across retrievers, how about that implement in a future PR?

stellasia (approved):
LGTM

willtai:

PR #83, opened 2024-07-22
Adds a knowledge graph writer

alexthomas93:
# Description

- Adds an abstract class `KGWriter` which allows a knowledge graphs to be written to an arbitrary data store.
- Adds a `Neo4jWriter` class which allows a knowledge graph to be written to a Neo4j database.
- Adds an `upsert_vector_on relationships` function to allow for vectors to be added to Neo4j relationships.
- Adds Pydantic models for a Neo4j graph.
- Adds an end-to-end test for the `Neo4jWriter` class.
- Fixes some broken copyright headers.

## Type of Change
- [X] New feature
- [X] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [X] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_genai/indexes.py
```diff
@@ -238,10 +238,11 @@ def drop_index_if_exists(
 
 def upsert_vector(
     driver: neo4j.Driver,
-    node_id: int,
+    id: int,
```
willtai, line 241: Is there a specific reason to change the variable name?

willtai: I'm wondering whether we want to have a separate function for uploading vectors to relationships instead. More of a separation of concerns thing

src/neo4j_genai/indexes.py
```diff
@@ -238,10 +238,11 @@ def drop_index_if_exists(
 
 def upsert_vector(
```
willtai, line 239: I think we can also add tests for upsert_vector for the relationship case

src/neo4j_genai/indexes.py
alexthomas93: Just because the function now works with nodes and vectors so keeping it as `node_id` might be a bit confusing. We can split it into 2 functions if you'd like they'd just be very similar and there'd be a bit of duplicated code.

src/neo4j_genai/indexes.py
alexthomas93: Good point, I'll add tests.

src/neo4j_genai/indexes.py
willtai: To me I think it makes sense to be separate as it's much more explicit and modular, also then we don't have breaking functionality for upsert_vector

alexthomas93: I think if we do that it would make sense to re-name `upsert_vector` though, otherwise people might get confused and assume it could be used to add embeddings to relationships.

src/neo4j_genai/indexes.py
alexthomas93: Done!

src/neo4j_genai/indexes.py
alexthomas93: @willtai In order to avoid a breaking change I've left `upsert_vector` as it is other than some minor formatting changes and added an additional `upsert_vector_on_relationship` function. Let me know what you think. I was considering renaming `upsert_vector` to `upsert_vector_on_node` but this would also introduce a breaking change.

willtai: I think that is good, as long we make it clear in the docs. Could you also please update the documentation for this? I think you can add the new `upsert_vector_on_relationship` there.

Could you also double check whether the example in the `upsert_vector` docstring is correct?

willtai: Ah apologies I see that it's already added!

src/neo4j_genai/indexes.py
```diff
@@ -278,13 +278,13 @@ def upsert_vector(
...
     try:
-        query = """
-        MATCH (n)
-        WHERE elementId(n) = $id
-        WITH n
-        CALL db.create.setNodeVectorProperty(n, $embedding_property, $vector)
-        RETURN n
-        """
+        query = (
+            "MATCH (n)"
```
willtai, line 282: Why change the multiline string here?

src/neo4j_genai/indexes.py
alexthomas93: I updated the docstring also, the example in there should be good now

src/neo4j_genai/indexes.py
alexthomas93: If you use `"""` to create multiline strings any indentation in your code will be included in the string. This means when you write your tests you need to make sure the indentation in the multi-line strings in the tests matches the indentation in the parts of the code they're testing, which can make the multi-line strings in tests look a bit weird. If you instead write multi-line strings this way you don't have to worry about it

src/neo4j_genai/kg_construction/kg_writer.py
```diff
@@ -0,0 +1,144 @@
...
+from neo4j_genai.indexes import upsert_vector, upsert_vector_on_relationship
+from neo4j_genai.kg_construction.types import Neo4jGraph, Neo4jNode, Neo4jRelationship
+from neo4j_genai.neo4j_queries import UPSERT_NODE_QUERY, UPSERT_RELATIONSHIP_QUERY
+
+
+class KGWriter(Component):
+    """Abstract class used to write a knowledge graph to a data store."""
+
+    @abstractmethod
+    async def run(self, graph: Neo4jGraph) -> None:
```
stellasia, line 34: The pipeline currently is not able to instantiate a Pydantic model as component input. However, we can add the `@validate_call` decorator from pydantic, that will do the type cast for us (and raise a ValidationError if it doesn't work)

src/neo4j_genai/kg_construction/kg_writer.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        Writes the graph to a data store.
+
+        Args:
+            graph (Neo4jGraph): The knowledge graph to write to the data store.
+        """
+        self.graph = graph
```
stellasia, line 41: Is this used somewhere?

src/neo4j_genai/kg_construction/types.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class Neo4jProperty(BaseModel):
+    key: str
+    value: Any
+
+
+class Neo4jEmbeddingProperty(BaseModel):
+    key: str
+    value: str
```
stellasia, line 29: Shouldn't the type of value be `list[float]`?

src/neo4j_genai/kg_construction/kg_writer.py
alexthomas93: So only base types can be inputs to the `run` function in a `component`?

src/neo4j_genai/indexes.py
willtai: Have you tested these queries? I think you may need to add a space to the end of each line this way.

Regardless, I think using """ makes it more readable though, we had an issue previously where with the proposed way, someone missed adding a space at the end as it was not very readable. I proposed switching back to using """ but for tests use something like https://docs.python.org/3/library/textwrap.html . What do you think?

src/neo4j_genai/kg_construction/types.py
stellasia: Oh ok, I see. It makes sense to me to delegate the embedding creation to another component. This way we can probably use the same component to embed the chunks if we want to.

src/neo4j_genai/indexes.py
alexthomas93: I'm not sure how that'd work for tests such as `test_upsert_node_with_embedding`. You'd need to call textwrap within `upsert_vector` and the test to make sure the strings match. Maybe the way around this is to just move the queries to the `neo4j_queries.py` file and import them in both tests and the upsert functions.

willtai: I think if it works, maybe there's no need to overcomplicate it for now. I'd be happy with the original approach provided the query runs

src/neo4j_genai/kg_construction/types.py
alexthomas93: Have removed it. Will raise a new PR with the embedding component.

src/neo4j_genai/kg_construction/kg_writer.py
alexthomas93: I'm confused, `graph` is still a `Neo4jGraph` in your snippet. Is there an easy way we could change the pipeline to allow for non-base types to be passed as inputs to the `run` function of a `Component`? I think we need a way to ensure the input to this component conforms to a specific graph structure.

src/neo4j_genai/kg_construction/kg_writer.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Args:
+        driver (neo4j.driver): The Neo4j driver to connect to the database.
+        neo4j_database (Optional[str]): The name of the Neo4j database to write to. Defaults to 'neo4j' if not provided.
+    """
+
+    def __init__(
+        self,
+        driver: neo4j.driver,
+        neo4j_database: Optional[str] = None,
```
willtai, line 58: Should this default to "neo4j" as described in the docstring?

tests/e2e/test_kg_construction_e2e.py
```diff
@@ -0,0 +1,93 @@
...
+    relationship = Neo4jRelationship(
+        start_node_id="1", end_node_id="2", label="NEXT_CHUNK"
+    )
+    graph = Neo4jGraph(nodes=[start_node, end_node], relationships=[relationship])
+
+    neo4j_writer = Neo4jWriter(driver=driver)
+    await neo4j_writer.run(graph=graph)
+
+    query = """
+    MATCH (a:Document {id: 1})-[r:NEXT_CHUNK]-(b:Document {id: 2})
```
willtai, line 59: Is it intentional that the relationship `-[r:NEXT_CHUNK]-` is bidirectional?

src/neo4j_genai/kg_construction/kg_writer.py
alexthomas93: Yeah it works the same was as `create_vector_index`, `create_fulltext_index`, and `drop_index_if_exists`, etc. from `src/neo4j_genai/indexes.py`. If `database` is set to `None` in the `execute_query` method of the Neo4j driver it defaults to using the `neo4j` database.

tests/e2e/test_kg_construction_e2e.py
alexthomas93: No that's a mistake, I'll fix it

stellasia (approved):

PR #85, opened 2024-07-31
Add entity / Relation extraction component

stellasia:
# Description

This PR adds the Entity & Relation extraction base component and implementation using LLM, using a pretty simple prompt that can be configured (similarly to what is done for RAG)

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_genai/components/entity_relation_extractor.py
```diff
@@ -0,0 +1,141 @@
...
+class OnError(enum.Enum):
+    RAISE = "RAISE"
+    IGNORE = "CONTINUE"
+
+
+class LLMEntityRelationExtractor(EntityRelationExtractor):
+    def __init__(
+        self,
+        llm: LLMInterface,
+        prompt_template: ERExtractionTemplate | str = ERExtractionTemplate(),
```
willtai, line 34: While I understand passing in a `prompt_template` allows for full flexibility, I'm wondering why not just let users pass in a `str` instead of `ERExtractionTemplate`. I'm wondering if this makes it overly complicated for a user.

src/neo4j_genai/components/types.py
```diff
@@ -100,13 +101,13 @@ class Neo4jRelationship(BaseModel):
...
 
     Attributes:
         nodes (list[Neo4jNode]): A list of nodes in the graph.
         relationships (list[Neo4jRelationship]): A list of relationships in the graph.
     """
 
-    nodes: list[Neo4jNode]
-    relationships: list[Neo4jRelationship]
+    nodes: list[Neo4jNode] = []
+    relationships: list[Neo4jRelationship] = []
```
willtai, line 113: Do we need defaults for these attributes? I think there will be an issue as these are mutable defaults

willtai: Sorry looks like this is fine in Pydantic https://docs.pydantic.dev/latest/concepts/models/#fields-with-non-hashable-default-values

stellasia: It's useful to create the objects when there is a failure. And yes pydantic handles the mutable default properly, one of the magic parts of pydantic :)

src/neo4j_genai/components/entity_relation_extractor.py
stellasia: The input param is a `Union` between prompt template and string :) At the moment, if string is provided, the user is responsible for building the full prompt, including injecting parameters - which in practice means adding an extra (custom) component in the pipeline to add the chunk.

stellasia: I think we could have a `PromptBuilder` component at some point, so that the prompt can be tuned, but let's see how things evolve.

src/neo4j_genai/components/entity_relation_extractor.py
```diff
@@ -0,0 +1,108 @@
...
+            prompt = self.prompt_template.format(
+                text=chunk.text, schema=schema, examples=examples
+            )
+            llm_result = self.llm.invoke(prompt)
+            try:
+                result = json.loads(llm_result.content)
+            except json.JSONDecodeError:
+                if self.on_error == OnError.RAISE:
+                    logger.error(f"LLM response is not valid JSON {llm_result.content}")
+                    raise LLMGenerationError(
```
willtai, line 89: Do we need to log to the logger if we already raise `LLMGenerationError` here?

src/neo4j_genai/components/entity_relation_extractor.py
```diff
@@ -0,0 +1,108 @@
...
+                chunk_graph = Neo4jGraph(**result)
+            except ValidationError:
+                if self.on_error == OnError.RAISE:
+                    logger.error(
+                        f"LLM response has improper format {llm_result.content}"
+                    )
+                    raise LLMGenerationError(
+                        f"LLM response has improper format {llm_result.content}"
+                    )
+                chunk_graph = Neo4jGraph()
```
willtai, line 103: So if the user chooses to set `on_error` to `IGNORE`, `chunk_graph` will be empty. Do you think is likely to happen? What should a user do in this situation? Try different prompting or LLM parameters?

stellasia: At the moment, there is no way to resume a failed pipeline, so I think it could make sense to ignore errors for a specific chunk, otherwise one failing chunk out of 100 and you'll have to rerun everything. But probably we could implement some retry mechanism in the future to make the user experience better, or make `pipeline.run` method idemptent somehow.

src/neo4j_genai/components/entity_relation_extractor.py
stellasia: Right, maybe we need that log when `on_error = IGNORE` instead?

stellasia: https://github.com/neo4j/neo4j-genai-python/pull/85/commits/947a2b002179eaf062dc9a10a5d4f2c7f3268cbb

tests/unit/components/test_entity_relation_extractor.py
```diff
@@ -0,0 +1,149 @@
...
+    chunks = TextChunks(chunks=[TextChunk(text="some text")])
+    with pytest.raises(LLMGenerationError):
+        await extractor.run(chunks=chunks)
+
+
+@pytest.mark.asyncio
+async def test_extractor_llm_badly_formatted_json_do_not_raise() -> None:
+    llm = MagicMock(spec=LLMInterface)
+    llm.invoke.return_value = LLMResponse(
+        content='{"nodes": [{"id": "0", "label": "Person", "properties": []}], "relationships": [}'
```
willtai, line 128: Is there a typo here for `"relationships": [`?

src/neo4j_genai/components/entity_relation_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+            rel.end_node_id = f"{chunk_index}:{rel.end_node_id}"
+        return graph
+
+    # TODO: fix the type of "schema" and "examples"
+    @validate_call
+    async def run(
+        self,
+        chunks: TextChunks,
+        schema: Union[BaseModel, Dict[str, Any], None] = None,
+        examples: Any = None,
```
willtai, line 76: Should example be of type `str` with default `""`?

stellasia: I was not sure about the format, but your proposal is definitely better than `Any`, let's stick to `str` for now yes

src/neo4j_genai/pipeline/component.py
```diff
@@ -41,20 +41,20 @@ def __new__(
...
+            return_model = get_type_hints(run_method).get("return")
+            # the type hint must be a subclass of DataModel
+            if not issubclass(return_model, DataModel):  # type: ignore
+                raise ValueError("The run method must return a subclass of DataModel")
+            attrs["component_outputs"] = {
+                f: {
+                    "has_default": field.is_required(),
+                    "annotation": field.annotation,
+                }
+                for f, field in return_model.model_fields.items()  # type: ignore
```
willtai, line 56: What is the reason to add `# type: ignore` here?

stellasia: This is because the `return_model` comes from `get_type_hints` which does not have annotation, hence defaulting to `Optional[Any]`, and then `mypy` complains that `Optional[Any]` does not have any `model_fields` attribute. 
In our implementation, we are forcing the `run` method to return a subclass of `DataModel`, so we are sure this attribute actually exists.

willtai: I'm trying to find a way to avoid using `type: ignore` in this case as it would defeat the purpose of having the type checker. Perhaps with either [`typing.cast`](https://docs.python.org/3/library/typing.html#typing.cast) or we can do an explicit check `if isinstance(return_model, DataModel)`. Will that work?

stellasia: I agree in general we should avoid these comments, but I'm also not convinced adding complexity in the code only for the type checker is the best option 🤔
Also, we're already performing a check on `issubclass(return_mode, DataModel)` a few lines before (with the same `type: ignore` issue btw) is there a way for `mypy` to stop complaining at least after we pass this test? I've tried to rename the variable but it doesn't change anything.

src/neo4j_genai/components/entity_relation_extractor.py
```diff
@@ -0,0 +1,114 @@
...
+from neo4j_genai.exceptions import LLMGenerationError
+from neo4j_genai.generation.prompts import ERExtractionTemplate, PromptTemplate
+from neo4j_genai.llm import LLMInterface
+from neo4j_genai.pipeline.component import Component
+
+logger = logging.getLogger(__name__)
+
+
+class EntityRelationExtractor(Component):
+    async def run(self, chunks: TextChunks, **kwargs: Any) -> Neo4jGraph:
```
alexthomas93, line 35: Should this not be an abstract method?

src/neo4j_genai/components/entity_relation_extractor.py
```diff
@@ -0,0 +1,114 @@
...
+                    raise LLMGenerationError(
+                        f"LLM response is not valid JSON {llm_result.content}"
+                    )
+                else:
+                    logger.error(
+                        f"LLM response is not valid JSON {llm_result.content} for chunk_index={chunk_index}"
+                    )
+                result = {"nodes": [], "relationships": []}
+            try:
+                chunk_graph = Neo4jGraph(**result)
```
alexthomas93, line 99: How do we connect the chunks to the extracted nodes using this approach? Do we not need to do it manually after the graph has been extracted? I'm also wondering if we might want to point the chunks to each other i.e. include a `NEXT_CHUNK` relationship which links them?

src/neo4j_genai/components/entity_relation_extractor.py
```diff
@@ -0,0 +1,114 @@
...
+        if isinstance(prompt_template, str):
+            template = PromptTemplate(prompt_template, expected_inputs=[])
+        else:
+            template = prompt_template
+        self.prompt_template = template
+        self.on_error = on_error
+
+    def update_ids(self, chunk_index: int, graph: Neo4jGraph) -> Neo4jGraph:
+        for node in graph.nodes:
+            node.id = f"{chunk_index}:{node.id}"
```
alexthomas93, line 61: Is this going to be sufficient as a unique ID? If the user runs an extractor several times and uploads to the same graph having IDs like '1:1' might cause issues. Maybe we could include something like a timestamp to ensure its always unique across multiple runs

src/neo4j_genai/components/entity_relation_extractor.py
stellasia: Absolutely: https://github.com/neo4j/neo4j-genai-python/pull/85/commits/9483d7adf7e54ccf857469ef98a8538acc67761c

src/neo4j_genai/components/entity_relation_extractor.py
stellasia: Entities have a "chunk_index" property, so that we can save them. But at the moment, chunks are not included as nodes in the graph. Do you think it should be done in this component? I'm not sure we're still in the scope of the entity and relation extraction.

src/neo4j_genai/components/entity_relation_extractor.py
stellasia: Good point! https://github.com/neo4j/neo4j-genai-python/pull/85/commits/9ea75d993f2e827bef5a216517202157c3c2480c

src/neo4j_genai/components/entity_relation_extractor.py
alexthomas93: I think Oskar mentioned we also want to be able to create lexical graphs (if that's the right terminology) which include the chunks and embeddings. If we do it would make sense to connect them to the other chunks and associated nodes so the queries for retrievers like the `VectorCypherRetriever` are easy to write. We could include another component to connect the chunks to the rest of the graph but imo it also makes sense to do it here as we're already looping through the chunks so it wouldn't be that hard to create the `NEXT_CHUNK` relationship between the chunks. Then for each chunk we could create a `chunk` node and loop through the other nodes and create a relationship which joins them.

stellasia: Done in https://github.com/neo4j/neo4j-genai-python/pull/85/commits/0c46445ca58008eda71a727af8a57df8c68e3bb2

I think we'll see in the future if we need to have a separate component, could be useful for people who might want to first do the chunk extraction, save them, and then run ER extraction for instance, but we're not there yet.

src/neo4j_genai/pipeline/component.py
willtai: We could add simple checks like this perhaps:

```python
            if isinstance(return_model, type) and not issubclass(return_model, DataModel):
                raise ValueError("The run method must return a subclass of DataModel")
            if return_model and return_model.model_fields:
                attrs["component_outputs"] = {
                    f: {
                        "has_default": field.is_required(),
                        "annotation": field.annotation,
                    }
                    for f, field in return_model.model_fields.items()
                }
```

IMO it's more of a idiot-proofing thing as well, so future developers can avoid messing things up when using `Component`

stellasia: Ok so, after a few more checks, it turns out that adding a check on `not None` (which makes totally sense because developers can forget to even use a type annotation), then the `type: ignore` comments are not needed anymore... See: https://github.com/neo4j/neo4j-genai-python/pull/85/commits/1450bbea05674e049891f02678bb0025edef595c

stellasia: So mypy is doing its job right :) 
From the return type `Optional[Any]`, we were testing for the exact type (so Any -> DataModel), but not the optional.

willtai (approved):
LGTM

PR #88, opened 2024-07-31
Add schema for kg builder

willtai:
# Description

Adds predefined schema builder for KG construction. Builds `SchemaBuilder` class as a subclass of `Component`.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


examples/pipeline/kg_builder_with_schema.py
```diff
@@ -0,0 +1,117 @@
+#  Copyright (c) "Neo4j"
```
stellasia, line 1: Is this different from the other kg_builder example?

tests/unit/kg_construction/test_schema.py
```diff
@@ -0,0 +1,234 @@
...
+from neo4j_genai.exceptions import SchemaValidationError
+from neo4j_genai.kg_construction.schema import Entity, Relation, SchemaBuilder
+from pydantic import ValidationError
+
+
+# Fixtures for test data
+@pytest.fixture
+def valid_entities() -> list[Entity]:
+    return [
+        Entity(name="PERSON", type="str", description="An individual human being."),
```
stellasia, line 27: I think only now I really understand this schema format :D It seems to be really bounds to RDF, where entities are flat (no properties).

willtai: I see, do you see a problem with this approach? Perhaps we can add "properties" to Entity and Relation ?

src/neo4j_genai/kg_construction/schema.py
```diff
@@ -0,0 +1,99 @@
...
+            for rel in rels:
+                if rel not in relations:
+                    raise SchemaValidationError(
+                        f"Relation '{rel}' is not defined in the provided relations."
+                    )
+
+        return data
+
+
+class SchemaBuilder:
```
alexthomas93, line 77: Could it be work implementing some sort of __str__ function here? I'm sure BaseModel has one but maybe we want to experiment with the string format of the schema somewhat when we're engineering the prompt for the entity extractor.

examples/pipeline/kg_builder_with_schema.py
willtai: Apologies that was an old example. I've pushed the correct one here https://github.com/neo4j/neo4j-genai-python/pull/88/commits/c8f8ba5692855bf054bc6145fdb5de2aa1ebb4f7

src/neo4j_genai/kg_construction/schema.py
```diff
@@ -0,0 +1,95 @@
...
+    name: str
+    type: str = Field(
+        ..., description="Type of the entity's name field, represented as a string."
+    )
+    description: str
+
+
+class Relation(BaseModel):
+    name: str
+    description: str
```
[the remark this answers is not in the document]
willtai: Does that assume `node <==> entity`?

src/neo4j_genai/kg_construction/schema.py
willtai: Interesting, how does that work in the overall experience. In `kg_builder_with_schema.py` the schema inputs are defined before `asyncio.run(pipe.run(pipe_inputs))` is run. Won't a `__str__` only be useful if it can run when defining the pipeline. Did I misunderstand it?

src/neo4j_genai/kg_construction/schema.py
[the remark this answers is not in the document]
willtai: https://github.com/neo4j/neo4j-genai-python/pull/88/commits/e7be9357a3f6a447d9d00e654abb569465f7755b added here

tests/unit/kg_construction/test_schema.py
willtai: https://github.com/neo4j/neo4j-genai-python/pull/88/commits/e7be9357a3f6a447d9d00e654abb569465f7755b added here

stellasia: Hey Will, sorry I've been thinking about the question quite a while, basically my concern was whether we should support ontologies AND labelled property graph (LPG) in the same object, and my conclusion is that it will be hard. WDYT?
I tend to think we should focus on LPG right now, so yes adding properties to entities and assuming the entities we extract are (for most of them) the final nodes we will store.
If we agree on this approach, I think properties should be objects, because it is useful to include a type (GPT-4o is able to normalize dates for instance, which is super useful), and a description could also help in some circumstances (but this last assertion is untested)

src/neo4j_genai/kg_construction/schema.py
stellasia: It could be useful I guess to have something similar in the `SchemaConfig` object, returned by the schema builder component. But for now the returned types are lost (we're just saving the dict), so at the moment we can't use it to serialize the schema for the prompt.

alexthomas93: Maybe this is something we want to do in the ER extractor instead in that case?

stellasia: Yes, to me that belongs to the prompt "builder" phase: you input the schema as a dict, and in the prompt you format it as you want (hence the idea of jinja2).

stellasia (approved):
Awesome!

src/neo4j_genai/kg_construction/schema.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class SchemaConfig(DataModel):
+    """
+    Represents possible relationships between entities and relations in the graph.
+    """
+
+    entities: Dict[str, Dict[str, Any]]
+    relations: Dict[str, Dict[str, Any]]
+    potential_schema: List[Tuple[str, str, str]]
```
willtai, line 85: I'm wondering whether this should be renamed to potential triplets instead? Or is potential_schema fine?

PR #92, opened 2024-08-02
Fixes and refactors the KG writer component

alexthomas93:
# Description

- Fixes bugs with the `Neo4jWriter` class.
- Removes the `Neo4jProperty` and `Neo4jEmbeddingProperty` classes.
- Refactors the `Neo4jNode`, `Neo4jRelationship`, and `Neo4jGraph` classes.
- Adds unit tests for the `Neo4jNode` class.

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [X] Unit tests
- [X] E2E tests
- [X] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_genai/components/kg_writer.py
+++ b/src/neo4j_genai/components/kg_writer.py
@@ -73,24 +73,23 @@ class Neo4jWriter(KGWriter):
             node (Neo4jNode): The node to upsert into the database.
         """
         # Create the initial node
-        properties = "{" + f"id: {node.id}"
+        parameters = {"id": node.id}
         if node.properties:
-            properties += (
-                ", " + ", ".join(f"{p.key}: {p.value}" for p in node.properties) + "}"
-            )
-        else:
-            properties += "}"
+            parameters.update(node.properties)
+        properties = (
+            "{" + ", ".join(f"{key}: ${key}" for key in parameters.keys()) + "}"
+        )
         query = UPSERT_NODE_QUERY.format(label=node.label, properties=properties)
-        result = self.driver.execute_query(query)
+        result = self.driver.execute_query(query, parameters_=parameters)
         node_id = result.records[0]["elementID(n)"]
         # Add the embedding properties to the node
         if node.embedding_properties:
-            for prop in node.embedding_properties:
+            for prop, vector in node.embedding_properties.items():
                 upsert_vector(
                     driver=self.driver,
                     node_id=node_id,
-                    embedding_property=prop.key,
-                    vector=prop.value,
+                    embedding_property=prop,
+                    vector=vector,
                     neo4j_database=self.neo4j_database,
                 )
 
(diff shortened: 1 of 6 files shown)


src/neo4j_genai/components/types.py
```diff
@@ -42,44 +42,29 @@ class TextChunks(DataModel):
...
+    properties: Optional[dict[str, Any]] = None
+    embedding_properties: Optional[dict[str, list[float]]] = None
+
+    @field_validator("properties", "embedding_properties")
+    @classmethod
+    def check_for_id_properties(
+        cls, v: Optional[dict[str, Any]]
+    ) -> Optional[dict[str, Any]]:
+        if v and "id" in v.keys():
+            raise TypeError("'id' as a property name is not allowed")
```
stellasia, line 66: Why this constraint? Could be an issue with the current implementation of the ERExtractor that asks the LLM to assign an unique ID to each extracted entity (and use this ID in the relation extraction).

stellasia: Ok, I got it, `id` is explicit, makes sense.

src/neo4j_genai/components/kg_writer.py
```diff
@@ -102,26 +101,27 @@ def _upsert_relationship(self, rel: Neo4jRelationship) -> None:
...
         # Create the initial relationship
         properties = (
-            "{" + ", ".join(f"{p.key}: {p.value}" for p in rel.properties) + "}"
+            "{" + ", ".join(f"{key}: ${key}" for key in rel.properties.keys()) + "}"
             if rel.properties
             else "{}"
         )
+        parameters = rel.properties
         query = UPSERT_RELATIONSHIP_QUERY.format(
             start_node_id=rel.start_node_id,
```
stellasia, line 110: Start and end node IDs could be parameters too, but I don't know if there is any risk in doing it that way.

src/neo4j_genai/components/types.py
```diff
@@ -42,44 +42,29 @@ class TextChunks(DataModel):
...
 class Neo4jNode(BaseModel):
     """Represents a Neo4j node.
 
     Attributes:
         id (str): The ID of the node.
         label (str): The label of the node.
-        properties (Optional[list[Neo4jProperty]]): A list of properties associated with the node.
-        embedding_properties (Optional[list[Neo4jEmbeddingProperty]]): A list of embedding properties associated with the node.
+        properties (Optional[dict[str, Any]]): A dictionary of properties attached to the node.
+        embedding_properties (Optional[dict[str, list[float]]]): A list of embedding properties attached to the node.
```
stellasia, line 52: Not for now, but I'm wondering if we could detect embeddings if they are stored in the global property list, since there must be an index in the database for them?

src/neo4j_genai/components/kg_writer.py
alexthomas93: In this context the parameters are just the properties saved to the relationship. Why would we need to include the start and end node IDs here also?

src/neo4j_genai/components/types.py
alexthomas93: Yeah maybe we could include a call to the DB which checks which properties are indexed before uploading anything. I guess the disadvantage there is that I believe you can upload the properties to the nodes then create the vector index, and that would disallow that workflow.

src/neo4j_genai/components/kg_writer.py
stellasia: They're used in the query later on (if I'm not mistaken): 
```
MATCH (start {{ id: '{start_node_id}' }}), (end {{ id: '{end_node_id}' }})
```
which could be written:
```
MATCH (start {{ id: $start_node_id }}), (end {{ id: $end_node_id }})
```

stellasia (approved):
Thanks!!

PR #93, opened 2024-08-05
Resolve issue with GitHub Actions runner running out of disk space

willtai:
# Description

- For E2E tests, prune docker images before installing dependencies to reduce disk file usage
- Adds a new workflow for running E2E tests on main every 3 hours
- Reduces scope (Python versions and Neo4j community Docker image) of PR E2E tests 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/.github/workflows/pr-e2e-tests.yaml
+++ b/.github/workflows/pr-e2e-tests.yaml
@@ -1,10 +1,6 @@
 name: 'Neo4j-GenAI PR E2E Tests'
 
 on:
-  push:
-    branches:
-      - main
-      - fix-out-of-space-error-on-gh-actions
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
--- a/.github/workflows/pr-e2e-tests.yaml
+++ b/.github/workflows/pr-e2e-tests.yaml
@@ -18,10 +18,8 @@ jobs:
   e2e-tests:
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 1
       matrix:
-        python-version: ['3.8', '3.12']
-#        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
         neo4j-version:
           - 5
         neo4j-edition:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-e2e-tests.yaml
```diff
@@ -1,6 +1,10 @@
 name: 'Neo4j-GenAI PR E2E Tests'
 
 on:
+  push:
+    branches:
+      - main
+      - fix-out-of-space-error-on-gh-actions
```
willtai, line 7: I will remove this after the review. This is to test that this workflow works when I push to my working branch

.github/workflows/scheduled-e2e-tests.yaml
```diff
@@ -0,0 +1,92 @@
+name: 'Neo4j-GenAI Scheduled E2E Tests'
+
+on:
+  schedule:
+    - cron:  '0 */3 * * *'  # Runs every 3 hours
```
oskarhane, line 5:
```suggestion
    - cron:  '0 6,9,12,15,18 * * 1-5'  # Runs every 3 hours daytime on working days
```

.github/workflows/scheduled-e2e-tests.yaml
```diff
@@ -0,0 +1,92 @@
+name: 'Neo4j-GenAI Scheduled E2E Tests'
+
+on:
+  schedule:
+    - cron:  '0 6,9,12,15,18 * * 1-5'  # Runs every 3 hours daytime on working days
```
stellasia, line 5: Do we really need a cron job? Can't we just run it every time there is a push to the main branch?

willtai: We could, but it won't run if there hasn't been a push to main in a long time, which is why I have both. Is there a concern with doing this every 3 hours?

stellasia: Sorry I'm confused, what are we testing if there is no change? The latest versions of the dependencies?

willtai: > The latest versions of the dependencies

Yes especially external integrations such as Pinecone and Weaviate. Do you think this will be likely to change?

I don't see a downside to having this run periodically as the job will likely take about 10 minutes.

stellasia: Ok, I see now, sounds good. Thanks!

alexthomas93: Do we not fix the versions of those though?

PR #113, opened 2024-08-29
Add more unit tests for pipeline and orchestrator

stellasia:
# Description

Adding more unit tests for pipelines, especially regarding input dict construction from pipeline input data and previous components. 

In order to make testing easier, the back reference to the pipeline in the task has been removed (which was probably a bad design in the first place IMHO (I know I did that, but that's why I can say it was a very bad idea :D )). It means that the orchestrator gets more responsibilities, especially it is now responsible for building the input dict for each components (on top of its previous duty to find the next task to run) (This input building step was previously part of the task itself, but the task doesn't have to know about the surroundings, it takes inputs, and generate outputs, that's it)

Note that this doesn't change the API, it's only about internal reorganization.

> **Note**
>
>
>

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


tests/unit/experimental/pipeline/test_orchestrator.py
```diff
@@ -16,13 +16,76 @@
...
+    pipe.add_component(ComponentPassThrough(), "a")
+    pipe.add_component(ComponentPassThrough(), "b")
+    pipe.connect("a", "b", input_config={"value": "a.result"})
+
+    # component "a" already run and results stored:
+    pipe._store.add("a", {"result": "a_result"})
+
+    orchestrator = Orchestrator(pipe)
+    data = orchestrator.get_component_inputs(
+        "b", {"value": "a.result"}, {"a": {"value": "text"}}
```
willtai, line 44: What does the argument `{"a": {"value": "text"}}` specify here?

stellasia: This is the pipeline.run input (it's called `data` within the code base, not very explicit I agree)

src/neo4j_genai/experimental/pipeline/pipeline.py
```diff
@@ -306,22 +242,79 @@ async def next(
...
+        # make sure dependencies are satisfied
+        # and save the inputs defs that needs to be propagated from parent components
+        for prev_edge in self.pipeline.previous_edges(task.name):
+            prev_node = self.pipeline.get_node_by_name(prev_edge.start)
+            prev_status = await prev_node.read_status()
+            if prev_status != RunStatus.DONE:
+                logger.critical(f"Missing dependency {prev_edge.start}")
+                raise PipelineMissingDependencyError(f"{prev_edge.start} not ready")
+            if prev_edge.data:
+                prev_edge_data = prev_edge.data.get("input_config") or {}
```
willtai, line 267: Super minor but you can specify a default on `.get()` 
```suggestion
                prev_edge_data = prev_edge.data.get("input_config", {})
```

tests/unit/experimental/pipeline/test_orchestrator.py
```diff
@@ -16,13 +16,76 @@
...
+    )
+    assert data == {"value": "a_result"}
```
willtai, line 46: I think I'm confused now, is this not meant to return `"text"`?

tests/unit/experimental/pipeline/test_orchestrator.py
```diff
@@ -16,13 +16,76 @@
...
+
+
+def test_pipeline_get_component_inputs_from_parent_specific() -> None:
+    pipe = Pipeline()
+    pipe.add_component(ComponentPassThrough(), "a")
+    pipe.add_component(ComponentPassThrough(), "b")
+    pipe.connect("a", "b", input_config={"value": "a.result"})
+
+    # component "a" already run and results stored:
+    pipe._store.add("a", {"result": "a_result"})
```
willtai, line 40: Is it worth renaming `a_result` to something else? I think I'm getting confused why `a_result` should be returned instead of `"text"`

tests/unit/experimental/pipeline/test_orchestrator.py
stellasia: We're asking for the inputs for component "b", which must be the output of component "a".

tests/unit/experimental/pipeline/test_orchestrator.py
stellasia: Component "a" takes "text" as input and outputs "a_result". Component "b" takes the output of "a" as input, so "a_result" is the input for "b". Does that make sense?
I'm happy with any suggestion for a better name for "a_result" if you think it's less confusing :)

tests/unit/experimental/pipeline/test_orchestrator.py
willtai: So does "text" not matter in this test?

stellasia: Indeed, it is not used. Would you prefer if it was deleted? I think it makes sense to test we're not messing up things when we have multiple components with same input names, but maybe there are other ways to test that?

stellasia: Indeed it's not used. I can remove it. I think it's good to test that we're not messing up with parameters when we have the same names in different locations, but I'll think about a way to test that explicitely.

tests/unit/experimental/pipeline/test_orchestrator.py
willtai: I see. Perhaps replace `a_result` with `a_output` and replace `text` with `a_input`?

willtai: I think if I were to reread this in future, I will get confused between the purpose of `a_result`, `a.result`, and `text`

tests/unit/experimental/pipeline/test_orchestrator.py
stellasia: Added a new test [here](https://github.com/neo4j/neo4j-genai-python/pull/113/commits/6a026e1c802e7fe2177e00ba0eba4404d91ba37f#diff-a8ed923705bb0b915fbd58a296fc07729642c6d48b4d1e6157a6cccf02837705R103)

tests/unit/experimental/pipeline/test_orchestrator.py
stellasia: Done, used a longish but meaningful string "output from component XX".

src/neo4j_genai/experimental/pipeline/pipeline.py
```diff
@@ -59,26 +61,28 @@ class RunResult(BaseModel):
     timestamp: datetime = Field(default_factory=datetime.utcnow)
 
 
+# Type used for annotating the partial on_task_complete parameter
+# and prevent type checking error when using kwargs
+class PartialOnTaskCompleteProtocol(Protocol):
+    async def __call__(self, task: TaskPipelineNode, res: RunResult) -> None: ...
```
alexthomas93, line 67: Is this complete?

stellasia: Yes, no implementation is needed for this method, it's really just used to define a "callable type" with proper kwargs.

alexthomas93 (approved):
LGTM

PR #115, opened 2024-08-30
Fix bug in Text2cypher custom prompt option

a-s-g93:
# Description

> **Note**
>
>
>

## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: LOW

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -140,7 +138,9 @@ def get_search_results(
                 query=validated_data.query_text,
             )
         else:
-            prompt = self.custom_prompt
+            prompt = self.custom_prompt.format(
```
stellasia, line 141: If parameters are expected in the custom prompt, do you think we can use `Text2CypherTemplate`? We can instantiate such a template with a custom prompt as well. Something like:
```
prompt_template = Text2CypherTemplate(template=custom_prompt)
```

That being said, the prompt template doesn't support at the moment the checks you implemented (the required `{query}` placeholder), but it makes sense to enforce it here, and having it in the base `PromptTemplate` class would make that feature available for the other types of prompts we have (for RAG for instance). What do you think?

a-s-g93: I think that makes sense to use the Template classes and extend the checks to other prompts as well. Would you like me to make the necessary changes on this branch?

stellasia: If you are comfortable with it, you can go ahead and implement it. Feel free to reach out of you have any question.

a-s-g93: No problem, I'll make the changes

a-s-g93: I changed all Template classes to contain a `query_text` placeholder and validate upon construction. The `format` methods of each Template class also now take `query_text` as an arg, but allow for the old input arg as well. A warning will occur if using the old input arg. I refactored tests to adhere to the argument changes.

src/neo4j_genai/generation/prompts.py
```diff
@@ -39,6 +42,10 @@ def __init__(
         self.template = template or self.DEFAULT_TEMPLATE
         self.expected_inputs = expected_inputs or self.EXPECTED_INPUTS
 
+        assert (
```
stellasia, line 45: Since we have the `expected_inputs`, can we use this list instead of a hard coded `query_text`? I'm thinking about something like this:

```
for e in self.expected_inputs:
    if f"{{{e}}}" not in self.template:
        raise SomeError
```

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -132,15 +130,18 @@ def get_search_results(
         except ValidationError as e:
             raise SearchValidationError(e.errors()) from e
 
-        if not self.custom_prompt:
-            prompt_template = Text2CypherTemplate()
+        prompt_template = Text2CypherTemplate(template=self.custom_prompt)
+
+        if self.custom_prompt is None:
```
stellasia, line 135: This if/else block should not be required anymore, isn't it? It's not an issue if have more parameters than placeholders in the Python `format` method, and we can maybe add `kwargs` to each template's `format` method to prevent errors.

src/neo4j_genai/types.py
```diff
@@ -241,3 +241,12 @@ class Text2CypherRetrieverModel(BaseModel):
     examples: Optional[list[str]] = None
     result_formatter: Optional[Callable[[neo4j.Record], RetrieverResultItem]] = None
     custom_prompt: Optional[str] = None
+
+    @field_validator("custom_prompt")
```
stellasia, line 245: I'd say this validation is now deferred to the prompt template class, so we can remove it from here?

tests/unit/retrievers/test_text2cypher.py
```diff
@@ -250,3 +256,30 @@ def test_t2c_retriever_invalid_custom_prompt_type(
...
+    with caplog.at_level(logging.DEBUG):
+        retriever = Text2CypherRetriever(driver=driver, llm=llm, custom_prompt=prompt)
+        driver.execute_query.return_value = (
+            [neo4j_record],
+            None,
+            None,
+        )
+        retriever.search(query_text=query, examples=examples)
+
+    assert (
```
stellasia, line 282: Generally speaking, I'm not sure we really care about what's in the logs. However you're right we need to test that the correct prompt was used. Since we have a mocked `llm` object, we can test that it was called with the correct prompt instead, something like this:
```
llm.invoke.assert_called_once_with("This is a custom prompt. test ['example A', 'example B']")
```

stellasia (commented):
Thanks @a-s-g93 , it's an important and super useful update to our poor `PromptTemplate` class. I've left a few comments, let me know what you think.

src/neo4j_genai/generation/prompts.py
a-s-g93: We could create a `PromptMissingPlaceholderError` and run this check upon construction? I don't think we should expect users to provide a list of expected inputs though, so for custom prompts I think we can only check for the `query_text` placeholder.

src/neo4j_genai/retrievers/text2cypher.py
a-s-g93: we can remove the `schema` and `examples` args, and use kwargs to cover everything besides the `query_text` arg. Then we could remove the if/else block.

a-s-g93:
I've made the changes you requested @stellasia, but kept the `custom_prompt` validation in the `Text2CypherRetrieverModel` class

src/neo4j_genai/generation/prompts.py
```diff
@@ -28,8 +35,8 @@ class PromptTemplate:
     missing, a `PromptMissingInputError` is raised.
     """
 
-    DEFAULT_TEMPLATE: str = ""
-    EXPECTED_INPUTS: list[str] = []
+    DEFAULT_TEMPLATE: str = "{query_text}"
```
stellasia, line 38: I'm not sure we need these defaults since we are testing against all the items in `expected_inputs`, do we?

src/neo4j_genai/generation/prompts.py
```diff
@@ -108,17 +121,42 @@ class Text2CypherTemplate(PromptTemplate):
...
+    EXPECTED_INPUTS = ["query_text"]
 
-    def format(self, query: str, schema: str, examples: str) -> str:
-        return super().format(query=query, schema=schema, examples=examples)
+    def format(
+        self,
+        schema: Optional[str] = None,
+        examples: Optional[str] = None,
+        query_text: str = "",
+        query: Optional[str] = None,
```
stellasia, line 139: Here as well, I think we can still use the `query` name, is there a reason to rename it?

src/neo4j_genai/generation/prompts.py
```diff
@@ -147,9 +185,32 @@ class ERExtractionTemplate(PromptTemplate):
...
+    EXPECTED_INPUTS = ["query_text"]
 
-    def format(self, text: str, schema: dict[str, Any], examples: str) -> str:
-        return super().format(text=text, schema=schema, examples=examples)
+    def format(
+        self,
+        schema: dict[str, Any],
+        examples: str,
+        query_text: str = "",
+        text: Optional[str] = None,
```
stellasia, line 197: Same here, why using `query_text` instead of `text`?

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -132,15 +132,22 @@ def get_search_results(
...
-            prompt_template = Text2CypherTemplate()
-            prompt = prompt_template.format(
-                schema=self.neo4j_schema,
-                examples="\n".join(self.examples) if self.examples else "",
-                query=validated_data.query_text,
-            )
-        else:
-            prompt = self.custom_prompt
+        # parse the schema and examples inputs
+        examples_to_use = kwargs.get("examples") or (
```
stellasia, line 136: I think there was a reason to have examples and schema in the init and not in the search, @alexthomas93 maybe knows more about it?

src/neo4j_genai/types.py
stellasia: I understand the concern, though baving it here looks a bit weird to me. Do you think we could have a validation method in the `PromptTemplate`, that could get called when we instantiate the template?

src/neo4j_genai/generation/prompts.py
a-s-g93: you're right, tests still pass when I remove the `query_text` and use the original empty defaults

src/neo4j_genai/generation/prompts.py
a-s-g93: renaming was for constancy between all the prompts.

src/neo4j_genai/retrievers/text2cypher.py
a-s-g93: I like having the `examples` option in the search for the custom prompt because it allows for dynamic example selection.

a-s-g93:
@stellasia What else needs to be complete for this to be accepted? I'm working on a project that is dependent on these changes.

src/neo4j_genai/retrievers/text2cypher.py
stellasia: Ok, I see the point, thanks.

src/neo4j_genai/retrievers/text2cypher.py
```diff
@@ -103,22 +103,22 @@ def __init__(
...
             error_message = getattr(e, "message", str(e))
             raise SchemaFetchError(
                 f"Failed to fetch schema for Text2CypherRetriever: {error_message}"
             ) from e
 
-    def get_search_results(
-        self,
-        query_text: str,
-    ) -> RawSearchResult:
+    def get_search_results(self, query_text: str, **kwargs: Any) -> RawSearchResult:
```
stellasia, line 115: Can we use a more explicit parameter instead of kwargs? I think it would make our lifes easier if in the future we want to add parameters to this method (and it will prevent having reserved parameter names and so on). So something like:
```
def get_search_results(self, query_text: str, prompt_params: Optional[dict[str, Any]] = None) -> RawSearchResult:
```

src/neo4j_genai/generation/prompts.py
stellasia, line 197: After discussing with the team, we see no real reason to rename this parameter at this stage. If that's important to you, may I suggest you open another PR with this change so that we can discuss this specifically? Thank.

src/neo4j_genai/generation/prompts.py
stellasia: We agree on this one, just see comment below about try/except block.

stellasia (changes requested):

src/neo4j_genai/generation/prompts.py
```diff
@@ -108,17 +121,42 @@ class Text2CypherTemplate(PromptTemplate):
...
+                        stacklevel=2,
+                    )
+                elif isinstance(query, str):
+                    warnings.warn(
+                        "'query' is deprecated and will be removed in a future version, please use 'query_text' instead.",
+                        DeprecationWarning,
+                        stacklevel=2,
+                    )
+                    query_text = query
+        except ValidationError as e:
```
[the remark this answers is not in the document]
a-s-g93: ok I'm removing the try/except, but keeping the warning conditionals

src/neo4j_genai/generation/prompts.py
a-s-g93: That's ok. I'll change this back.

src/neo4j_genai/generation/prompts.py
```diff
@@ -29,7 +33,7 @@ class PromptTemplate:
     """
 
     DEFAULT_TEMPLATE: str = ""
-    EXPECTED_INPUTS: list[str] = []
+    EXPECTED_INPUTS: list[str] = list()
```
stellasia, line 36: For my culture, what's the difference here between `[]` and `list()`?

src/neo4j_genai/generation/prompts.py
```diff
@@ -151,5 +184,10 @@ class ERExtractionTemplate(PromptTemplate):
 """
     EXPECTED_INPUTS = ["text"]
 
-    def format(self, text: str, schema: dict[str, Any], examples: str) -> str:
+    def format(
+        self,
+        schema: dict[str, Any],
+        examples: str,
+        text: str = "",
```
stellasia, line 191: Maybe a detail, but `text` is not optional anymore, right?

stellasia (approved):
LGTM, thanks for this nice improvement! 

We're in the process of renaming the package, so we might need to wait a little bit more before we can merge it, I will come back to you very soon with some updates.

src/neo4j_genai/generation/prompts.py
a-s-g93: `[]` is a little faster than `list()`, but `list()` is just more explicit. I suppose it's an opinion, but also for `{}` I prefer to explicitly use `dict()` or `set()` for clarity.

src/neo4j_genai/generation/prompts.py
a-s-g93: Nope it's not optional, i can remove the default if you'd like.

a-s-g93:
I had some issues during rebase, but all tests seem to be passing now.

CHANGELOG.md
```diff
@@ -1,64 +1,32 @@
-# @neo4j/neo4j-graphrag-python
+# @neo4j/neo4j-genai-python
 
 ## Next
-
-## 0.6.2
-
-## 0.6.1
```
stellasia, line 7: I think you need to take the changelog from main up to version 0.6.2.

docs/source/api.rst
```diff
@@ -12,56 +12,56 @@ Components
 KGWriter
 ========
 
-.. autoclass:: neo4j_graphrag.experimental.components.kg_writer.KGWriter
```
stellasia, line 15: Revert these changes please.

docs/source/index.rst
```diff
@@ -1,12 +1,12 @@
-.. neo4j-graphrag-python documentation master file, created by
+.. neo4j-genai-python documentation master file, created by
    sphinx-quickstart on Tue Apr  9 16:36:43 2024.
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Neo4j GraphRAG for Python
```
stellasia, line 6: Same here, wrong version of file.

docs/source/types.rst
```diff
@@ -7,78 +7,78 @@ Types
 RawSearchResult
 ===============
 
-.. autoclass:: neo4j_graphrag.types.RawSearchResult
```
stellasia, line 10: Also here.

docs/source/user_guide_kg_builder.rst
```diff
@@ -34,7 +34,7 @@ A Knowledge Graph (KG) construction pipeline requires a few components:
 This package contains the interface and implementations for each of these components, which are detailed in the following sections.
 
 To see an end-to-end example of a Knowledge Graph construction pipeline,
-refer to `this example <https://github.com/neo4j/neo4j-graphrag-python/blob/main/examples/pipeline/kg_builder.py>`_.
```
stellasia, line 37: Also here.

stellasia (changes requested):
There are many places where we are changing `neo4j_graphrag` back to `neo4j_genai`, such as in documentation files, examples, README... I have left a few comments but there are many more. Can you check again please?

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## Next
 
+### Changed
+- `GraphRAG.search` method first parameter has been renamed `query_text` (was `query`) for consistency with the retrievers interface.
```
stellasia, line 6: This is probably a duplicate, it was done in a previous version.

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
+
+### Added 
+- Add `template` validation in `PromptTemplate` class upon construction.
+- `custom_prompt` arg is now converted to `Text2CypherTemplate` class within the `Text2CypherRetriever.get_search_results` method.
+- All prompt templates now require `query_text` arg and will error if it is not present. Previous args (`query`, `text`) may be used, but will warn of deprecation.
```
stellasia, line 11: The only affected prompt template is `RagTemplate` now.

stellasia (changes requested):
We're almost there!

stellasia (approved):
Awesome, thank you! Will merge by the end of the day.

PR #116, opened 2024-09-02
Save Document node in lexical graph

stellasia:
# Description

Save the document as a node in the lexical graph. Create relationships between this node and each processed chunk.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_genai/experimental/components/types.py
```diff
@@ -30,6 +30,7 @@ class TextChunk(BaseModel):
     """
 
     text: str
+    index: int
```
alexthomas93, line 33: Why create an index parameter and not just include it in the metadata?

src/neo4j_genai/experimental/components/text_splitters/llamaindex.py
```diff
@@ -54,8 +54,9 @@ async def run(self, text: str) -> TextChunks:
         Returns:
             TextChunks: The text split into chunks.
         """
+        chunks = self.text_splitter.split_text(text)
         return TextChunks(
             chunks=[
-                TextChunk(text=chunk) for chunk in self.text_splitter.split_text(text)
+                TextChunk(text=chunk, index=index) for index, chunk in enumerate(chunks)
```
alexthomas93, line 60: This might be a minor point but do you know for sure that the `split_text` method always returns chunks in the correct order for all LangChain and LlamaIndex splitters? I think LlamaIndex splitters have a `split_text_metadata_aware` method which includes an index which could be safer to use, and LangChain splitters have a similar method.

src/neo4j_genai/experimental/components/entity_relation_extractor.py
```diff
@@ -139,30 +143,42 @@ def __init__(
...
             node.properties.update({"chunk_index": chunk_index})
         for rel in graph.relationships:
-            rel.start_node_id = f"{self._id_prefix}:{chunk_index}:{rel.start_node_id}"
-            rel.end_node_id = f"{self._id_prefix}:{chunk_index}:{rel.end_node_id}"
+            rel.start_node_id = f"{prefix}:{rel.start_node_id}"
+            rel.end_node_id = f"{prefix}:{rel.end_node_id}"
         return graph
 
+
+class LexicalGraphBuilder:
```
alexthomas93, line 176: Why create a class for this if it's mostly just a collection of useful methods?

src/neo4j_genai/experimental/pipeline/pipeline.py
```diff
@@ -126,6 +126,7 @@ async def execute(self, **kwargs: Any) -> RunResult | None:
         except PipelineStatusUpdateError:
             logger.info(f"Component {self.name} already running or done {self.status}")
             return None
+        print(self.name, kwargs)
```
alexthomas93, line 129: Is this meant to be here?

src/neo4j_genai/experimental/components/pdf_loader.py
```diff
@@ -26,17 +28,30 @@
 from neo4j_genai.experimental.pipeline import Component, DataModel
 
 
+class DocumentInfo(DataModel):
+    path: str
+    metadata: Optional[Dict[str, str]] = None
```
alexthomas93, line 33: This is a bit of a meta-point not related directly to this review but are all of the small Pydantic classes like this really needed? We have a lot of them in the package now and they seem a bit unnecessary in some instances, especially when they're just internal to the package and aren't used for validating user input. I think type hints would be enough in those scenarios

src/neo4j_genai/experimental/components/types.py
stellasia: I included it here instead of passing it along with the chunk for most methods of the `ERExtractor`. Which metadata are you talking about here?

src/neo4j_genai/experimental/components/text_splitters/llamaindex.py
stellasia: Good point, I'll check what's in LC/LI, it's important to have the right chunk order.

src/neo4j_genai/experimental/components/entity_relation_extractor.py
stellasia: Yes, what do you suggest? It used to be in the `EntityRelationExtractor` base class but it was forcing us to instantiate an abstract class to test it.

src/neo4j_genai/experimental/components/pdf_loader.py
stellasia: In this specific case, we use it to validate the parameter propagation defined in the `Pipeline.connect` method. But in general this is true, we can rethink how we do this.

src/neo4j_genai/experimental/components/types.py
alexthomas93: The `TextChunk` metadata. I think previously we just included the index as a key in the metadata. Now it's its own parameter. Was just wondering why you changed it? It's not a big deal though, makes just as much sense to have it as its own parameter tbh

src/neo4j_genai/experimental/components/text_splitters/llamaindex.py
alexthomas93: Yeah I'm pretty sure we're fine here. I would just double check as it's been a while since I looked into the text splitter classes for both libraries and I could be wrong

src/neo4j_genai/experimental/components/entity_relation_extractor.py
alexthomas93: You can't run the unit tests on these methods without moving them to a separate class?

src/neo4j_genai/experimental/components/types.py
stellasia: Oh, now I see, sorry. I'd still keep it in a separate property because it is now required and would raise errors in the following components if not there.

src/neo4j_genai/experimental/components/text_splitters/llamaindex.py
stellasia: I think we can assume it's OK. All the other methods I've seen are actually calling the `split_text` method and returning the same results with different format but without reordering of any kind.

alexthomas93 (approved):
LGTM

PR #119, opened 2024-09-04
Make pipeline work with parallel runs

stellasia:
# Description

## Initial issue

The component's results were saved using the component name as key, and the component status were saved in the component directly, which makes the following not behaving as expected (and returning the same result for both 'runs'):

```
    pipe = Pipeline()
    pipe.add_component(ComponentAdd(), "add")
    run_params = [[1, 20], [10, 2]]
    runs = []
    for a, b in run_params:
        runs.append(pipe.run({"add": {"number1": a, "number2": b}}))
    results = await asyncio.gather(*runs)
    print(result)
```

(note: this code has been turned into a unit test)

## What's in this PR?

- This PR introduces a `run_id` (uuid) in the `Orchestrator` and this run ID is used to build the key to access the components' status and results.

- By doing so, it appeared that the `callback` was not needed anymore and was removed, all the processing now happen in the `Orchestrator.run_task` method that can call its own `on_task_complete` method without need for partial and protocol and such.

- In order to make the intermediate results still available after a run is done, this PR introduces the `ResultStore`, that's able to find the results for a given run ID and a given component. Only InMemoryStore implemented so far. `Pipeline.run` method also return the `run_id` in a new `PipelineResult` object (breaking change).

- Not mandatory but [possible follow-up PR](https://github.com/stellasia/neo4j-genai-python/pull/3), that moves the status to the `store` instead of having them hanging in the `Task`.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist


examples/pipeline/rag.py
```diff
@@ -54,7 +55,7 @@ class PromptTemplateComponent(Component):
     def __init__(self, prompt: PromptTemplate) -> None:
         self.prompt = prompt
 
-    async def run(self, query: str, context: list[str]) -> StringDataModel:
+    async def run(self, query: str, context: List[str]) -> StringDataModel:
```
willtai, line 58: I'm not a fan of naming things after their type. Perhaps `RunOutput` or something else here?

examples/pipeline/rag.py
```diff
@@ -111,6 +112,6 @@ async def run(self, prompt: str) -> StringDataModel:
     result = asyncio.run(
         pipe.run({"retrieve": {"query": query}, "augment": {"query": query}})
     )
-    print(result["generate"]["result"])
+    print(result.result["generate"]["result"])
```
willtai, line 115: Where does generate and come from here again? IMO this syntax can get confusing. Perhaps we can make it clearer with better naming? e.g.

```python
print(pipe_output.result["generate"]["result"])
```

src/neo4j_genai/experimental/pipeline/pipeline.py
```diff
@@ -125,35 +118,22 @@ async def execute(self, **kwargs: Any) -> RunResult | None:
...
-        self.status = RunStatus.SCHEDULED
-
-    async def run(
-        self, inputs: dict[str, Any], callback: PartialOnTaskCompleteProtocol
-    ) -> None:
+    async def run(self, inputs: dict[str, Any]) -> RunResult | None:
         """Main method to execute the task."""
         logger.debug(f"TASK START {self.name=} {inputs=}")
         res = await self.execute(**inputs)
         if res is None:
```
willtai, line 133: In what cases will `res` be `None` again?

examples/pipeline/rag.py
stellasia: `generate` is the name of the component (at the moment, we include in the returned result all the output from the leaf components) and `result` is a field from the output model from that component.

src/neo4j_genai/experimental/pipeline/pipeline.py
stellasia: Right, it doesn't happen anymore. It used to be when the task `run` is called but it was already started in another process. But this check is now done in the orchestrator. So removing this `if`, good catch :)

examples/pipeline/rag.py
stellasia: Done in https://github.com/neo4j/neo4j-genai-python/pull/119/commits/11a47b7c21c836a729cce2d26b7752efd1fc1484

examples/pipeline/rag.py
```diff
@@ -37,35 +39,37 @@
...
-        return StringDataModel(result="\n".join(c.content for c in res.items))
+        return ComponentResultDataModel(text="\n".join(c.content for c in res.items))
 
 
 class PromptTemplateComponent(Component):
     def __init__(self, prompt: PromptTemplate) -> None:
         self.prompt = prompt
 
-    async def run(self, query: str, context: list[str]) -> StringDataModel:
+    async def run(self, query: str, context: List[str]) -> ComponentResultDataModel:
```
willtai, line 61: Is it possible to use `list[str]` here?

tests/unit/experimental/pipeline/test_orchestrator.py
```diff
@@ -181,8 +189,8 @@ async def test_orchestrator_aggregation(pipeline_aggregation: Pipeline) -> None:
 async def test_orchestrator_aggregation_waiting(pipeline_aggregation: Pipeline) -> None:
     orchestrator = Orchestrator(pipeline=pipeline_aggregation)
     node_a = pipeline_aggregation.get_node_by_name("a")
-    node_a.status = RunStatus.DONE
+    node_a.status = {orchestrator.run_id: RunStatus.DONE}
     node_b = pipeline_aggregation.get_node_by_name("a")
-    node_b.status = RunStatus.UNKNOWN
+    node_b.status = {orchestrator.run_id: RunStatus.UNKNOWN}
```
willtai, line 194: When does a `run_id` get set to `RunStatus.UNKNOWN`?

examples/pipeline/rag.py
stellasia: It doesn't work with `python 3.8` when we do the introspection to find the expected inputs.

tests/unit/experimental/pipeline/test_orchestrator.py
stellasia: Not needed indeed, removing.

examples/pipeline/rag.py
willtai: Dumb question: is there a reason why `generate` and `result` need to have a dictionary data structure here? I guess the component name could change, so `generate` makes sense. I'm wondering whether this could be more accessible

```python
print(pipe_output.result["generate"].result)
```

stellasia: This is because we store the results as dicts in the "stores" and hence we loose the data type. We could probably change that for the results that are returned by the pipeline (different from the component' "intermediate" results), but that would be another PR ;)

willtai: Sounds good, happy to defer that

PR #122, opened 2024-09-05
Pipeline viz

stellasia:
# Description

Add support for pipeline visualisation as image.

To make things easier and avoid to build (again) the parameter mapping dict when building the visualisation graph, a `parameter_mapping` variable is added to the `Pipeline` class. This variable is fed during the "connection validation". That means that the previous `input_config_validation` is now split into two methods:
1. Validate the "connection" (`validate_connection_parameters`), ie the input config defined in the `connect` method: this method checks that the component and parameter name actually exists. It feeds the `param_mapping` dict and the `missing_inputs` dict (list of required parameters for components that are not mapped in any `connect` and that must be provided by user). The `param_mapping` dict is reused when building the component inputs during a run, and by the visualisation methods.
2. Full validation with user input (`validate_all_parameters`): checks that the missing parameters are actually provided by users. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Medium

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_genai/experimental/pipeline/pipeline.py
```diff
@@ -428,21 +482,50 @@ def connect(
             )
         if self.is_cyclic():
             raise PipelineDefinitionError("Cyclic graph are not allowed")
+        # invalidate the pipeline if it was already validated
+        self.is_validated = False
+
+    def validate_connection_parameters(self) -> None:
```
alexthomas93, line 488: I'm wondering if this is the best name for this method. If this method just validates that the parameters passed between components internally in the pipeline are correct maybe something like `validate_internal_conn_params` might be better? The method below could be `validate_all_conn_params` then. I worry `validate_connection_parameters` makes a reader think the inputs to the pipeline are also being validated here.

src/neo4j_genai/experimental/pipeline/pipeline.py
```diff
@@ -274,27 +270,23 @@ def get_component_inputs(
 
         Args:
             component_name (str): the component/task name
-            input_config (dict[str, Any]): the input config
+            param_mapping (dict[str, Any]): the input config
```
willtai, line 273:
```suggestion
            param_mapping (dict[str, dict[str, str]]]): the input config
```

src/neo4j_genai/experimental/pipeline/pipeline.py
```diff
@@ -381,19 +386,68 @@ def show_as_dict(self) -> dict[str, Any]:
...
+    def draw(
+        self, path: str, layout: str = "dot", hide_unused_outputs: bool = True
+    ) -> Any:
+        G = self.get_pygraphviz_graph(hide_unused_outputs)
+        G.layout(layout)
+        G.draw(path)
+
+    def get_pygraphviz_graph(self, hide_unused_outputs: bool = True) -> pgv.AGraph:
+        self.validate_connection_parameters()
+        G = pgv.AGraph(strict=False, directed=True)
```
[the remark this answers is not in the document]
stellasia: Yes, similar to what `networkx` does... Not the best idea IMHO but that's the convention. Now it's used in three lines for us, we can still rename it.

src/neo4j_genai/experimental/pipeline/pipeline.py
stellasia: Wdyt about `validate_parameter_mapping`? The parameter definition is defined by the user, it's not purely internal.
Also, there is no user input to this method, so it can't validate them, but the docstring is misleading, will be fixed.

alexthomas93: I meant this method doesn't validate the inputs which are provided to components when running the pipeline (if I'm understanding it correctly) whereas `validate_all_parameters` does do this. Maybe all that's needed is to update the docstring a bit here to make it clear the difference between the 2 methods.

src/neo4j_genai/experimental/pipeline/pipeline.py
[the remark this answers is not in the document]
alexthomas93: Makes sense. Happy to keep it if it's a convention. Would probably make things a bit easier when debugging if all the examples on line use the same convention.

src/neo4j_genai/experimental/pipeline/pipeline.py
stellasia: Yes that's exactly what these methods are doing. Potentially improved naming [here](https://github.com/neo4j/neo4j-genai-python/pull/122/commits/4494799b5724282f74f4ad45abb00beae5aa1a28). If you think it's still confusing, we can go with your first proposal.

CHANGELOG.md
```diff
@@ -2,9 +2,16 @@
...
 
-## Fixed
+### Added
+- Ability to visualise pipeline with `my_pipeline.draw("pipeline.png")`
+
+### Fixed
 - Pipelines now return correct results when the same pipeline is run in parallel.
 
+### Changed
+- Pipeline run method now return a PipelineResult object.
```
stellasia, line 12: Note: this was a previous PR but should be included

alexthomas93 (approved):
LGTM!

PR #124, opened 2024-09-09
Raise error on wrong parameter mapping during pipeline definition

stellasia:
# Description

This PR improves the pipeline definition check by raising an error before the pipeline run starts if:

- the same parameter is mapped twice: for instance:
```
pipe.connect("a", "b", {"value": "a.value"})
pipe.connect("c", "b", {"value": "c.value"})
```
"value" for component "b" can only be mapped to "a" or "b", not both.

- A parameter is defined in the mapping but not expected by component, for instance:
```
class MyComponent(Component):
    def run(self, number: int) -> Result:
        pass

pipe.connect("something", "my_component", {"a": "something"})
```
"a" is not an valid input for "MyComponent".

In both cases, `PipelineDefinitionError` is raised.

Also, introduce an `invalidate` method, so that the instance attributes are cleaned properly when invalidating the pipeline (if not, when adding a component after the pipeline has already been validated, the validation would fail because the parameter already exists in the param mapping dict from the previous validation).

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist


tests/unit/experimental/pipeline/test_pipeline.py
```diff
@@ -97,6 +97,25 @@ def test_pipeline_parameter_validation_one_component_all_good() -> None:
...
+    pipe.add_component(component_b, "b")
+    pipe.connect("a", "b", {"value": "a.result"})
+    is_valid = pipe.validate_parameter_mapping_for_task(pipe.get_node_by_name("b"))
+    assert is_valid is True
+    with pytest.raises(PipelineDefinitionError):
+        pipe.validate_parameter_mapping_for_task(pipe.get_node_by_name("b"))
+    pipe.invalidate()
+    assert pipe.is_validated is False
+    assert len(pipe.param_mapping) == 0
+    assert len(pipe.missing_inputs) == 0
```
willtai, line 114: Doesn't this only test that `pipe.invalidate()` gets called? In theory we don't need these assertions here right?

stellasia: Yes could remove them

stellasia: Maybe better: https://github.com/neo4j/neo4j-graphrag-python/pull/124/commits/f228dbf7e01cf0f91a318a2772b819e8fd0d84ac

willtai (approved):

PR #138, opened 2024-09-17
Moved Embedder class to under neo4j_graphrag.embeddings

willtai:
# Description
Move the Embedder class to be under `neo4j_graphrag.embeddings` directory alongside the other custom Embedders e.g. OpenAIEmbeddings, VertexAIEmbeddings etc

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


willtai:

src/neo4j_graphrag/embeddings/openai.py
```diff
@@ -17,7 +17,7 @@
 
 from typing import Any
 
-from neo4j_graphrag.embedder import Embedder
```
stellasia, line 20: It's maybe a detail, for the other packages (`llm` and `retrievers`), the interface (base class) is in `base.py`. Is it worth following the same sort of convention?

willtai: I was thinking that but at the same time we expect users to create custom embedders so I was thinking this is more explicit. Although following this convention could make it less confusing, wdyt?

stellasia: We also expect users to create custom retrievers, right? But maybe the 'base' convention is not a good one and we should move to something more explicit everywhere?

stellasia (approved):

PR #164, opened 2024-10-01
Add pre-release actions

oskarhane:
# Description
- For prerelease for next Major `x.y.z -> (x+1).0.0a0`
- For next prerelease bump `a0 -> a1`

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -1,0 +1,46 @@
+name: Publish a new Alpha release (x.y.0a0 -> x.y.0a1) 🚀
+
+on:
+  workflow_dispatch:
+
+jobs:
+  bump-version:
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GIT_PUSH_PAT }}
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+
+      - name: Install and configure Poetry
+        uses: snok/install-poetry@v1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Get version
+        id: get-version
+        run: echo version=`poetry version -s` >> "$GITHUB_OUTPUT"
+      - name: Print version
+        run: |
+          echo Version: ${{ steps.get-version.outputs.version }}
+      - name: Update CHANGELOG.md (cross platform supported)
+        run: |
+          sed -i.bak -e 's/## Next/## Next\n\n## ${{ steps.get-version.outputs.version }}/' CHANGELOG.md && rm CHANGELOG.md.bak
```
stellasia, line 39: Maybe we should leave the 'Next' section and only rename it when we release the final version? It's gonna be tricky otherwise to rename dynamic sections.

oskarhane: Since these will be versions on pypi, I still think they should have version entries in the changelog. Once we do the GA release we can manually copy and paste from the pre-releases.
So I don't think we should change anything here.
Makes sense?

stellasia: Yeah makes sense. So we'll need to commit the changes in the CHANGELOG  between alpha and final release? I'm ok with that, let's try to keep things simple and not introduce black magic. That would require an update to the template card though, because I'm sure I'll forget that step :)

oskarhane: I've updated the release card. Would you mind approving this?

stellasia (approved):

PR #176, opened 2024-10-12
Improved schema definition in the SimpleKGPipeline

stellasia:
# Description

Adds the ability to specify some entities or relationships as dict in the SimpleKGPipeline, adding a description and property list (both optional):

```
valid_entities = [
  "Location",  # this or the next items are equivalent
  {
    "label": "Event",
  },
  {
    "label": "Game",
    "description": "A match between two persons",
  },
  {
    "label": "Organization",
    "properties": [
      {"name": "name", "type": "STRING"},
      {"name": "createdIn", "type": "INT", "description": "Creation year"},
    ]
  },
  {
    "label": "Person",
    "description": "A human being",
    "properties": [
      {"name": "name", "type": "STRING"},
      {"name": "birthDate", "type": "DATE"},
    ]
  },
]
```

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/experimental/pipeline/kg_builder.py
```diff
@@ -150,6 +162,18 @@ def __init__(
 
         self.pipeline = self._build_pipeline()
 
+    @staticmethod
+    def to_schema_entity(entity: EntityInputType) -> SchemaEntity:
+        if isinstance(entity, dict):
+            return SchemaEntity(**entity)
```
willtai, line 168: Is there a case where `entity` cannot be converted to `SchemaEntity` (and `SchemaRelation` below)?

stellasia: You're right, it will happen if the dict has a wrong format. Do you think we should raise a specific error there? The default will be the Pydantic validation error.

willtai: I think that is sufficient imo 😄

src/neo4j_graphrag/experimental/pipeline/types.py
```diff
@@ -35,3 +37,7 @@ class ConnectionConfig(BaseModel):
 class PipelineConfig(BaseModel):
     components: list[ComponentConfig]
     connections: list[ConnectionConfig]
+
+
+EntityInputType = Union[str, dict[str, Any]]
+RelationInputType = Union[str, dict[str, Any]]
```
willtai, line 43: Is there a way to avoid using `Any` here?

stellasia: I've tried to make Pydantic output the type based on the `BaseModel`, but I failed. 

The real type would be something like:
```
dict[str, Union[str, list[dict[str, str]]]
```

Assuming (key => type):

- label => str
- description => str
- properties => list[]
  - name: str
  - description: str
  - type: str 

Do you still prefer the "real" type?

willtai: Imo I think we should have that here otherwise it defeats the purpose of having Pydantic. Are there any issues with having this?

stellasia: No, actually I didn't do it because I thought it would be much more complicated, but it's still readable. Will update with a comment to explain where it comes from.

stellasia: https://github.com/neo4j/neo4j-graphrag-python/pull/176/commits/f2f467ff9f80c73d482f9162b18ea35278a45efc

willtai (approved):
Great work 🛰️

PR #195, opened 2024-10-21
Fix: AzureOpenAIEmbeddings inheritance issue

brandon-amg:
# Description

This pull request addresses issue #190. AzureOpenAIEmbeddings object is inheriting from OpenAIEmbeddings class vs. the base class (BaseOpenAIEmbeddings). As such when instantiating the AzureOpenAIEmbeddings object, you an unexpected keyword argument error.

**Example:** got an unexpected keyword argument 'azure_endpoint'

## Type of Change
- [ ] New feature
- [X ] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [ ] Project configuration change

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [X] Manual tests

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [ ] Documentation has been updated
- [ ] Unit tests have been updated
- [ ] E2E tests have been updated
- [ ] Examples have been updated
- [ ] New files have copyright header
- [x] CLA (https://neo4j.com/developer/cla/) has been signed
- [ ] CHANGELOG.md updated if appropriate

PR #175, opened 2024-10-09
Improve import system

jonbesga:
# Description

> **Note**
>
> Please provide a description of the work completed in this PR below
>
>

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [x] Project configuration change

## Complexity

> **Note**
>
> Please provide an estimated complexity of this PR of either Low, Medium or High
>
>

Complexity:

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [ ] Documentation has been updated
- [ ] Unit tests have been updated
- [ ] E2E tests have been updated
- [ ] Examples have been updated
- [ ] New files have copyright header
- [ ] CLA (https://neo4j.com/developer/cla/) has been signed
- [ ] CHANGELOG.md updated if appropriate

PR #203, opened 2024-10-24
Fix bug where AzureOpenAIEmbeddings inherits from OpenAIEmbeddings

willtai:
# Description

This PR addresses the bug in https://github.com/neo4j/neo4j-graphrag-python/pull/195

In this PR, a separate `embed_query` method is created for AzureOpenAIEmbeddings and it now inherits from BaseOpenAIEmbeddings, which would avoid the openai.OpenAI client being used.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


willtai:

src/neo4j_graphrag/embeddings/openai.py
```diff
@@ -64,7 +64,20 @@ def embed_query(self, text: str, **kwargs: Any) -> list[float]:
         return response.data[0].embedding
 
 
-class AzureOpenAIEmbeddings(OpenAIEmbeddings):
+class AzureOpenAIEmbeddings(BaseOpenAIEmbeddings):
     def __init__(self, model: str = "text-embedding-ada-002", **kwargs: Any) -> None:
         super().__init__(model, **kwargs)
         self.openai_client = openai.AzureOpenAI(**kwargs)
+
+    def embed_query(self, text: str, **kwargs: Any) -> list[float]:
```
stellasia, line 72: Shall we move this method to the base class? If it's exactly the same one for both OpenAI and AzureOpenAI

tests/unit/embeddings/test_openai_embedder.py
```diff
@@ -58,3 +58,26 @@ def test_azure_openai_embedder_happy_path(mock_openai: Mock) -> None:
...
+        mock_azure_openai_client.return_value = MagicMock()
+
+        AzureOpenAIEmbeddings(
+            model="text-embedding-ada-002",
+            azure_endpoint="https://test.openai.azure.com/",
+            api_key="my_key",
+            api_version="2023-05-15",
+        )
+
+        mock_openai_client.assert_not_called()
```
stellasia, line 78: Nice one!

src/neo4j_graphrag/embeddings/openai.py
willtai: There would be a type hint warning as `self.opena_client` isn't initialized. I tried initializing `self.open_client = None`  but 

```python
self.openai_client.embeddings.create(
            input=text, model=self.model, **kwargs
        )
```

would raise an error with mypy

stellasia: Ok I see. We can maybe use a similar approach to @jonbesga in #175 for the OpenAI LLM class: 

```
if TYPE_CHECKING:
    import openai
```

then, in the base class:
```
    client: openai.OpenAI
```

and I think AzureOpenAI inherits from OpenAI, so it mypy should be happy.

stellasia: We can also wait until the type import fixes PR is merged before harmonizing, I'm OK with that solution too.

src/neo4j_graphrag/embeddings/openai.py
```diff
@@ -53,10 +79,35 @@ def embed_query(self, text: str, **kwargs: Any) -> list[float]:
...
+
+    Args:
+        model (str): The name of the Azure OpenAI embedding model to use. Defaults to "text-embedding-ada-002".
+        kwargs: All other parameters will be passed to the openai.AzureOpenAI init.
+    """
+
+    def _initialize_client(self, **kwargs: Any) -> Any:
+        return self.openai.AzureOpenAI(**kwargs)
+
+    def embed_query(self, text: str, **kwargs: Any) -> list[float]:
```
jonbesga, line 99: The embed query is exactly the same for both clients. You can put it directly in the Base class?

willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/203/commits/4e86acbefbb530ac7bf61276a3b25ae6d402f7c2

src/neo4j_graphrag/embeddings/openai.py
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/203/commits/4e86acbefbb530ac7bf61276a3b25ae6d402f7c2

PR #220, opened 2024-11-27
Pipeline from config file

stellasia:
# Description

Adds the ability to instantiate a `Pipeline` from a config file. Used to simplify the internal implementation of `SimpleKGPipeline` (without any API change).

Supports JSON and YAML file format. Example config for SimpleKGPipeline

```yaml
    version_: 1
    template_: SimpleKGPipeline
    neo4j_config:
        params_:
            uri: bolt://
            user: neo4j
            password:
              resolver_: ENV
              var_: NEO4J_PASSWORD
    llm_config:
      class_: OpenAILLM
      params_:
        model_name: gpt-4o
        api_key:
          resolver_: ENV
          var_: OPENAI_API_KEY
        model_params:
          temperature: 0
          max_tokens: 2000
          response_format:
            type: json_object
    embedder_config:
```

The goal of this implementation was to be able to use the same objects when user writes Python code and when he is using a config file that needs to be parsed to instantiate objects such as neo4j driver and LLMs, without duplicating code.

## The main additions are

- `neo4j_graphrag.experimental.pipeline.config.runner.PipelineRunner` class, able to create a pipeline from an `AbstractPipelineConfig` (code) object or config file
- `PipelineConfigWrapper` is the 'router': depending on the `template_` field, it will instantiate either a `PipelineConfig` or `SimpleKGPipelineConfig`. In the former case, all cmponents and connections need to be defined in the config. In the later, shortcuts are possible (same interface as the `SimpleKGPipeline`).
- The other things to note:
  - `ParamToResolveConfig`: used to define parameter in config file other than providing its direct value. For instance, to read Neo4j password from env variables, user can write: 

     ```
      "password": {
          "resolver_": "ENV",
          "var_": "NEO4J_PASSWORD"
       }
       ```
    Only two "resolvers" are supported for now: 
     - Read from env 
     - Read from another key in the config file. The later is useful to reference a neo4j driver (defined in the top level key `neo4j_config`) later in the config file (in a component definition) (note that `neo4j_config` is a top level config because it can be reused in multiple components)

  - `Neo4jDriver`-, `LLM`- and `Embeddings`- `Type`: wrapper around an already instantiated object (when the config is used in code) and a `*Config` object that contains the relevant field to do the instantiation (when used from config file). The `parse` method always return an object. For instance:

      ```python
       driver_type = Neo4jDriverType(driver)
       driver = driver_type.parse()

       driver_type = Neo4jDriverType(
           Neo4jDriverConfig(
               params_={
                   "uri": "bolt://",
                   "user": "",
                   "password": "",
               }
           )
       )
       driver = driver_type.parse()
      ```

## Changes

- `SimpleKGPipeline` internals changed to use the new `PipelineRunner`.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: High

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/experimental/pipeline/config/parser.py
```diff
@@ -0,0 +1,199 @@
...
+    `model_name` and resolved `api_key`.
+
+    Args:
+        config (SimpleKGPipelineConfig): An instance of the SimpleKGPipelineConfig.
+    """
+
+    def __init__(self, config: SimpleKGPipelineConfig) -> None:
+        self.config = config
+
+    def _get_class(self, klass_path: str, optional_module: Optional[str] = None) -> Any:
```
willtai, line 67:
```suggestion
    def _get_class(self, class_path: str, optional_module: Optional[str] = None) -> Any:
```

And the same for klass below

stellasia: I've updated all the 'klass_*` names except the `klass` one because `class` is a reserved keyword.

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
```diff
@@ -0,0 +1,739 @@
...
+            for param_name, param in params.items()
+        }
+
+    @abc.abstractmethod
+    def parse(self, resolved_data: dict[str, Any] | None = None) -> Any:
+        raise NotImplementedError()
+
+
+T = TypeVar("T")
+"""Generic type to help (a bit) mypy with the return type of the parse method"""
```
willtai, line 146: Maybe clarify how it can help with mypy? Otherwise I would just remove this

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
```diff
@@ -0,0 +1,739 @@
...
+        if not issubclass(klass, self.get_interface()):
+            raise ValueError(
+                f"Invalid class '{klass}'. Expected a subclass of '{self.get_interface()}'"
+            )
+        params = self.resolve_params(self.params_)
+        try:
+            obj = klass(**params)
+        except TypeError as e:
+            raise e
+        return obj  # type: ignore[return-value]
```
willtai, line 196: Must we do `# type: ignore[return-value]` here?

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
```diff
@@ -0,0 +1,739 @@
...
+
+class Neo4jDriverConfig(ObjectConfig[neo4j.Driver]):
+    REQUIRED_PARAMS = ["uri", "user", "password"]
+
+    @field_validator("class_", mode="before")
+    @classmethod
+    def validate_class(cls, class_: Any) -> str:
+        """`class_` parameter is not used because we're always using the sync driver."""
+        if class_:
+            # logger.info("Parameter class_ is not used")
```
willtai, line 207: Should this be removed?

tests/unit/experimental/pipeline/test_kg_builder.py
```diff
@@ -248,60 +171,40 @@ async def test_knowledge_graph_builder_with_entities_and_file(_: Mock) -> None:
...
     )
 
-    internal_entities = [SchemaEntity(label=label) for label in entities]
-    internal_relations = [SchemaRelation(label=label) for label in relations]
-    assert kg_builder.entities == internal_entities
-    assert kg_builder.relations == internal_relations
-    assert kg_builder.potential_schema == potential_schema
+    # assert kg_builder.entities == entities
+    # assert kg_builder.relations == relations
+    # assert kg_builder.potential_schema == potential_schema
```
willtai, line 176: Are these still true?

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
stellasia: Yes, we have to, because we're dynamically loading a class type using `importlib`, and this returns `Any` (because we can load a function or any other variable declared in the module using the same function).

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
stellasia: I actually think it should be there (ie not commented), WDYT?

src/neo4j_graphrag/experimental/pipeline/config/config_poc_test.py
```diff
@@ -0,0 +1,248 @@
+from unittest.mock import MagicMock, Mock, patch
```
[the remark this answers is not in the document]
stellasia: Sure, moved this file and increased the test coverage in https://github.com/neo4j/neo4j-graphrag-python/pull/220/commits/a7d37e42a672cf1ef42dbb84b4909a9c9bdfcbc9 (more tests to come to cover the SimpleKGPielineConfig)

tests/unit/experimental/pipeline/test_kg_builder.py
stellasia: No, the `kg_builder` does not need these attributes anymore. Will clean up these tests very soon!

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
willtai: I would generally prefer to avoid ignoring types unless it is in some test files or special cases. Is there a way around this, maybe combining a runtime check with `typing.cast`?

```python
T = TypeVar("T", bound=object)

...

if not issubclass(obj, self.get_interface()):
        raise TypeError(...)
return cast(T, obj)
```

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
willtai: I'm unclear on how this class gets used, does it always return `"not used"`?

From the name I expected to return a `bool`

stellasia: It's a pydantic field validator, returning the validated field. In the case of the neo4j driver, we do not allow the class to be configured, that's why I think the logger could be useful. We could also have two separate base classes: one for neo4j (without the `class_` attribute) and one for the other types (LLM, embedders) with the `class_` attribute, but I was thinking it's already quite a complex structure.

src/neo4j_graphrag/experimental/pipeline/config/config_parser.py
stellasia: Ok, done something along what you propose in https://github.com/neo4j/neo4j-graphrag-python/pull/220/commits/ce21353a4b81e8315f1b77070d28c049b97572c1 (and also in https://github.com/neo4j/neo4j-graphrag-python/pull/220/commits/3122edaf35d91bbfdd8f3eed4ad67066973a5fba for `_get_class`)

The type check with `get_interface` is performed above with `issubclass`.

src/neo4j_graphrag/experimental/pipeline/config/object_config.py
```diff
@@ -0,0 +1,245 @@
...
+    """Configuration for any LLMInterface object.
+
+    By default, will try to import from `neo4j_graphrag.llm`.
+    """
+
+    DEFAULT_MODULE = "neo4j_graphrag.llm"
+    INTERFACE = LLMInterface
+
+
+class LLMType(RootModel):  # type: ignore[type-arg]
```
willtai, line 192: It's unclear to me what is `LLMType` used for - perhaps either a more explicit name or a docstring?

stellasia: All `*Type` models (`Neo4jType`, `LLMType`, `EmbedderType` and `ComponentType`) are used to deal with "things" that can be provided either as already instantiated objects (neo4j.Driver, instance of LLMInterface...) or using a config dict like `{"class_": "OpenAILLM"}`. 

The goal is to treat them in the same way, and make it easier for the classes using them. For instance, in the `AbstractPipelineConfig`, we use `LLMType`, and just call the `parse` method of that type. The `parse` method deals with parsing the config if needed.

I can add a docstring similar to the one in the `Neo4jType`?

willtai: Sounds good 👍

src/neo4j_graphrag/experimental/pipeline/kg_builder.py
```diff
@@ -120,133 +86,29 @@ def __init__(
...
-        if self.perform_entity_resolution:
-            pipe.add_component(
-                SinglePropertyExactMatchResolver(
-                    self.driver, neo4j_database=self.neo4j_database
-                ),
-                "resolver",
+            config = SimpleKGPipelineConfig(
+                # argument type are fixed in the Config object
+                llm_config=llm,  # type: ignore[arg-type]
+                neo4j_config=driver,  # type: ignore[arg-type]
```
willtai, line 93: These `type: ignore[arg-type]` can't be avoided?

stellasia: We could, but in that case that complexifies the code IMHO. The expected type is:

```
{"default": LLMType(llm)}
```

That's why we have validators in the `AbstractPipelineConfig` that transform the input to dict if it's not already one. The normal pydantic validation will introduce the `LLMType`.

src/neo4j_graphrag/experimental/pipeline/config/object_config.py
stellasia: Added comments in https://github.com/neo4j/neo4j-graphrag-python/pull/220/commits/663df47d163e03f751e6f53eac397cc84e2c3d7f (for each class and also at the top of the file for the generic information in my comment above)

willtai (approved):
Nice work!

Issue #221, opened 2024-11-28
Pipeline creates Chunks with duplicate ids when executed multiple times


risafj:
When I run the `Pipeline()` on a loop with multiple documents, a Chunk node with an id property of `":1"` and index of `1` is created for each run. This causes problems, since the ids are no longer unique.

For example, when the lexical graph gets created, a Chunk node with an id of `":1"` has a NEXT_NODE relation to every Chunk node that has an id of ":2".

After running the pipeline with 4 documents, it looks like this:

<img width="895" alt="Neo4j_Aura" src="https://github.com/user-attachments/assets/0d767ee6-6333-4dfa-9322-ed5b5588f8bf">

The same issue is occuring with FROM_CHUNK, where an entity that's supposed to have a relation like `(n:Entity)-[:FROM_CHUNK]->(c:Chunk {id: ":1", index: "1"})` actually has that relation to all documents' chunks with an index of 1.

Is there any workaround for this?
I'm guessing this issue would be solved if I could somehow pass document-specific id_prefix so each chunk gets a unique id?

https://github.com/neo4j/neo4j-graphrag-python/blob/bc6dd9c7b3f8fcfffb9ed360648ea80c6cbb17dc/src/neo4j_graphrag/experimental/components/lexical_graph.py#L78-L79 

Additional info:
I use v1.2.0.
I have a standard pipeline setup that has these components.
```python
    pipe = Pipeline()
    # skipping the config code
    pipe.add_component(text_splitter, "splitter")
    pipe.add_component(embedder, "chunk_embedder")
    pipe.add_component(schema_builder, "schema")
    pipe.add_component(extractor, "extractor")
    pipe.add_component(writer, "writer")
    pipe.add_component(resolver, "resolver")
```

stellasia:
Are you using a custom entity and relation extractor?

risafj:
Hi @stellasia ,

Thank you so much for the quick turnaround and helpful response! Your solution worked perfectly!

> Are you using a custom entity and relation extractor?

No, I'm using the one defined in this library:
```python
from neo4j_graphrag.experimental.components.entity_relation_extractor import (
    LLMEntityRelationExtractor, OnError)

extractor = LLMEntityRelationExtractor(
    llm=llm,
    on_error=OnError.RAISE,
    prompt_template=custom_prompt,
)
```

stellasia:
Hi @risafj ,

[...]

PR #222, opened 2024-11-29
Improve Document/Chunk ID management

stellasia:
# Description

Related to #221 

Some explanations: 
When running a pipeline (SimpleKGPipeline or not) over the same document multiple times, the relationships between chunks were messed up because we were recreating chunk nodes but using the same chunk ID.

What this PR does, is stop trying to build a unique ID from the chunk index, and add a unique ID to DocumentInfo and TextChunk model instead, so the unique ID are always there and truly unique.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

Complexity: High

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/llm/mistralai_llm.py
```diff
@@ -84,10 +84,10 @@ def invoke(self, input: str) -> LLMResponse:
                 messages=self.get_messages(input),
                 **self.model_params,
             )
-            if response is None or response.choices is None or not response.choices:
+            if response is None or not response.choices:
                 content = ""
             else:
-                content = response.choices[0].message.content or ""
+                content = str(response.choices[0].message.content) or ""
```
stellasia, line 90: This is to fix this mypy error: https://github.com/neo4j/neo4j-graphrag-python/actions/runs/12086043143/job/33704535789

src/neo4j_graphrag/experimental/components/pdf_loader.py
```diff
@@ -25,17 +25,8 @@
 from fsspec.implementations.local import LocalFileSystem
 
 from neo4j_graphrag.exceptions import PdfLoaderError
-from neo4j_graphrag.experimental.pipeline.component import Component, DataModel
-
-
-class DocumentInfo(DataModel):
```
stellasia, line 31: Moved to `types.py`

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Next
 
+### Fixed
+- IDs for the Document and Chunk nodes in the lexical graph are now randomly generated and unique across multiple runs, fixing issues in the lexical graph where relationships where created between chunks that were created by different pipeline runs.
```
willtai, line 6:
```suggestion
- IDs for the Document and Chunk nodes in the lexical graph are now randomly generated and unique across multiple runs, fixing issues in the lexical graph where relationships were created between chunks that were created by different pipeline runs.
```

src/neo4j_graphrag/experimental/components/lexical_graph.py
```diff
@@ -103,19 +108,23 @@ def create_document_node(self, document_info: DocumentInfo) -> Neo4jNode:
         """
         document_metadata = document_info.metadata or {}
         return Neo4jNode(
-            id=document_info.path,
+            id=document_info.document_id,
             label=self.config.document_node_label,
             properties={
                 "path": document_info.path,
+                "createdAt": str(datetime.now()),
```
willtai, line 115: I think this should be timezone aware. Maybe:

```python
"createdAt": datetime.now(timezone.utc).isoformat(),
```

stellasia: Updated in https://github.com/neo4j/neo4j-graphrag-python/pull/222/commits/96451c5830cc308cc3bb2306e7bb7a39ac9513b7 (same syntax as RunResult that prevents the deprecation warning and works for all supported Python versions).

Nice suggestion of the `isoformat` method btw :)

src/neo4j_graphrag/llm/mistralai_llm.py
willtai: I think this could hide problems if `response.choices[0].message.content` is `None` or a non-string type.

Maybe use `getattr` for more defensive programming?

```python
content = getattr(response.choices[0].message, "content", "")
if not isinstance(content, str):
    content = ""
```

If this is overkill maybe something like:

```python
message_content = response.choices[0].message.content
content: str = message_content if isinstance(message_content, str) else ""
```

stellasia: Thanks, will take a look, actually I don't know why we're running into this issue now, mistralai package was just upgraded from `1.1.0` to `1.2.3`.

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
```diff
@@ -153,17 +152,19 @@ async def run(
...
+        chunk: TextChunk,
     ) -> Neo4jGraph:
-        """Make node IDs unique across chunks and pipeline runs by
-        prefixing them with a custom prefix (set in the run method)
-        and chunk index."""
-        prefix = f"{run_id}:{chunk_index}"
+        """Make node IDs unique across chunks, document and pipeline runs
+        by prefixing them with a unique prefix.
+        """
+        prefix = f"{chunk.chunk_id}"
```
alexthomas93, line 162: If you remove the `run_id` from the node ID will this not cause an issue if you run a pipeline multiple times over the same document? Or is the idea that it should fail in that situation?

stellasia: Now the `chunk_id` is unique (it's a UUID), even if we rerun the pipeline on the same document (previously, we were using the chunk index as chunk ID, this is why we needed a prefix to make the entity IDs unique, but this was the cause of the bug: the chunk IDs were not unique, and so rerunning the pipeline multiple times on the same doc was messing up the relationships between chunks)

alexthomas93: I see. So, all else being equal, would two runs over the same document produce two graphs with no links between them?

stellasia: Yes

src/neo4j_graphrag/llm/mistralai_llm.py
willtai: [Looks like it's because of this change.](https://github.com/mistralai/client-python/commit/0fd42415435e5c6314e252148802aed77a1d0c6a#diff-70d6ed7b626ffd0b1a78a90a45537cbc0ac7af6819822eb92ff6fd0f376074f2R34)

Do you still think it's ok to just cast to `str` here? Or should we take into `ContentChunk`?

stellasia: Resolving as we take your solution from #226

alexthomas93 (approved):
LGTM

PR #225, opened 2024-12-06
Chat with message history (draft)

leila-messallem:
# Description

Suggestion for adding chat functionality with message history.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,17 +18,3 @@ repos:
         language: system
         types: [ python ]
         stages: [ commit, push ]
-      - id: mypy
-        name: Mypy Type Check
-        entry: mypy .
-        language: system
-        types: [ python ]
-        stages: [ commit, push ]
-        pass_filenames: false
-        args: [
-          --strict,
-          --ignore-missing-imports,
-          --allow-untyped-calls,
-          --allow-subclassing-any,
-          --exclude='./docs/'
-        ]
--- a/src/neo4j_graphrag/generation/graphrag.py
+++ b/src/neo4j_graphrag/generation/graphrag.py
@@ -83,6 +83,7 @@ class GraphRAG:
     def search(
         self,
         query_text: str = "",
+        chat_history: Optional[list[str]] = None,
         examples: str = "",
         retriever_config: Optional[dict[str, Any]] = None,
         return_context: bool | None = None,
@@ -100,6 +101,7 @@ class GraphRAG:
 
         Args:
             query_text (str): The user question
+            chat_history: (Optional[list]): A list of previous messages in the conversation
             examples (str): Examples added to the LLM prompt.
             retriever_config (Optional[dict]): Parameters passed to the retriever
                 search method; e.g.: top_k
@@ -134,7 +136,10 @@ class GraphRAG:
         )
         logger.debug(f"RAG: retriever_result={retriever_result}")
         logger.debug(f"RAG: prompt={prompt}")
-        answer = self.llm.invoke(prompt)
+        if chat_history is not None:
+            answer = self.llm.chat(prompt, chat_history)
+        else:
+            answer = self.llm.invoke(prompt)
         result: dict[str, Any] = {"answer": answer.content}
         if return_context:
             result["retriever_result"] = retriever_result
--- a/src/neo4j_graphrag/llm/base.py
+++ b/src/neo4j_graphrag/llm/base.py
@@ -33,10 +33,12 @@ class LLMInterface(ABC):
         self,
         model_name: str,
         model_params: Optional[dict[str, Any]] = None,
+        system_instruction: Optional[str] = None,
         **kwargs: Any,
     ):
         self.model_name = model_name
         self.model_params = model_params or {}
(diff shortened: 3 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -18,17 +18,3 @@ repos:
         language: system
         types: [ python ]
         stages: [ commit, push ]
-      - id: mypy
```
leila-messallem, line 21: Sorry about this, but I couldn't bother with mypy just for opening a d Draft.

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -134,7 +136,10 @@ def search(
         )
         logger.debug(f"RAG: retriever_result={retriever_result}")
         logger.debug(f"RAG: prompt={prompt}")
-        answer = self.llm.invoke(prompt)
+        if chat_history is not None:
```
leila-messallem, line 139: I know this part isn't great. The idea I want to demonstrate is to have two different flows for chat with message history and single invokation.

src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -64,6 +65,32 @@ def get_messages(
...
         ]
 
+    def get_conversation_history(
+        self,
+        input: str,
+        chat_history: list[str],
+    ) -> Iterable[ChatCompletionMessageParam]:
+        messages = [{"role": "system", "content": self.system_instruction}]
+        for i, message in enumerate(chat_history):
+            if i % 2 == 0:
```
leila-messallem, line 75: This assumes that the chat history is a list of strings, alternating between `user` and `assistant` messages. It's obviously fragile and should probably be changed to some kind of `UserMessage` and `AssistantMessage` objects, or smtn like that.

.pre-commit-config.yaml
stellasia: It's a good point actually, maybe we should disable this for draft PRs

src/neo4j_graphrag/llm/openai_llm.py
stellasia: Maybe we can use dicts with "role" and "content" keys (it'd work for OpenAI, I don't know about the others)

src/neo4j_graphrag/llm/base.py
```diff
@@ -52,6 +54,21 @@ def invoke(self, input: str) -> LLMResponse:
             LLMGenerationError: If anything goes wrong.
         """
 
+    @abstractmethod
+    def chat(self, input: str, chat_history: list[str]) -> LLMResponse:
```
stellasia, line 58: I feel like it could be done in one method, with an optional `chat_history`. The difference would be dealt with in a `get_messages` method, that would do something like this:

```
def get_messages(self, input, chat_history) -> list:
    messages = []
    if self.system_instructions:
        messages.append({"role": "system", "content": self.system_instructions}})
    if chat_history:
        # handle chat history messages format
        messages.extend(chat_history)
    messages.append({"role": "user", "content": input})
    return messages
```

This is an advantage if most of the LLM providers share the same format for messages, in which case we can implement this method only once in the interface.

src/neo4j_graphrag/llm/vertexai_llm.py
```diff
@@ -55,15 +60,18 @@ def __init__(
...
     ):
         if GenerativeModel is None or ResponseValidationError is None:
             raise ImportError(
                 "Could not import Vertex AI Python client. "
                 "Please install it with `pip install google-cloud-aiplatform`."
             )
-        super().__init__(model_name, model_params)
-        self.model = GenerativeModel(model_name=model_name, **kwargs)
+        super().__init__(model_name, model_params, system_instruction)
+        self.model = GenerativeModel(
```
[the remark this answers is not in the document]
stellasia: This is quite bad news actually, that means that we can use an LLM instance only in one specific use case? For instance, with RAG with a text2Cypher retriever, we can't use the same LLM for the retrieval and the generation? (I mean, if we want to use the system prompt, which is not what we do at the moment)

src/neo4j_graphrag/llm/base.py
leila-messallem: Ok, that works.

src/neo4j_graphrag/llm/openai_llm.py
leila-messallem: I can look into it.

src/neo4j_graphrag/llm/vertexai_llm.py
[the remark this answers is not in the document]
leila-messallem: I guess we don't have to always use the `system_instruction`. Although, it's the recommended way.

leila-messallem:
https://github.com/neo4j/neo4j-graphrag-python/pull/234

PR #226, opened 2024-12-06
Use json-repair package to fix LLM generated json

willtai:
# Description

This PR improves on fixing LLM-generated JSON objects for the EntityRelationExtractor with the following:
- Usage of the `json-repair` package to fix LLM generated JSON objects. 
- Additional instructions to the LLM prompt.

The previous function to clean up LLM generated JSON objects is now removed.

This was tested using several LLMs using Ollama. During testing, the prompt and model temperature were changed to encourage the LLM to generate invalid JSON. These outputs were used to validate the new methods.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


examples/build_graph/simple_kg_builder_from_pdf_ollama.py
```diff
@@ -0,0 +1,105 @@
...
+import neo4j
+from neo4j_graphrag.experimental.pipeline.kg_builder import SimpleKGPipeline
+from neo4j_graphrag.experimental.pipeline.pipeline import PipelineResult
+from neo4j_graphrag.llm import LLMInterface, LLMResponse
+
+from llama_index.embeddings.ollama import OllamaEmbedding
+from neo4j_graphrag.embeddings.base import Embedder
+
+
+class OllamaEmbedder(Embedder):
```
stellasia, line 23: - Why not using the OpenAIEmbeddings class, which is compatible with ollama? I think having this here might be misleading for the users, as they do not have to implement their own embedder to use Ollama.
- In general, does this Ollama stuff belong to this PR? I see you've also added the dependency to ollama?

pyproject.toml
```diff
@@ -44,10 +44,13 @@ google-cloud-aiplatform = {version = "^1.66.0", optional = true }
 cohere = {version = "^5.9.0", optional = true}
 mistralai = {version = "^1.0.3", optional = true}
 qdrant-client = {version = "^1.11.3", optional = true}
-llama-index = {version = "^0.10.55", optional = true }
```
stellasia, line 47: I think we need this dep because an example shows how to use a text splitter from llama-index?

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
```diff
@@ -100,28 +102,19 @@ def balance_curly_braces(json_string: str) -> str:
...
-    invalid_json_string = re.sub(r"{{+", "{", invalid_json_string)
-    invalid_json_string = re.sub(r"}}+", "}", invalid_json_string)
+    if isinstance(repaired_json, str):
+        repaired_json = repaired_json.strip()
+    else:
+        repaired_json = ""
 
-    # Balance curly braces
-    return balance_curly_braces(invalid_json_string)
+    if repaired_json.strip() == '""':
```
stellasia, line 113: It's already `strip`ed, right?

tests/unit/experimental/components/test_entity_relation_extractor.py
```diff
@@ -154,8 +155,8 @@ async def test_extractor_llm_badly_formatted_json() -> None:
         llm=llm,
     )
     chunks = TextChunks(chunks=[TextChunk(text="some text", index=0)])
-    with pytest.raises(LLMGenerationError):
-        await extractor.run(chunks=chunks)
+
+    await extractor.run(chunks=chunks)
```
stellasia, line 159: Do we need to add some `assert` here? Maybe to check the json was actually fixed as we expect? (or maybe remove this test if it's the same as `test_extractor_llm_badly_formatted_json_gets_fixed`?)

tests/unit/experimental/components/test_entity_relation_extractor.py
```diff
@@ -205,6 +210,14 @@ async def test_extractor_custom_prompt() -> None:
     llm.ainvoke.assert_called_once_with("this is my prompt")
 
 
+def test_fix_invalid_json_empty_result() -> None:
+    json_string = "invalid json"
+
+    with patch("json_repair.repair_json", return_value=""):
```
stellasia, line 216: Why do we need to patch the return value if the input json is invalid?

src/neo4j_graphrag/llm/mistralai_llm.py
```diff
@@ -84,10 +84,11 @@ def invoke(self, input: str) -> LLMResponse:
                 messages=self.get_messages(input),
                 **self.model_params,
             )
-            if response is None or response.choices is None or not response.choices:
-                content = ""
-            else:
-                content = response.choices[0].message.content or ""
+            content: str = ""
```
stellasia, line 87: Regarding this, I think we should wait a bit, because they've changed this part of the code again last week. So maybe just pin the version to before the pb first arose, and come back to it when their code is stabilized. WDYT?

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/226/commits/85f2b450bd026074028f383670c30bd0cc5404fa

src/neo4j_graphrag/llm/mistralai_llm.py
willtai: Agreed

tests/unit/experimental/components/test_entity_relation_extractor.py
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/226/commits/ab42906ebdf1d9d63682081132d62b1a9e1fb419

I've modified the test scenario to have an unfixable json

pyproject.toml
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/226/commits/c1ef1d637c63c5fa6c14c0a25c8bd9562705534b

examples/build_graph/simple_kg_builder_from_pdf_ollama.py
willtai: Ollama is only used here for manual testing. Removing now

tests/unit/experimental/components/test_entity_relation_extractor.py
willtai: When json_repair is unable to fix the JSON string, it returns `""`

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
```diff
@@ -100,28 +102,19 @@ def balance_curly_braces(json_string: str) -> str:
...
-
-    # Correct the specific issue: remove trailing commas within arrays or objects before closing braces or brackets
-    invalid_json_string = re.sub(r",\s*(?=[}\]])", "", invalid_json_string)
+def fix_invalid_json(raw_json: str) -> str:
+    repaired_json = json_repair.repair_json(raw_json)
 
-    # Normalize excessive curly braces
-    invalid_json_string = re.sub(r"{{+", "{", invalid_json_string)
-    invalid_json_string = re.sub(r"}}+", "}", invalid_json_string)
+    if isinstance(repaired_json, str):
```
stellasia, line 108: In which cases it's not a string?

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
```diff
@@ -100,28 +102,19 @@ def balance_curly_braces(json_string: str) -> str:
...
+        repaired_json = repaired_json.strip()
+    else:
+        repaired_json = ""
 
-    # Balance curly braces
-    return balance_curly_braces(invalid_json_string)
+    if repaired_json == '""':
+        raise JSONRepairError("JSON repair resulted in an empty or invalid JSON.")
```
stellasia, line 114: Regarding the naming, shall we use something less tied to the library we use, for instance `InvalidJsonError`?

tests/unit/experimental/components/test_entity_relation_extractor.py
```diff
@@ -190,7 +192,11 @@ async def test_extractor_llm_badly_formatted_json_do_not_raise() -> None:
     )
     chunks = TextChunks(chunks=[TextChunk(text="some text", index=0)])
     res = await extractor.run(chunks=chunks)
-    assert res.nodes == []
+    print("res.nodes", res.nodes)
```
stellasia, line 195: Print can be removed :)

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
```diff
@@ -100,28 +102,19 @@ def balance_curly_braces(json_string: str) -> str:
...
-
-    # Fix missing quotes around string values, correctly ignoring null, true, false, and numeric values
-    invalid_json_string = re.sub(
-        r"(?<=:\s)(?!(null|true|false|\d+\.?\d*))([a-zA-Z_][a-zA-Z0-9_]*)\s*(?=[,}])",
-        r'"\2"',
-        invalid_json_string,
-    )
-
-    # Correct the specific issue: remove trailing commas within arrays or objects before closing braces or brackets
-    invalid_json_string = re.sub(r",\s*(?=[}\]])", "", invalid_json_string)
```
alexthomas93, line 117: Given we're replacing the old JSON fixing code here does json_repair completely cover the situations these regexes were meant to fix? Would there be any reason to keep these around?

willtai: Yes there were previous unit tests that still pass 👍 

When I combined both json_repair and the old fixing code, it causes more problems. I think this was because the original code to fix the json string was not perfect and lacked many cases. So IMO it's better to just rely on json_repair

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/226/commits/63e26cc0f1d1b65a9cc62386597d2949d4a83e21

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
willtai: Looks like it can be: [JSONReturnType](https://github.com/mangiucugna/json_repair/blob/main/src/json_repair/json_parser.py#L6)

Do you think this check is appropriate?

stellasia: I see. However, the `repair_json` function returns a `json.dumps`, so it will always be a string in our case (https://github.com/mangiucugna/json_repair/blob/main/src/json_repair/json_repair.py#L72)

If mypy is complaining, maybe we can use `cast` here as well?

willtai: Hmm in this case wouldn't it better to have a runtime check using `isinstance`? In case this implementation changes or `repair_json` is incorrectly used

stellasia: This is not a part that we let to developers, it's our code. So for now we know how it behaves and if we change it later on, it will be the time to add checks. Same for a potential change in a later version of the library. Basically my view is: let's not add complexity to cover future potential issues, that we are not even sure will ever happen.

willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/226/commits/639cae231ea040dd92053a5637c403208e0a0e04

docs/source/api.rst
```diff
@@ -509,3 +509,10 @@ PipelineStatusUpdateError
 
 .. autoclass:: neo4j_graphrag.experimental.pipeline.exceptions.PipelineStatusUpdateError
    :show-inheritance:
+
+
+InvalidJSONError
```
stellasia, line 514: I think the exceptions are also listed above, though I don't know why we have this list :)

pyproject.toml
```diff
@@ -48,6 +48,8 @@ llama-index = {version = "^0.10.55", optional = true }
 openai = {version = "^1.51.1", optional = true }
 anthropic = { version = "^0.36.0", optional = true}
 sentence-transformers = {version = "^3.0.0", optional = true }
+setuptools = "^75.6.0"
```
stellasia, line 51: Do you know we need this addition?

stellasia (approved):
Minor comments, but looks good! Thanks for adding this @willtai , I think it will really improve user experience!

pyproject.toml
willtai: Not any more, I've removed it

docs/source/api.rst
willtai: Do you mean the hyperlinks listed above?

willtai:

docs/source/api.rst
stellasia: Yes, thank you!

PR #230, opened 2024-12-10
Add ability to load PDF from HTTP + specify file system from string (useful for config files)

stellasia:
# Description

> **Note**
>
>
>

## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/experimental/components/pdf_loader.py
```diff
@@ -85,16 +82,21 @@ def load_file(
...
-        filepath: Path,
+        filepath: Union[str, Path],
         metadata: Optional[Dict[str, str]] = None,
-        fs: Optional[AbstractFileSystem] = None,
+        fs: Optional[Union[AbstractFileSystem, str]] = None,
     ) -> PdfDocument:
-        fs = fs or LocalFileSystem()
+        if not isinstance(filepath, str):
+            filepath = str(filepath)
+        if isinstance(fs, str):
```
alexthomas93, line 91: Why pass a str for the filesystem and initialise it here and not just pass the `fsspec.filesystem` to the method directly? We could remove the `Union` part of the type hint if we did. We could also set the default to `LocalFileSystem`

src/neo4j_graphrag/experimental/pipeline/config/object_config.py
```diff
@@ -254,6 +254,10 @@ class ComponentConfig(ObjectConfig[Component]):
     DEFAULT_MODULE = "neo4j_graphrag.experimental.components"
     INTERFACE = Component
 
+    def get_run_params(self, resolved_data: dict[str, Any]) -> dict[str, Any]:
+        self._global_data = resolved_data
+        return self.resolve_params(self.run_params_)
+
 
 class ComponentType(RootModel):  # type: ignore[type-arg]
```
alexthomas93, line 262: Not part of this PR but do we know why this type: ignore comment is here?

alexthomas93 (approved):
LGTM, just a few minor comments

src/neo4j_graphrag/experimental/components/pdf_loader.py
stellasia: The reason for the string is to make it work with config files. We could have a parser in the config parser to initialize a file system though, but I thought it was easier to just support str since this is allowed by `fsspec`.

alexthomas93: Makes sense. I don't have a strong view on whether or not it should be here or in the config parser so let's just keep it here for now

src/neo4j_graphrag/experimental/pipeline/config/object_config.py
stellasia: Yes, this is because `pydantic.RootModel` can be used either like this:
```
class ComponentType(RootModel):  # type: ignore[type-arg]
    root: Union[Component, ComponentConfig]
```
 or with a simple declaration like this:
```
ComponentType = RootModel[Union[Component, ComponentConfig]]
```
We use the former to add other methods to the class, but mypy is complaining that the type annotation `[]` is missing for RootModel.

alexthomas93: Hmm, that makes sense. Is there no way to point mypy to `root` in your example to resolve this issue? Feels a bit wrong to have this here. Maybe something we could look at in another PR in the future?

stellasia: I couldn't find a workaround but I have to admit I've not spent a lot of time on it. Will create a card so that we can investigate further.

PR #234, opened 2024-12-13
Support for conversations with message history

leila-messallem:
# Description

Adding chat functionality with message history.

* Added an optional `chat_history` parameter to the `invoke` method of the `LLMInterface`.
* The `chat_history` is a dict with `role` and `content` keys, where `role` can be either "user" or "assistant".
* Added a `system_instruction` parameter to the LLM class instantiation (meaning that you have to create a separate LLM instance for each use case. The motivation being that on `vertexai`, `system_instruction` is set on the `GenerativeModel` object, so separated from the question prompt).
* Added a summary of the chat history to the query embedding (so that the relevant context gets included in the embedding).

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: High

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -124,18 +127,25 @@ def search(
...
+        answer = self.llm.invoke(prompt, chat_history)
         result: dict[str, Any] = {"answer": answer.content}
         if return_context:
             result["retriever_result"] = retriever_result
         return RagResultModel(**result)
+    
+    def build_query(self, query_text: str, chat_history: list[dict[str, str]]) -> str:
+        if chat_history:
+            summarization_prompt = ChatSummaryTemplate().format(chat_history=chat_history)
+            summary = self.llm.invoke(summarization_prompt).content
```
leila-messallem, line 149: Just realised that we have to reset the `system_instruction` on the llm for this to work. And this will be a problem for `VertexAI` because `system_instruction` is set directly on the GenerativeModel object.

src/neo4j_graphrag/generation/prompts.py
```diff
@@ -196,3 +196,34 @@ def format(
...
+            for message in chat_history
+        ]
+        history = "\n".join(message_list)
+        return super().format(chat_history=history)
+
+
+class ConversationTemplate(PromptTemplate):
+    DEFAULT_TEMPLATE = """
+Chat Summary: 
+{summary}
```
stellasia, line 221: Why is the chat summary required here? Isn't the chat history handling the past messages?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -49,62 +58,74 @@ def __init__(
         self,
         model_name: str,
         model_params: Optional[dict[str, Any]] = None,
+        system_instruction: Optional[str] = None,
         **kwargs: Any,
     ):
-        try:
-            import anthropic
-        except ImportError:
+        if anthropic is None:
```
stellasia, line 64: What is the reason for this change? We had something similar in the past and moved to this new approach of importing the module locally.

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -49,62 +58,74 @@ def __init__(
...
-    def invoke(self, input: str) -> LLMResponse:
+    def get_messages(
+        self, input: str, chat_history: Optional[list[Any]] = None
+    ) -> Iterable[MessageParam]:
+        messages = []
+        if chat_history:
+            try:
+                MessageList(messages=chat_history)
+            except ValidationError as e:
+                raise LLMGenerationError(e.errors()) from e
```
stellasia, line 82: Detail: I don't know if it should be an LLMGenerationError or some sort of ValidationError?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -49,62 +58,74 @@ def __init__(
...
+            messages.extend(chat_history)
+        messages.append(UserMessage(content=input).model_dump())
+        return messages
+
+    def invoke(
+        self, input: str, chat_history: Optional[list[Any]] = None
```
stellasia, line 88: Isn't `Any` here just a `dict[str, str]`?

src/neo4j_graphrag/generation/graphrag.py
stellasia: Would that work if we instantiate a client for each call to `invoke` (for VertexAI I mean)?

leila-messallem: Something like that. To be continued next week 🙂

src/neo4j_graphrag/generation/prompts.py
leila-messallem: Yes, but this template is for the query embedding. The chat history can potentially be large, so the idea is to first generate a summary and embed it together with the current query, instead of embedding the entire chat history.

src/neo4j_graphrag/llm/anthropic_llm.py
leila-messallem: It's actually a silly reason, and let me know how I can handle it in a better way 😄 - I wanted to use the `MessageParam` type from the anthropic package as the return type for the `get_messages()` method (on ln 76), and this was the only way I could make it work 🤷‍♀️

src/neo4j_graphrag/llm/anthropic_llm.py
leila-messallem: Yes, I had that initially, but mypy was complaining on it for some reason.

src/neo4j_graphrag/llm/anthropic_llm.py
leila-messallem: I was trying to follow the example from the already implemented method in the [BaseOpenAILLM class](https://github.com/leila-messallem/neo4j-graphrag-python/blob/bc6dd9c7b3f8fcfffb9ed360648ea80c6cbb17dc/src/neo4j_graphrag/llm/openai_llm.py#L62).

Would it be better to just change the return type to `list[dict[str, str]]`?

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -83,6 +87,7 @@ def __init__(
     def search(
         self,
         query_text: str = "",
+        chat_history: Optional[list[dict[str, str]]] = None,
```
oskarhane, line 90: `chat_history` is an optional list of type `{"role": "system"|"assistant"|"user", "content": str}`, which usually are referred to as `messages`, so I wonder if we can 

1. be more specific in the type declaration 
2. and also rename the argument (everywhere) to `message_history`

Might seem as a unnecessary nitpick, but the messages might not come from a chat.

src/neo4j_graphrag/llm/anthropic_llm.py
stellasia: To use a type from a dev dependency, you can do like this:

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from anthropic import MessageParam 
```

See how it's done in https://github.com/stellasia/neo4j-graphrag-python/blob/main/src/neo4j_graphrag/llm/openai_llm.py#L26

stellasia: To use a type from a dev dependency, you can do like this:

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from anthropic import MessageParam
```

Then MessageParam is available for mypy, and Python just ignores the annotations at runtime so it's not an issue.

src/neo4j_graphrag/llm/anthropic_llm.py
stellasia: IMHO, we can keep `Any` at the moment, maybe just add a comment so that we remember why and we know where to start from if we really want to remove it in the future.

src/neo4j_graphrag/generation/graphrag.py
leila-messallem: WDYT about [this](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/5bb56f6130dff46ced90e693e7bdeaa7b5339173)? It's getting quite convoluted, I don't know if there's a simpler way 🤔

stellasia: It makes sense, I like the idea of having the system instructions with the prompt, they are quite tied together. 

Do we still need the system instruction in the init of the LLMInterface if we have it in the invoke method?

leila-messallem: > Do we still need the system instruction in the init of the LLMInterface if we have it in the invoke method?

Technically, no. But then it would be required to send the system instruction on each invokation, which would be kind of contradictory to it's purpose (to set it once, as a frame for the conversation), at least the way I understand it.

src/neo4j_graphrag/llm/anthropic_llm.py
leila-messallem: [6aea7fa](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/6aea7fa31b056af58c75feb910f236397681c5d3)

src/neo4j_graphrag/generation/graphrag.py
stellasia: Yeah, but having said that, the system instruction is just another message, that gets inserted at the beginning of the message list, so it could be treated in the same way. Also, I think the `llm` object could be used for different purposes. For instance, consider GraphRAG with a Text2Cypher retriever: the same LLM object can serve as text2cypher generator and answer generator. In that case, we won't use the system instructions in the init.

src/neo4j_graphrag/generation/prompts.py
stellasia: That's a great idea. I thought the ChatSummaryTemplate was used to summarized the chat history and embed it?

src/neo4j_graphrag/generation/graphrag.py
leila-messallem: Rename: [37225fd](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/37225fdd124d21701cbfa7a0f7256c3d91e8c3b4), type declaration: [66b2358](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/66b235872500fadbe727d6983695571146419e83).

src/neo4j_graphrag/generation/graphrag.py
leila-messallem: But how are you thinking it should work in the case of chat conversation? We want to be able to set the system instruction once before the conversation starts, and avoid having to send it along with each user message, right?

src/neo4j_graphrag/generation/prompts.py
leila-messallem: The `ChatSummaryTemplate` is for summarizing the chat history, this summary is then used together with the user query for embedding, using the `ConversationTemplate`. See [graphrag.py](https://github.com/leila-messallem/neo4j-graphrag-python/blob/66b235872500fadbe727d6983695571146419e83/src/neo4j_graphrag/generation/graphrag.py#L150).

src/neo4j_graphrag/generation/graphrag.py
leila-messallem: Tried to declare the type as [BaseMessage](https://github.com/leila-messallem/neo4j-graphrag-python/blob/a749a9e2eb596763d79c6375170be9bc8b0478d0/src/neo4j_graphrag/llm/types.py#L9), but this doesn't work out if we want to receive a dictionary. So how exactly do you mean that we can achieve a more specific type declaration?

src/neo4j_graphrag/llm/ollama_llm.py
```diff
@@ -12,19 +12,25 @@
...
+from pydantic import ValidationError
 
 from neo4j_graphrag.exceptions import LLMGenerationError
 
 from .base import LLMInterface
-from .types import LLMResponse
+from .types import LLMResponse, SystemMessage, UserMessage, MessageList
+
+if TYPE_CHECKING:
+    from ollama import Message
```
leila-messallem, line 25: Why isn't this working? It's giving `NameError: name 'Message' is not defined`.

src/neo4j_graphrag/generation/graphrag.py
oskarhane: This should do it, if I understand your question.

```python
from typing import TypedDict, Literal

class Message(TypedDict):
    role: Literal["system", "user", "assistant"]
    content: str
```

That would make this pass the type checker:

```python
message: Message = {"role": "user", "content": "Hello, world!"}
```

but not 

```python
invalid_message: Message = {"role": "bot", "content": "Invalid role"}
```

leila-messallem: How is this different from abef33c](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/abef33c195b7c23ec0a18482baa4190775fe6b10)?

src/neo4j_graphrag/llm/ollama_llm.py
leila-messallem: [17db6b1](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/17db6b18de6708ffac48fb557d5991b387d2fac2)

src/neo4j_graphrag/generation/graphrag.py
oskarhane: The pydantic model subclass needs to be instantiated like `message = BaseMessage(role="user", content="Hello!")` where the typings type can be a plain dict `message: Message = {"role": "user", "content": "Hello, world!"}`.

The difference is that the latter is easier to integrate with when relaying data that already are plain dicts, but there's no runtime validation.

leila-messallem: So you want to replace the pydantic model with a TypedDict and scrap [the validation](https://github.com/leila-messallem/neo4j-graphrag-python/blob/17db6b18de6708ffac48fb557d5991b387d2fac2/src/neo4j_graphrag/llm/openai_llm.py#L79)?

oskarhane: On the specific line I commented on, the typing is not helpful for developers `Optional[list[dict[str, str]]]`.

src/neo4j_graphrag/generation/prompts.py
stellasia: Ok I see, this is only used in the Retriever. I got confused because `Template` for me refer to prompt template. They were introduced with the idea that users can define the template and the placeholders. But here we (developers) do both, so we could skip them and use simple Python string formatting. Up to you :)

leila-messallem: If it's confusing, I'll change it to a string.

src/neo4j_graphrag/generation/graphrag.py
leila-messallem: Ok, I get it now, [d5a287b](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/d5a287b94aaeb33bbfe9634e8b2715fd26fdab38). 

Mypy doesn't accept it, but I guess it's something you work around with fake "casting"?

src/neo4j_graphrag/generation/prompts.py
leila-messallem: [bd34e1a](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/bd34e1ae4d882886d4da1433dbd967f3548b6357). Is it fine to keep those functions in the `prompts` module?

leila-messallem:

leila-messallem:
@CodiumAI-Agent  /update_changelog --pr_update_changelog.push_changelog_changes=true

src/neo4j_graphrag/generation/prompts.py
stellasia: If used in only one place, can be moved closer to where they are used. The convention for function names is to use `snake_case` :D

leila-messallem: Right 🤦‍♀️ 😄  [f5a9833](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/f5a983373b3fcedecff9a575369286c038393873)

examples/customize/llms/custom_llm.py
```diff
@@ -1,21 +1,32 @@
...
 from neo4j_graphrag.llm import LLMInterface, LLMResponse
+from neo4j_graphrag.llm.types import LLMMessage
 
 
 class CustomLLM(LLMInterface):
     def __init__(self, model_name: str, **kwargs: Any):
         super().__init__(model_name, **kwargs)
 
-    def invoke(self, input: str) -> LLMResponse:
+    def invoke(
```
stellasia, line 13: The `system_instruction` param is missing in the init I think?

src/neo4j_graphrag/llm/base.py
```diff
@@ -17,7 +17,7 @@
 from abc import ABC, abstractmethod
 from typing import Any, Optional
 
-from .types import LLMResponse
+from .types import LLMMessage, LLMResponse
```
stellasia, line 20: Do we need to add the LLMessage type to the doc? We have a `types` page for the types that are user-facing.

leila-messallem: I have built the docs 👍

examples/customize/llms/custom_llm.py
leila-messallem: [a15a514](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/a15a514b8c28e887a8d80cac69556872a6c73f9e)

src/neo4j_graphrag/llm/base.py
leila-messallem: ... and added the type, [7557b07](https://github.com/neo4j/neo4j-graphrag-python/pull/234/commits/7557b07581f8ba1371c4727b3a36548413f53643).

stellasia (approved):
Nice work! 🥳🥳

Thanks for dealing with all the issues, especially the vendor specificity 🙏

PR #235, opened 2024-12-13
Improve logging

stellasia:
# Description

Truncate long lists and strings in logs. The max length of list of string is configurable using env vars.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/pipeline/config/runner.py
+++ b/src/neo4j_graphrag/experimental/pipeline/config/runner.py
@@ -48,7 +48,7 @@ from neo4j_graphrag.experimental.pipeline.config.template_pipeline.simple_kg_bui
 from neo4j_graphrag.experimental.pipeline.config.types import PipelineType
 from neo4j_graphrag.experimental.pipeline.pipeline import PipelineResult
 from neo4j_graphrag.experimental.pipeline.types import PipelineDefinition
-from neo4j_graphrag.utils import prettyfier
+from neo4j_graphrag.utils.logging import prettyfier
 
 logger = logging.getLogger(__name__)
 
--- a/src/neo4j_graphrag/experimental/pipeline/config/runner.py
+++ b/src/neo4j_graphrag/experimental/pipeline/config/runner.py
@@ -70,6 +70,7 @@ class PipelineConfigWrapper(BaseModel):
     ] = Field(discriminator=Discriminator(_get_discriminator_value))
 
     def parse(self, resolved_data: dict[str, Any] | None = None) -> PipelineDefinition:
+        logger.debug("PIPELINE_CONFIG: start parsing config...")
         return self.config.parse(resolved_data)
 
     def get_run_params(self, user_input: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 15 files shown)


src/neo4j_graphrag/experimental/pipeline/config/runner.py
```diff
@@ -101,10 +103,14 @@ def from_config(
         cls, config: AbstractPipelineConfig | dict[str, Any], do_cleaning: bool = False
     ) -> Self:
         wrapper = PipelineConfigWrapper.model_validate({"config": config})
+        logger.debug(
+            f"PIPELINE_RUNNER: instantiate Pipeline from config type: {wrapper.config.template_}"
```
alexthomas93, line 107: Minor but this should probably be `instantiating` and not `instantiate`

src/neo4j_graphrag/utils/logging.py
```diff
@@ -0,0 +1,80 @@
...
+        if isinstance(value, BaseModel):
+            return self(value.model_dump())
+        if isinstance(value, list):
+            return self._prettyfy_list(value)
+        if isinstance(value, str):
+            return self._prettyfy_str(value)
+        return value
+
+
+prettyfier = Prettyfier()
```
alexthomas93, line 80: Why instantiate an instance of the Prettyfier class here and then import the instance? Why not just import the class and then create an instance of it where it's used?

src/neo4j_graphrag/utils/logging.py
```diff
@@ -0,0 +1,80 @@
...
+import os
+from typing import Any
+
+from pydantic import BaseModel
+
+DEFAULT_MAX_LIST_LENGTH: int = 5
+DEFAULT_MAX_STRING_LENGTH: int = 200
+
+
+class Prettyfier:
```
alexthomas93, line 26: Do we have any unit tests for this class? I think testing the nested functionality would be especially useful

src/neo4j_graphrag/utils/logging.py
stellasia: The main reason is to keep things simple when logging, we already need an extra step (calling the function), if we also have to instantiate the class every time we use a logger, I think it makes the code more complicated.

We could also have a function like this:

```
def prettify(value):
   prettifer = Prettifyer()
   return prettifier(value)
```

and import this function when logging, but since the `Prettifyer` class is always the same, I though it would be easier like this. But I'm happy to change if you prefer the function approach or another one.

alexthomas93: I'm happy to leave it for now as I don't have a good intuition for what would be the best approach in this situation. It might be something worth changing in a future refactor though.

src/neo4j_graphrag/experimental/pipeline/config/runner.py
stellasia: https://github.com/neo4j/neo4j-graphrag-python/pull/235/commits/ea0d0daeb1f88a850205d8db2c1b050ac81615da

alexthomas93 (approved):
LGTM, nice work on tests!

PR #239, opened 2025-01-03
Allow thresholding on vector and fulltext indexes for Hybrid retrievers

willtai:
# Description

Allow thresholding on vector and fulltext indexes for Hybrid retrievers. Two thresholds can be provided by the user during search to determine the importance of the search results from either vector or fulltext index. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_graphrag/neo4j_queries.py
```diff
@@ -122,25 +122,29 @@ def _get_hybrid_query(neo4j_version_is_5_23_or_above: bool) -> str:
...
-            f"RETURN n.node AS node, (n.score / vector_index_max_score) AS score "
+            f"RETURN n.node AS node, CASE WHEN (n.score / vector_index_max_score) >= $threshold_vector_index "
+            f"THEN (n.score / vector_index_max_score) ELSE 0 END AS score "
             f"UNION "
             f"{FULL_TEXT_SEARCH_QUERY} "
             f"WITH collect({{node:node, score:score}}) AS nodes, max(score) AS ft_index_max_score "
             f"UNWIND nodes AS n "
-            f"RETURN n.node AS node, (n.score / ft_index_max_score) AS score }} "
+            f"RETURN n.node AS node, CASE WHEN (n.score / ft_index_max_score) >= $threshold_fulltext_index "
+            f"THEN (n.score / ft_index_max_score) ELSE 0 END AS score }} "
```
alexthomas93, line 147: Should the `}}` be here?

src/neo4j_graphrag/retrievers/hybrid.py
```diff
@@ -159,6 +161,8 @@ def get_search_results(
             query_text (str): The text to get the closest neighbors of.
             query_vector (Optional[list[float]], optional): The vector embeddings to get the closest neighbors of. Defaults to None.
             top_k (int, optional): The number of neighbors to return. Defaults to 5.
+            threshold_vector_index (float, optional): The minimum normalized score from the vector index to include in the top k search.
+            threshold_fulltext_index (float, optional): The minimum normalized score from the fulltext index to include in the top k search.
```
alexthomas93, line 165: Minor point but users might not know or understand the normalisation process. It could be worth adding something to the docs with an example which explains a bit about what's going on with these parameters as I think on their own these descriptions might not be enough

tests/unit/test_neo4j_queries.py
```diff
@@ -37,12 +37,14 @@ def test_hybrid_search_basic() -> None:
         "YIELD node, score "
```
alexthomas93, line 37: Could it be worth adding a test to check the threshold process is working as expected? i.e. the correct scores are set to zero, etc.

alexthomas93 (approved):
LGTM, just a few minor points

stellasia:
I'd be interested in seeing an example of this. 

Here are the points bothering me:
- Aren't these parameters too query-dependent, since the range of normalized scores can vary a lot for each query?
- What if most of the scores end up being 0? Then we do not have any ordering? 
- Is this compatible with the effective search ratio implemented in langchain? Typically, this ratio would bring more items in this part of the query and so change the normalization.

tests/unit/test_neo4j_queries.py
willtai: If the scores are 0 they are now removed in https://github.com/neo4j/neo4j-graphrag-python/pull/239/commits/ef33344d22cbb1e5379b9198f991f5743d272520

src/neo4j_graphrag/retrievers/hybrid.py
willtai: Added more detail in the retriever docstring here: https://github.com/neo4j/neo4j-graphrag-python/pull/239/commits/9cd16af50c3cec8223c8b5f3cd00e6477edc7bfd

willtai:

src/neo4j_graphrag/neo4j_queries.py
```diff
@@ -116,33 +116,50 @@
...
-            f"UNWIND nodes AS n "
-            f"RETURN n.node AS node, (n.score / vector_index_max_score) AS score "
-            f"UNION "
-            f"{FULL_TEXT_SEARCH_QUERY} "
-            f"WITH collect({{node:node, score:score}}) AS nodes, max(score) AS ft_index_max_score "
-            f"UNWIND nodes AS n "
-            f"RETURN n.node AS node, (n.score / ft_index_max_score) AS score }} "
-            f"WITH node, max(score) AS score ORDER BY score DESC LIMIT $top_k"
-        )
+        return f"""CALL () {{
```
alexthomas93, line 128: Why change this to a multi-line string here?

src/neo4j_graphrag/neo4j_queries.py
```diff
@@ -116,33 +116,50 @@
...
+  UNWIND nodes AS n
+  WITH n.node AS node, CASE WHEN (n.score / vector_index_max_score) >= $threshold_vector_index
+  THEN (n.score / vector_index_max_score) ELSE 0 END AS score{vector_where_clause}
+  RETURN node, score
+  UNION
+  {FULL_TEXT_SEARCH_QUERY}
+  WITH collect({{node:node, score:score}}) AS nodes, max(score) AS ft_index_max_score
+  UNWIND nodes AS n
+  WITH n.node AS node, CASE WHEN (n.score / ft_index_max_score) >= $threshold_fulltext_index
+  THEN (n.score / ft_index_max_score) ELSE 0 END AS score{fulltext_where_clause}
```
alexthomas93, line 140: Should there not be a space between `score` and `{fulltext_where_clause}` here?

src/neo4j_graphrag/neo4j_queries.py
```diff
@@ -116,33 +116,50 @@
...
```
alexthomas93, line 158: Same here

src/neo4j_graphrag/neo4j_queries.py
willtai: This helps me modify the unit tests more easily without worrying about correct indentation and spaces

src/neo4j_graphrag/neo4j_queries.py
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/239/commits/454b2319e037040a4db336af5313b501e95dfdcb

src/neo4j_graphrag/neo4j_queries.py
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/239/commits/454b2319e037040a4db336af5313b501e95dfdcb

src/neo4j_graphrag/neo4j_queries.py
alexthomas93: With multi-line strings, you do have to worry about correct indentation though. With the
```python
(
"Hello "
"world"
)
``` 
approach you only have to worry about there being a space at the end of every line.

willtai: Hmm I still generally prefer multi-line strings as I find them more readable and tend to avoid mistakes after changes. Do you think we should revert this back?

alexthomas93: I’m inclined to prefer the older approach, but I’m also open to this option. @stellasia what are your thoughts?

NathalieCharbel:
I apologise for chiming in a bit late, but here's my opinion. 
While thresholding in a hybrid retriever can be useful, it may also introduce nuances that are misaligned with the main objective of identifying the most relevant results from both full-text and vector-based searches (with thresholding, you often filter out any results that do not exceed a certain threshold in one search method before considering their relevance in the other method, so relevant results can be lost prematurely).
Also, from a user's perspective, I find that adjusting how much weight is given to each type of score is more intuitive than tuning two threshold parameters based on score, which might be an obscure value for the user.
One way to do this is by combining the vector and full-text scores into a single one. So in general, you fetch top candidates from each search, compute one combined score, and then pick the best overall. This way the user only needs to tune a single parameter, something like:
`combined_score = α * score_vector + (1−α) * score_fulltext` where `α ∈ [0,1]`
wdyt?

PR #242, opened 2025-01-09
Improve text splitter to avoid cutting words in chunks

NathalieCharbel:
# Description

Add optional approximate splitting Flag to FixedSizeSplitte to ensure chunks start and end cleanly without words cut in the middle, honouring a desired chunk size and overlap. Also, “no whitespace found” fallbacks are considered during both chunk start and end adjustments, ensuring that if a large word has no whitespace, we revert to the original boundary to avoid unexpected behaviour.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_graphrag/experimental/components/text_splitters/fixed_size_splitter.py
```diff
@@ -56,10 +111,34 @@ async def run(self, text: str) -> TextChunks:
...
+            else:
+                # apply fixed size splitting with possibly words cut in half at chunk boundaries
+                start = approximate_start
+                end = min(start + self.chunk_size, text_length)
+
             chunk_text = text[start:end]
             chunks.append(TextChunk(text=chunk_text, index=index))
             index += 1
+
+            approximate_start = start + step
```
stellasia, line 142: Why not using the approximate chunk size here, so that we only have to adjust the chunk end once and restart from the same position for the next chunk?

In very simple code:

```python

start = 0
while start < text_length:
    chunk = text[start:start+chunk_length]
    if self.approximate:
        chunk = adjust_end(chunk)
    start += len(chunk) - chunk_overlap
```
?

src/neo4j_graphrag/experimental/components/text_splitters/fixed_size_splitter.py
```diff
@@ -33,16 +87,17 @@ class FixedSizeSplitter(TextSplitter):
...
 
         pipeline = Pipeline()
-        text_splitter = FixedSizeSplitter(chunk_size=4000, chunk_overlap=200)
+        text_splitter = FixedSizeSplitter(chunk_size=4000, chunk_overlap=200, approximate=True)
         pipeline.add_component(text_splitter, "text_splitter")
     """
 
     @validate_call
-    def __init__(self, chunk_size: int = 4000, chunk_overlap: int = 200) -> None:
+    def __init__(self, chunk_size: int = 4000, chunk_overlap: int = 200, approximate: bool = True) -> None:
```
stellasia, line 95: It's ok to default to `True`, even if it's a breaking change. Just remember to add a warning in the CHANGELOG when you get there.

src/neo4j_graphrag/experimental/components/text_splitters/fixed_size_splitter.py
```diff
@@ -18,12 +18,66 @@
...
+        text (str): The text being split.
+        proposed_start (int): The initial starting index of the chunk.
+
+    Returns:
+        int: The adjusted starting index, ensuring the chunk does not begin in the
+             middle of a word if possible.
+    """
+    start = proposed_start
+    if start > 0 and not text[start].isspace() and not text[start - 1].isspace():
+        while start > 0 and not text[start - 1].isspace():
```
stellasia, line 36: Have you considered using [`find`](https://docs.python.org/3/library/stdtypes.html#str.find) (or [`index`](https://docs.python.org/3/library/stdtypes.html#str.index)) methods?
(there are also `rfind` and `rindex` methods to get the last occurence)

src/neo4j_graphrag/experimental/components/text_splitters/fixed_size_splitter.py
```diff
@@ -18,12 +18,66 @@
...
+            start -= 1
+
+        # fallback if no whitespace is found
+        if start == 0 and not text[0].isspace():
+            start = proposed_start
+    return start
+
+
+def _adjust_chunk_end(text: str, start: int, approximate_end: int) -> int:
```
stellasia, line 45: Is there a risk to skip a word if `chunk_size=0`?

stellasia: My bad, I meant `chunk_overlap=0`? (though it still makes sense to avoid `chunk_size=0` you're right :) )

NathalieCharbel: In case `chunk_overlap=0` the next chunk will start from where the previous one ended, even it was adjusted.

src/neo4j_graphrag/experimental/components/text_splitters/fixed_size_splitter.py
NathalieCharbel: The reason I used `is.space()` is to handle all types of whitespace characters including space, tabs, newline, etc.

tests/unit/experimental/components/text_splitters/test_fixed_size_splitter.py
```diff
@@ -75,3 +78,139 @@ def test_invalid_chunk_overlap() -> None:
...
+
+@pytest.mark.parametrize(
+    "text, approximate_start, expected_start",
+    [
+        # Case: approximate_start is at word boundary already
+        ("Hello World", 6, 6),
+        # Case: approximate_start is at a whitespace already
+        ("Hello World", 5, 5),
+        # Case: approximate_start is at the middle of word and no whitespace is found
+        ("Hello World", 2, 2),
```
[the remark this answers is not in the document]
stellasia: It seems this is what is done in the '1234567890' example below, then I don't understand why the result is not 0 in this case?

stellasia:
LGTM, just a minor question in the tests. 

Also, remember to add a line in the CHANGELOG (in a `Changed` subsection). And I think you need some ruff formatting for the workflows to pass.

tests/unit/experimental/components/text_splitters/test_fixed_size_splitter.py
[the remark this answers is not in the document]
NathalieCharbel: The idea is to maintain the logic of fixed size splitting as is unless the words are cut in the middle, then go back to the first space we found (i.e. the beginning of the word) **if possible**. If the splitter is based on sentences split going back to the beginning of the text would have made more sense then. 
This test together with the one with '1234567890' are **fallback** cases, where going back to the beginning of the word means start = end, this is why these are rare cases where avoiding splitting words is impossible and the initially proposed end is maintained.

src/neo4j_graphrag/experimental/components/text_splitters/fixed_size_splitter.py
NathalieCharbel: done [https://github.com/neo4j/neo4j-graphrag-python/pull/242/commits/385492b8e43bb250c0bdae23eaed48a5365ed62c](url)

tests/unit/experimental/components/text_splitters/test_fixed_size_splitter.py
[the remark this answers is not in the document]
stellasia: Got it, thanks!

stellasia (approved):

PR #243, opened 2025-01-13
Override neo4j user agent when driver is injected

jonbesga:
# Description

Override neo4j user agent when driver is injected

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Low

Complexity:

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [X] Manual tests

# Checklist

### DIFF
--- a/examples/customize/build_graph/components/writers/custom_writer.py
+++ b/examples/customize/build_graph/components/writers/custom_writer.py
@@ -6,11 +6,12 @@ import neo4j
 from neo4j_graphrag.experimental.components.kg_writer import KGWriter, KGWriterModel
 from neo4j_graphrag.experimental.components.types import LexicalGraphConfig, Neo4jGraph
 from pydantic import validate_call
+from neo4j_graphrag.utils import driver_config
 
 
 class MyWriter(KGWriter):
     def __init__(self, driver: neo4j.Driver) -> None:
-        self.driver = driver
+        self.driver = driver_config.override_user_agent(driver)
 
     @validate_call
     async def run(
--- a/src/neo4j_graphrag/experimental/components/kg_writer.py
+++ b/src/neo4j_graphrag/experimental/components/kg_writer.py
@@ -38,6 +38,7 @@ from neo4j_graphrag.utils.version_utils import (
     get_version,
     is_version_5_23_or_above,
 )
+from neo4j_graphrag.utils import driver_config
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


src/neo4j_graphrag/utils/telemetry.py
```diff
@@ -0,0 +1,8 @@
+import neo4j
```
stellasia, line 1: Remeber to add the copyright notice at the top of the file :)

stellasia:
LGTM. Just one point: I think we're doing things wrong [here](https://github.com/neo4j/neo4j-graphrag-python/blob/main/src/neo4j_graphrag/retrievers/external/qdrant/qdrant.py#L125), by setting `self.driver` again, while it is already defined in the parent class (and only the parent class has this telemetry change, which makes sense). Do you think we can remove this line? (we have a similar pattern in the other external retrievers)

jonbesga:
> LGTM. Just one point: I think we're doing things wrong [here](https://github.com/neo4j/neo4j-graphrag-python/blob/main/src/neo4j_graphrag/retrievers/external/qdrant/qdrant.py#L125), by setting `self.driver` again, while it is already defined in the parent class (and only the parent class has this telemetry change, which makes sense). Do you think we can remove this line? (we have a similar pattern in the other external retrievers)

Good catch! The `super()` is called earlier so you are right. We are overriding. I'll fix it

src/neo4j_graphrag/utils/telemetry.py
jonbesga: 19ddec7e4c682600522c21733199c388dfb597b9

jonbesga:
> LGTM. Just one point: I think we're doing things wrong [here](https://github.com/neo4j/neo4j-graphrag-python/blob/main/src/neo4j_graphrag/retrievers/external/qdrant/qdrant.py#L125), by setting `self.driver` again, while it is already defined in the parent class (and only the parent class has this telemetry change, which makes sense). Do you think we can remove this line? (we have a similar pattern in the other external retrievers)

I think is fine because I'm modifying the driver object that is being passed around. If you check the `id` of the driver object that gets passed to the parent class and the `validated_data.driver_model.driver` that gets reassigned. They are the same.

examples/customize/build_graph/components/writers/custom_writer.py
```diff
@@ -6,11 +6,12 @@
 from neo4j_graphrag.experimental.components.kg_writer import KGWriter, KGWriterModel
 from neo4j_graphrag.experimental.components.types import LexicalGraphConfig, Neo4jGraph
 from pydantic import validate_call
+from neo4j_graphrag.utils import telemetry
```
oskarhane, line 9: I'm not sure this should be classified as telemetry, so this name is a bit misleading. 
I'd say it's more of a driver configuration change. So maybe name the module `driver_config` or similar?

oskarhane (commented):
Great!

Is there a way to verify that this work, in a test or so?

stellasia:
LGTM, I'll let @oskarhane approve when there is an agreement on the naming :)

oskarhane (approved):

PR #268, opened 2025-02-10
LLM interface function calling and implementation for OpenAILLM

willtai:
# Description

This PR enables function calling for the OpenAILLM. It updates the `OpenAILLM` class and also `LLMInterface` type. An example script is also added. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -120,28 +127,35 @@ async def ainvoke(
...
+            choice = response.choices[0]
+            content = choice.message.content or ""
+
+            raw_calls = getattr(choice.message, "tool_calls", None) or []
+            if raw_calls:
+                calls_list = []
+                for c in raw_calls:
+                    calls_list.append(
+                        {
+                            "id": c.id,
```
stellasia, line 149: Does it make sense to use a pydantic model to represent the tool call? I mean something ~ like this:
```
class ToolCall(BaseModel):
   id: str  (?)
   type: str # or a LiteralStr?
   name: str
   arguments: dict[str, Any]
```
or it is too strict?

src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -120,28 +127,35 @@ async def ainvoke(
...
-            input (str): Text sent to the LLM.
-            message_history (Optional[list]): A collection previous messages, with each message having a specific role assigned.
-            system_instruction (Optional[str]): An option to override the llm system message for this invokation.
-
-        Returns:
-            LLMResponse: The response from OpenAI.
-
-        Raises:
-            LLMGenerationError: If anything goes wrong.
+        """
```
stellasia, line 130: The previous docstring is still valid, but now we return content OR tool calls.

examples/question_answering/tool_calling.py
```diff
@@ -0,0 +1,115 @@
...
+    return ", ".join(platforms)
+
+
+query_text = """
+Tell me more about Avatar movies and find
+out what streaming platform it is available on
+"""
+
+
+rag = GraphRAG(retriever=retriever, llm=llm)
```
stellasia, line 93: Maybe this example needs to be updated to remove the GraphRAG part?

examples/question_answering/tool_calling.py
```diff
@@ -0,0 +1,115 @@
...
+                "additionalProperties": False,
+            },
+            "strict": True,
+        },
+    }
+]
+
+llm = OpenAILLM(
+    model_name="gpt-4o",
+    model_params={"temperature": 0, "tools": tool_defs},
```
stellasia, line 78: The more I see it used, the more I wonder why these `model_params` are in the `init` and not the `invoke` method... But that's for a later discussion :)

src/neo4j_graphrag/llm/openai_llm.py
willtai: I think that's a perfect use case. 

https://github.com/neo4j/neo4j-graphrag-python/pull/268/commits/3425d3b01513e844473c501a208bfa4933300df7

Used Literal["function"] because [it should be the only possible value.](https://platform.openai.com/docs/guides/function-calling#:~:text=and%20arguments)-,type,-Type%20of%20tool_call)

src/neo4j_graphrag/llm/openai_llm.py
willtai: https://github.com/neo4j/neo4j-graphrag-python/pull/268/commits/4801ed02b915fc16635b841cec03ee7d194c932f

examples/question_answering/tool_calling.py
willtai: I wanted to show an example of how the tool can be used after being called. I was thinking to keep unless you prefer to have it removed to just show the tool call?

stellasia: I'm a bit uncomfortable with it at the moment, but it's mainly because of our design: using the same `llm` object which has the tool definition (in `model_params`), we have to say in the prompt "do not use any tool", which I feel is the way it is intended to be used. We might have to reconsider this design before using this new feature in that way. WDYT?

PR #269, opened 2025-02-10
Refactors index functions

alexthomas93:
# Description

This PR refactors the code for interacting with vector and hybrid indexes. Key changes include:
- Refactors `get_search_query` to make it compatible with LangChain Neo4j.
- Adds `retrieve_fulltext_index_info` utility function: Retrieves metadata about existing full-text indexes.
- Adds `retrieve_vector_index_info` utility function: Retrieves metadata about existing vector indexes.
- Adds the `effective_search_ratio` parameter for vector and hybrid search.

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Breaking change
- [X] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [X] Unit tests
- [X] E2E tests
- [X] Manual tests

# Checklist

- [X] Documentation has been updated


src/neo4j_graphrag/types.py
```diff
@@ -120,6 +120,7 @@ class VectorSearchModel(BaseModel):
     query_vector: Optional[list[float]] = None
     query_text: Optional[str] = None
     top_k: PositiveInt = 5
+    effective_search_ratio: PositiveInt = 1
```
willtai, line 123: How do we know that the default is `1`? Also are you sure this is a `PositiveInt` and not `float`?

tests/unit/test_neo4j_queries.py
```diff
@@ -147,8 +184,9 @@ def test_hybrid_search_with_properties() -> None:
     properties = ["name", "age"]
     expected = (
         "CALL { "
-        "CALL db.index.vector.queryNodes($vector_index_name, $top_k, $query_vector) "
+        "CALL db.index.vector.queryNodes($vector_index_name, $top_k * $effective_search_ratio, $query_vector) "
```
willtai, line 187: I see, it's used in this way, I was confusing this with `ef` in HNSW. Is this necessarily an integer though?

src/neo4j_graphrag/types.py
alexthomas93: The default is 1 so that the behaviour doesn't change by default.

tests/unit/test_neo4j_queries.py
alexthomas93: Yes it needs to be a positive int as the `numberOfNearestNeighbours` parameter for `db.index.vector.queryNodes` is an int (see [here](numberOfNearestNeighbours))

src/neo4j_graphrag/retrievers/hybrid.py
```diff
@@ -172,25 +176,28 @@ def get_search_results(
...
                 top_k=top_k,
+                effective_search_ratio=effective_search_ratio,
             )
         except ValidationError as e:
             raise SearchValidationError(e.errors()) from e
 
         parameters = validated_data.model_dump(exclude_none=True)
         parameters["vector_index_name"] = self.vector_index_name
         parameters["fulltext_index_name"] = self.fulltext_index_name
+        sanitized_query_text = _remove_lucene_chars(query_text)
```
willtai, line 187: Why do we need to run `_remove_lucene_chars` on the query text? Will these characters affect the lucene query?

alexthomas93:
> > Deprecates the id property returned by all retrievers.
> 
> This will be a breaking change. Why do we make this change in this PR?

We don't necessarily. Happy to remove it and make the change in another PR. It just felt like given I was already refactoring these functions now might be a good time to do it. I think there have also been a few releases now since the warning this property would be removed was added.

src/neo4j_graphrag/retrievers/hybrid.py
alexthomas93: Lucene will throw an error when searching over a full-text index if the query text contains special characters, so sanitisation is necessary in that case. While it’s not strictly required for vector index searches, I applied the same sanitisation there to ensure the same sanitised query was used to search both indexes. Happy to change it so only the full text query is sanitised though if you think that makes more sense?

src/neo4j_graphrag/neo4j_queries.py
```diff
@@ -172,76 +221,116 @@ def _get_filtered_vector_query(
...
+            raise ValueError(f"Search type is not supported: {search_type}")
+        fallback_return = (
+            f"RETURN node {{ .*, `{embedding_node_property}`: null }} AS node, "
+            "labels(node) AS nodeLabels, elementId(node) AS elementId, score"
+        )
+    elif index_type == IndexType.RELATIONSHIP:
+        if filters:
+            raise Exception("Filters are not supported for relationship indexes")
+        if search_type == SearchType.HYBRID:
+            raise Exception("Hybrid search is not supported for relationship indexes")
```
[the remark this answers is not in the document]
alexthomas93: There doesn’t seem to be a technical limitation preventing hybrid search from working with relationships, as they support both vector and full-text indexes. However, it’s not currently implemented in this repository or the LangChain repo, which suggests it may not be a highly requested feature. I did try implementing it briefly though when I started working on this PR but I realised it'd probably make this already large PR a lot larger. We could consider adding it to the backlog for future implementation or wait to see if there’s demand for it?

src/neo4j_graphrag/types.py
```diff
@@ -157,6 +159,24 @@ class SearchType(str, Enum):
...
 
 
+class IndexType(str, Enum):
+    """Enumerator of the index types."""
+
+    NODE = "NODE"
+    RELATIONSHIP = "RELATIONSHIP"
+
+
+class DistanceStrategy(str, Enum):
```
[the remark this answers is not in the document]
alexthomas93: Sorry no it isn't. I ported it over early on when working on the PR then forgot to remove it when I realised it wasn't actually needed. Have removed it now in [cc8214b](https://github.com/neo4j/neo4j-graphrag-python/pull/269/commits/cc8214b7a8c024be7fc8f59d6603d78d5e19ed72)

willtai:
> We don't necessarily. Happy to remove it and make the change in another PR. It just felt like given I was already refactoring these functions now might be a good time to do it. I think there have also been a few releases now since the warning this property would be removed was added.

Don't think we need it for this PR. We can remove it in a separate PR, wdyt?

src/neo4j_graphrag/retrievers/hybrid.py
willtai: I think it makes more sense to only apply it to the full-text index query then. The characters could have meaning in the vector search query

alexthomas93: Updated in [138f2fe](https://github.com/neo4j/neo4j-graphrag-python/pull/269/commits/138f2fe0135df52bbbafa249c755ff83a1888884)

alexthomas93:
> e773de4

Makes sense. I'll add something to the backlog to remind us to deprecated it at some point in the future. Have re-added it in [e773de4](https://github.com/neo4j/neo4j-graphrag-python/pull/269/commits/e773de4c5359b0b7048b28584ba0cdf5b637af3c). PR should no longer have any breaking changes in it.

src/neo4j_graphrag/neo4j_queries.py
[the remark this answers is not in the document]
stellasia: Yes sure, we can wait for demand, I was just curious if this was possible or not. Thanks!

src/neo4j_graphrag/neo4j_queries.py
```diff
@@ -18,11 +18,20 @@
...
+from neo4j_graphrag.types import EntityType, SearchType
 
-VECTOR_INDEX_QUERY = (
-    "CALL db.index.vector.queryNodes($vector_index_name, $top_k, $query_vector) "
-    "YIELD node, score"
+NODE_VECTOR_INDEX_QUERY = (
+    "CALL db.index.vector.queryNodes"
+    "($vector_index_name, $top_k * $effective_search_ratio, $query_vector) "
+    "YIELD node, score "
+    "WITH node, score LIMIT $top_k"
```
willtai, line 27: Do we need `LIMIT $top_k` for both this and the REL_VECTOR_INDEX_QUERY?

alexthomas93: Yes otherwise you get `top_k * $effective_search_ratio` results returned rather than `top_k` results

willtai (approved):
☄️

PR #273, opened 2025-02-18
Neo4j message history

alexthomas93:
# Description

This pull request introduces a Neo4jMessageHistory class for storing LLM message histories in a Neo4j database and updates existing classes to use it. It also provides an InMemoryMessageHistory class for in-memory storage, along with a new MessageHistory abstract base class that both implementations inherit from. Lastly, it includes examples and updates the documentation to illustrate how these classes can be used.

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Breaking change
- [X] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [X] Unit tests
- [X] E2E tests
- [X] Manual tests

# Checklist

- [X] Documentation has been updated
- [X] Examples have been updated

### DIFF
--- a/examples/customize/llms/custom_llm.py
+++ b/examples/customize/llms/custom_llm.py
@@ -1,9 +1,10 @@
 import random
 import string
-from typing import Any, Optional
+from typing import Any, List, Optional, Union
 
 from neo4j_graphrag.llm import LLMInterface, LLMResponse
 from neo4j_graphrag.llm.types import LLMMessage
+from neo4j_graphrag.message_history import MessageHistory
 
 
 class CustomLLM(LLMInterface):
@@ -15,7 +16,7 @@ class CustomLLM(LLMInterface):
     def invoke(
         self,
         input: str,
-        message_history: Optional[list[LLMMessage]] = None,
+        message_history: Optional[Union[List[LLMMessage], MessageHistory]] = None,
         system_instruction: Optional[str] = None,
     ) -> LLMResponse:
         content: str = (
@@ -26,7 +27,7 @@ class CustomLLM(LLMInterface):
     async def ainvoke(
         self,
         input: str,
-        message_history: Optional[list[LLMMessage]] = None,
+        message_history: Optional[Union[List[LLMMessage], MessageHistory]] = None,
         system_instruction: Optional[str] = None,
     ) -> LLMResponse:
         raise NotImplementedError()
--- a/src/neo4j_graphrag/generation/graphrag.py
+++ b/src/neo4j_graphrag/generation/graphrag.py
@@ -131,7 +131,7 @@ class GraphRAG:
             raise SearchValidationError(e.errors())
         if isinstance(message_history, MessageHistory):
             message_history = message_history.messages
-        query = self.build_query(validated_data.query_text, message_history)
+        query = self._build_query(validated_data.query_text, message_history)
         retriever_result: RetrieverResult = self.retriever.search(
             query_text=query, **validated_data.retriever_config
         )
@@ -151,16 +151,14 @@ class GraphRAG:
             result["retriever_result"] = retriever_result
         return RagResultModel(**result)
 
-    def build_query(
+    def _build_query(
         self,
         query_text: str,
-        message_history: Optional[Union[List[LLMMessage], MessageHistory]] = None,
+        message_history: Optional[List[LLMMessage]] = None,
     ) -> str:
         summary_system_message = "You are a summarization assistant. Summarize the given text in no more than 300 words."
         if message_history:
-            if isinstance(message_history, MessageHistory):
-                message_history = message_history.messages
-            summarization_prompt = self.chat_summary_prompt(
+            summarization_prompt = self._chat_summary_prompt(
                 message_history=message_history
             )
             summary = self.llm.invoke(
@@ -170,11 +168,7 @@ class GraphRAG:
             return self.conversation_prompt(summary=summary, current_query=query_text)
         return query_text
 
-    def chat_summary_prompt(
-        self, message_history: Union[List[LLMMessage], MessageHistory]
-    ) -> str:
-        if isinstance(message_history, MessageHistory):
-            message_history = message_history.messages
+    def _chat_summary_prompt(self, message_history: List[LLMMessage]) -> str:
         message_list = [
             f"{message['role']}: {message['content']}" for message in message_history
         ]
--- a/src/neo4j_graphrag/generation/graphrag.py
+++ b/src/neo4j_graphrag/generation/graphrag.py
@@ -28,7 +28,7 @@ from neo4j_graphrag.generation.prompts import RagTemplate
 from neo4j_graphrag.generation.types import RagInitModel, RagResultModel, RagSearchModel
 from neo4j_graphrag.llm import LLMInterface
 from neo4j_graphrag.llm.types import LLMMessage
-from neo4j_graphrag.message_history import InMemoryMessageHistory, MessageHistory
+from neo4j_graphrag.message_history import MessageHistory
 from neo4j_graphrag.retrievers.base import Retriever
 from neo4j_graphrag.types import RetrieverResult
 
@@ -129,8 +129,8 @@ class GraphRAG:
             )
         except ValidationError as e:
             raise SearchValidationError(e.errors())
-        if isinstance(message_history, list):
-            message_history = InMemoryMessageHistory(messages=message_history)
+        if isinstance(message_history, MessageHistory):
+            message_history = message_history.messages
         query = self.build_query(validated_data.query_text, message_history)
         retriever_result: RetrieverResult = self.retriever.search(
             query_text=query, **validated_data.retriever_config
@@ -143,7 +143,7 @@ class GraphRAG:
         logger.debug(f"RAG: prompt={prompt}")
         answer = self.llm.invoke(
             prompt,
-            message_history.messages if message_history else None,
+            message_history,
             system_instruction=self.prompt_template.system_instructions,
         )
         result: dict[str, Any] = {"answer": answer.content}
@@ -158,8 +158,8 @@ class GraphRAG:
     ) -> str:
         summary_system_message = "You are a summarization assistant. Summarize the given text in no more than 300 words."
         if message_history:
-            if isinstance(message_history, list):
-                message_history = InMemoryMessageHistory(messages=message_history)
+            if isinstance(message_history, MessageHistory):
+                message_history = message_history.messages
             summarization_prompt = self.chat_summary_prompt(
                 message_history=message_history
             )
@@ -173,11 +173,10 @@ class GraphRAG:
     def chat_summary_prompt(
         self, message_history: Union[List[LLMMessage], MessageHistory]
     ) -> str:
-        if isinstance(message_history, list):
-            message_history = InMemoryMessageHistory(messages=message_history)
+        if isinstance(message_history, MessageHistory):
+            message_history = message_history.messages
         message_list = [
-            f"{message['role']}: {message['content']}"
-            for message in message_history.messages
+            f"{message['role']}: {message['content']}" for message in message_history
         ]
         history = "\n".join(message_list)
         return f"""
--- a/src/neo4j_graphrag/generation/graphrag.py
+++ b/src/neo4j_graphrag/generation/graphrag.py
@@ -103,7 +103,8 @@ class GraphRAG:
 
         Args:
             query_text (str): The user question.
-            message_history (Optional[list]): A collection previous messages, with each message having a specific role assigned.
+            message_history (Optional[Union[List[LLMMessage], MessageHistory]]): A collection previous messages,
+                with each message having a specific role assigned.
             examples (str): Examples added to the LLM prompt.
             retriever_config (Optional[dict]): Parameters passed to the retriever.
                 search method; e.g.: top_k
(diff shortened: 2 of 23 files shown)


src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -127,6 +129,8 @@ def search(
             )
         except ValidationError as e:
             raise SearchValidationError(e.errors())
+        if isinstance(message_history, MessageHistory):
```
stellasia, line 132: Not for now, but do you see any interest to do the other way around in the future, something like this: 
```
if not isinstance(message_history, MessageHistory):
   message_history = MessageHistory.parse(message_history)
```

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -148,10 +152,14 @@ def search(
...
 
     def build_query(
-        self, query_text: str, message_history: Optional[list[LLMMessage]] = None
+        self,
+        query_text: str,
+        message_history: Optional[Union[List[LLMMessage], MessageHistory]] = None,
     ) -> str:
         summary_system_message = "You are a summarization assistant. Summarize the given text in no more than 300 words."
         if message_history:
+            if isinstance(message_history, MessageHistory):
```
stellasia, line 161: Can this happen? Since we have the same check in `search` before we call this method?

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -162,10 +170,13 @@ def build_query(
             return self.conversation_prompt(summary=summary, current_query=query_text)
         return query_text
 
-    def chat_summary_prompt(self, message_history: list[LLMMessage]) -> str:
+    def chat_summary_prompt(
+        self, message_history: Union[List[LLMMessage], MessageHistory]
+    ) -> str:
+        if isinstance(message_history, MessageHistory):
```
stellasia, line 176: Same question here, we already have this check in the method calling this one.

examples/customize/llms/custom_llm.py
```diff
@@ -15,7 +16,7 @@ def __init__(
     def invoke(
         self,
         input: str,
-        message_history: Optional[list[LLMMessage]] = None,
+        message_history: Optional[Union[List[LLMMessage], MessageHistory]] = None,
```
willtai, line 19: Why the decision to have 

```
message_history: Union[List[LLMMessage], MessageHistory]
    )
```

throughout this PR? Seems like we don't only check for `MessageHistory` type? Perhaps I've missed something

src/neo4j_graphrag/generation/graphrag.py
alexthomas93: What do you see as the advantage of doing it that way over the current approach?

src/neo4j_graphrag/generation/graphrag.py
alexthomas93: No it can't but I wanted to be safe given `build_query` is a public method. I can remove these lines if you like but if I do would it be also worth changing the method name to `_build_query`?

src/neo4j_graphrag/message_history.py
```diff
@@ -0,0 +1,212 @@
...
+    Neo4jDriverModel,
+    Neo4jMessageHistoryModel,
+)
+
+CREATE_SESSION_NODE_QUERY = "MERGE (s:`{node_label}` {{id:$session_id}})"
+
+CLEAR_SESSION_QUERY = (
+    "MATCH (s:`{node_label}`)-[:LAST_MESSAGE]->(last_message) "
+    "WHERE s.id = $session_id MATCH p=(last_message)<-[:NEXT]-() "
+    "WITH p, length(p) AS length ORDER BY length DESC LIMIT 1 "
```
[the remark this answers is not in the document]
alexthomas93: Yes, there should only ever be a single `LAST_MESSAGE` relationship in a given chain. The node containing the session ID maintains this relationship, pointing to the most recent message. As a result, each chain should have exactly one `LAST_MESSAGE` relationship.

No, multiple chains cannot share the same message ID. I think that's why this query is written this way. Each message chain is expected to be a simple sequence of messages, linked by `NEXT` relationships that connect each message to its predecessor and successor. The final message in the chain is then linked to the session ID node via the `LAST_MESSAGE` relationship. So I don't think we have to worry about other connections between messages creating other chains here.

src/neo4j_graphrag/generation/graphrag.py
stellasia: Not saying it's better, I just like the `MessageHistory` interface, and I feel like the `InMemory` implementation could replace the `List[LLMMessage]`. Maybe we could imagine adding more features in theses classes (like `def to_anthropic(self)`), but I'm not sure it's better compared to what we have at the moment.

src/neo4j_graphrag/generation/graphrag.py
stellasia: It makes sense to make these methods private actually.

examples/customize/llms/custom_llm.py
alexthomas93: It's so we don't introduce a breaking change. Previously all of these classes took `List[LLMMessage]`. The `messages` property of the `MessageHistory` class is also of type `List[LLMMessage]`. So when the user is using `MessageHistory` you can just get the messages from the `messages` property and everything works as it did before.

src/neo4j_graphrag/message_history.py
```diff
@@ -0,0 +1,212 @@
...
+        node_label (str, optional): Label used for session nodes in Neo4j. Defaults to "Session".
+        window (Optional[PositiveInt], optional): Number of previous messages to return when retrieving messages.
+
+    """
+
+    def __init__(
+        self,
+        session_id: Union[str, int],
+        driver: neo4j.Driver,
+        node_label: str = "Session",
```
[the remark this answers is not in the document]
alexthomas93: Do we want the relationship types to be customisable? I think it makes sense to make the message node type customisable though. I'll add something in for this.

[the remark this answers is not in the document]
alexthomas93: Alternatively I could remove this completely and not give users any option over the node label names?

src/neo4j_graphrag/message_history.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+    def clear(self) -> None:
+        """Clear the message history."""
+        self._driver.execute_query(
+            query_=CLEAR_SESSION_QUERY.format(node_label=self._node_label),
+            parameters_={"session_id": self._session_id},
+        )
+
+    def __del__(self) -> None:
```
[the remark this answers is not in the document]
alexthomas93: Removed in [baa0bed](https://github.com/neo4j/neo4j-graphrag-python/pull/273/commits/baa0bed3795346c1156846af0ae1acbd846a8475)

src/neo4j_graphrag/generation/graphrag.py
alexthomas93: That make sense. Do we want to think about deprecating support for using `List[LLMMessage]` with these classes and encouraging users to move over to the `MessageHistory` classes? I think it that case it'd make sense to re-write the LLM and GraphRAG classes to use the `MessageHistory` interface by default.

src/neo4j_graphrag/generation/graphrag.py
alexthomas93: Updated in [b2e94d8](https://github.com/neo4j/neo4j-graphrag-python/pull/273/commits/b2e94d8556f1e7d3fdb5010daa32f614746a5d40)

src/neo4j_graphrag/generation/graphrag.py
alexthomas93: Updated in [b2e94d8](https://github.com/neo4j/neo4j-graphrag-python/pull/273/commits/b2e94d8556f1e7d3fdb5010daa32f614746a5d40)

src/neo4j_graphrag/message_history.py
```diff
@@ -0,0 +1,209 @@
...
+        messages (Optional[List[LLMMessage]]): List of messages to initialize the history with. Defaults to None.
+
+    """
+
+    def __init__(self, messages: Optional[List[LLMMessage]] = None) -> None:
+        self._messages = messages or []
+
+    @property
+    def messages(self) -> List[LLMMessage]:
+        return self._messages
```
willtai, line 100: I think this will not be a thread-safe, I think we should at least make a note of it in the docstring

Either that or consider implementing a lock:

```python
import threading

class InMemoryMessageHistory(MessageHistory):
    def __init__(self, messages: Optional[List[LLMMessage]] = None) -> None:
        self._lock = threading.Lock()
        self._messages = messages or []

    @property
    def messages(self) -> List[LLMMessage]:
        with self._lock:
            # This or maybe a copy?
            return self._messages

    def add_message(self, message: LLMMessage) -> None:
        with self._lock:
            self._messages.append(message)

    def clear(self) -> None:
        with self._lock:
            self._messages = []
```

so that only one thread mutates `self._messages` at a time. A bit simplistic but could work for now.

Also happy to go with the "Not thread-safe" documentation option

alexthomas93: Updated in [c3c28c8](https://github.com/neo4j/neo4j-graphrag-python/pull/273/commits/c3c28c897d0660b8f843022d3dba2165796bb002)

willtai (approved):

src/neo4j_graphrag/generation/graphrag.py
stellasia: Yes I think we can consider that for a later improvement.

src/neo4j_graphrag/message_history.py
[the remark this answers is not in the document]
stellasia: It's a good option here but that would introduce a breaking change in langchain, is it ok ?

src/neo4j_graphrag/message_history.py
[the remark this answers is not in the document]
stellasia: Hum, this query is weird (unless I'm not building the graph properly). But if I use:
```
CREATE (s:Session {id: "1"})
CREATE (m1:Message)
CREATE (m2:Message)<-[:NEXT]-(m1)
CREATE (m3:Message)<-[:NEXT]-(m2)
CREATE (s)-[:LAST_MESSAGE]->(m3)
```

and then run:

```
MATCH (s:Session)-[:LAST_MESSAGE]->(last_message)
WHERE s.id = "1" MATCH p=(last_message)<-[:NEXT]-()
WITH p, length(p) AS length ORDER BY length DESC LIMIT 1
UNWIND nodes(p) as node DETACH DELETE node
```

only two nodes are deleted, because the path only contains the last message and the previous one, so it only deletes two nodes and leaves the other messages (that won't be attached to anything anymore since the last message was deleted and so was the `LAST_MESSAGE` relationship)


Aren't we missing the variable length pattern here:
```
    "WHERE s.id = $session_id MATCH p=(last_message)<-[:NEXT*0..{window_size}]-() "
```
in which case I would understand the order by and limit.

src/neo4j_graphrag/message_history.py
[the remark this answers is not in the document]
alexthomas93: I've removed `node_label` as an option from the class itself here [8ae9423](https://github.com/neo4j/neo4j-graphrag-python/pull/273/commits/8ae9423d7b09563c260cce04367e4b4c951f65d9) but left the queries as they are so they can still be used by the LangChain version. I wasn't actually planning on directly adapting this class into LangChain as it's already very simple and maintaining the adapter code probably wouldn't save us any maintenance load.

src/neo4j_graphrag/message_history.py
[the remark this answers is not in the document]
alexthomas93: Just tested again with both this and the LangChain implementation and you're right! Will update the query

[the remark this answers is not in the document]
alexthomas93: Have updated the query in [63b549e](https://github.com/neo4j/neo4j-graphrag-python/pull/273/commits/63b549e08a3a2b74063441c8a733727c3220f873). Hopefully it should work now!

src/neo4j_graphrag/message_history.py
```diff
@@ -0,0 +1,212 @@
...
+    Neo4jMessageHistoryModel,
+)
+
+CREATE_SESSION_NODE_QUERY = "MERGE (s:`{node_label}` {{id:$session_id}})"
+
+CLEAR_SESSION_QUERY = (
+    "MATCH (s:`{node_label}`) "
+    "WHERE s.id = $session_id "
+    "OPTIONAL MATCH p=(s)-[:LAST_MESSAGE]->(:Message)<-[:NEXT*0..]-(:Message) "
+    "WITH CASE WHEN p IS NULL THEN [s] ELSE nodes(p) + [s] END AS nodes "
```
stellasia, line 36: Isn't `s` already included in `nodes(p)` when `p` is not null?

alexthomas93: Nope if s doesn't have any messages attached the optional match clause doesn't return anything, as such s won't be deleted, you need the case here to deal with that situation.

stellasia: Yes if `p` is null I understand, but in the other case we have `nodes(p) + [s]`, but isn't `s` already in the path?

alexthomas93: Apologies I see what you mean now. Yeah you're right. Updated in [e84230a](https://github.com/neo4j/neo4j-graphrag-python/pull/273/commits/e84230ad0a9557a50c0daa27759560a530280594)

stellasia (approved):

PR #303, opened 2025-03-12
chore: Fallback to Qdrant point id

### LINKED ISSUE #294: Matching between neo4j and qdrant ids fails
Hello, 

in the QdrantNeo4jRetriever class (qdrant.py file) in line 214, it tries to append:

`[f"{point.payload[self.id_property_external]}", point.score]`

That means, even if the id_property_external from qdrant is its "id", it will look for `point.payload("id")`. However, in the case of 'id', it is a separate attribute of a qdrant point object and not contained in the payload itself ({"id": ..., "payload": {"page_content": ..., "vector": ..., "metadata": ..., ...}). So the correct way to retrieve it would be `point.id` instead (for all other cases the current way should work).

Error message:
```
File "C:Users\...\Lib\site-packages\neo4j_graphrag\retrievers\external\qdrant\qdrant.py", line 214, in get_search_results
    [f"{point.payload[self.id_property_external]}", point.score]
        ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'id'
```

Anush008:
# Description

Resolves #294.

This PR updates `QdrantNeo4jRetriever` to use the point's ID as a fallback if the dedicated ID field is not found in the payload.

### DIFF
--- a/src/neo4j_graphrag/retrievers/external/qdrant/qdrant.py
+++ b/src/neo4j_graphrag/retrievers/external/qdrant/qdrant.py
@@ -211,7 +211,7 @@ class QdrantNeo4jRetriever(ExternalRetriever):
         for point in points:
             assert point.payload is not None
             result_tuples.append(
-                [f"{point.payload[self.id_property_external]}", point.score]
+                [point.payload.get(self.id_property_external, point.id), point.score]
             )
 
         search_query = get_match_query(


stellasia (approved):
Thank you for the super fast PR!

PR #304, opened 2025-03-13
Pipeline streaming

stellasia:
# Description

Add a `Pipeline.stream` method that yields events about pipeline progress leveraging the callback mechanism already in place:

- Make `pipeline.callbacks` a list
- Implement a new `Pipeline.stream` method with the following logic:
   - Create a new callback that put events in an `asyncio.Queue` and append it to the callbcks list
   - Create asyncion task to `run` pipeline in background
   - Loop until the run task is done and the event queue is empty:
      - check run task status (done or pending)
      - yield event from queue if any

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: High

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/pipeline/pipeline.py
+++ b/src/neo4j_graphrag/experimental/pipeline/pipeline.py
@@ -51,6 +51,8 @@ from neo4j_graphrag.experimental.pipeline.types.context import RunContext
 from neo4j_graphrag.experimental.pipeline.notification import (
     EventCallbackProtocol,
     Event,
+    PipelineEvent,
+    EventType,
 )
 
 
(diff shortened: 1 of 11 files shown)


src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -412,6 +416,59 @@ def validate_parameter_mapping_for_task(self, task: TaskPipelineNode) -> bool:
     async def get_final_results(self, run_id: str) -> dict[str, Any]:
         return await self.final_results.get(run_id)  # type: ignore[no-any-return]
 
+    async def stream(self, data: dict[str, Any]) -> AsyncGenerator[Event, None]:
```
willtai, line 419:
```python
data: dict[str, Any]
```

I wonder if it's about time we have a data type for pipeline inputs. Or would that be overkill for this PR?

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -412,6 +416,59 @@ def validate_parameter_mapping_for_task(self, task: TaskPipelineNode) -> bool:
...
+                    return_when=asyncio.FIRST_COMPLETED,
+                )
+
+                is_run_task_running = run_task not in done
+                is_queue_empty = event_queue.empty()
+
+                for event_future in done:
+                    if event_future == run_task:
+                        continue
+                    yield event_future.result()  # type: ignore
```
willtai, line 461: What errors do you get without `type: ignore`?

src/neo4j_graphrag/experimental/pipeline/pipeline.py
stellasia: We can't have a fixed data type because it depends on the components: SimpleKGPipeline and a RAGPipeline (if we ever add it) won't have the same inputs. I think it makes sense for the "template" pipeline though, and we are close to it with the "SimpleKGPipelineConfig", but for the generic pipeline Idk how we could achieve it.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
stellasia: >  Incompatible types in "yield" (actual type "BaseModel", expected type "Event")

I have no clue where the `BaseModel` comes from here, the queue is defined as `Queue[Event]` so it should always return an `Event` (and this is what I observed in my tests)

alexthomas93 (approved):

oskarhane (approved):
Looks like a good change! 🍶

willtai (approved):

PR #306, opened 2025-03-14
Replace pygraphviz with neo4j-viz for graph visualization

oskarhane:
Neo4j has its own viz library, let's use that!

- Replace pygraphviz imports with neo4j-viz in pipeline.py
- Update draw() method to save visualizations as HTML files
- Implement get_neo4j_viz_graph() method for interactive visualizations
- Update example in visualization.py to use HTML output
- Update tests to work with neo4j-viz implementation
- Add color coding for better visual distinction

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -182,40 +189,100 @@ def show_as_dict(self) -> dict[str, Any]:
...
+        """
+        G = self.get_neo4j_viz_graph(hide_unused_outputs)
+        if path.endswith(".html"):
+            # Save as HTML file
+            with open(path, "w") as f:
+                f.write(G.render()._repr_html_())
+        else:
+            # For other formats, we'll use the render method and save the image
+            G.render()
+            # Note: neo4j-viz doesn't support direct saving to image formats
```
stellasia, line 213: Is the comment saying that it can't save to PNG? Because code in the 'else' looks very similar to the one in the former 'if' block to me.

src/neo4j_graphrag/embeddings/mistral.py
```diff
@@ -24,7 +24,9 @@
 try:
     from mistralai import Mistral
 except ImportError:
-    Mistral = None  # type: ignore
+    # Define placeholder type for type checking
+    class Mistral:  # type: ignore
```
stellasia, line 28: Is this better compared to the previous approach? (out of curiosity)

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -182,40 +189,100 @@ def show_as_dict(self) -> dict[str, Any]:
...
     def draw(
-        self, path: str, layout: str = "dot", hide_unused_outputs: bool = True
+        self, path: str, layout: str = "force", hide_unused_outputs: bool = True
     ) -> Any:
-        G = self.get_pygraphviz_graph(hide_unused_outputs)
-        G.layout(layout)
-        G.draw(path)
+        """Draw the pipeline graph using neo4j-viz.
 
-    def get_pygraphviz_graph(self, hide_unused_outputs: bool = True) -> pgv.AGraph:
```
stellasia, line 191: I don't remember why we haven't made this method private, but since we are removing it we should mention it in the changelog.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -182,40 +189,100 @@ def show_as_dict(self) -> dict[str, Any]:
...
+                f.write(G.render()._repr_html_())
+        else:
+            # For other formats, we'll use the render method and save the image
+            G.render()
+            # Note: neo4j-viz doesn't support direct saving to image formats
+            # If image format is needed, consider using a screenshot or other methods
+            with open(path, "w") as f:
+                f.write(G.render()._repr_html_())
+
+    def get_neo4j_viz_graph(self, hide_unused_outputs: bool = True) -> NeoVizGraph:
```
stellasia, line 218: And maybe make this method private?

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -182,40 +189,100 @@ def show_as_dict(self) -> dict[str, Any]:
...
+        Returns:
+            NeoVizGraph: The neo4j-viz visualization graph.
+        """
+        if not HAS_NEO4J_VIZ:
             raise ImportError(
-                "Could not import pygraphviz. "
-                "Follow installation instruction in pygraphviz documentation "
-                "to get it up and running on your system."
+                "Could not import neo4j-viz. "
+                "Install it with 'pip install neo4j-viz'."
```
stellasia, line 230: I think using extra groups instead reduces a bit dependency issues:

```suggestion
                'Install it with 'pip install "neo4j-graphrag[experimental]".'
```

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -182,40 +189,100 @@ def show_as_dict(self) -> dict[str, Any]:
...
             )
-            G.add_node(
-                n,
-                node_type="component",
-                shape="rectangle",
-                label=f"{node.component.__class__.__name__}: {n}({comp_inputs})",
+            node_ids[n] = node_counter
+            nodes.append(
+                Node(
+                    id=node_counter,
```
stellasia, line 248: According to [the doc](https://neo4j.com/docs/nvl-python/preview/api-reference/node.html), it looks like `id` can be a string? That could simplify the id and node matching part.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -224,13 +291,30 @@ def get_pygraphviz_graph(self, hide_unused_outputs: bool = True) -> pgv.AGraph:
...
+                    )
+
+        # Filter unused outputs if requested
         if hide_unused_outputs:
-            for n in G.nodes():
-                if n.attr["node_type"] == "output" and G.out_degree(n) == 0:  # type: ignore
-                    G.remove_node(n)
-        return G
+            used_nodes = set()
+            for rel in relationships:
```
stellasia, line 310: I think there is a logic issue here: we do not want to filter out isolated nodes (this should not happen I think?), but node of type "output" with no outgoing relationship.

tests/unit/experimental/pipeline/test_pipeline.py
```diff
@@ -379,39 +379,48 @@ async def test_pipeline_async() -> None:
...
-    g = pipe.get_pygraphviz_graph()
-    # 3 nodes:
+    g = pipe.get_neo4j_viz_graph()
+    # 4 nodes:
     #   - 2 components 'a' and 'b'
-    #   - 1 output 'a.result'
-    assert len(g.nodes()) == 3
-    g = pipe.get_pygraphviz_graph(hide_unused_outputs=False)
+    #   - 2 outputs 'a.result' and 'b.result' (neo4j-viz implementation includes both)
+    assert len(g.nodes) == 4
```
stellasia, line 393: Related to my comment above about logic, the test should not need to be updated.

src/neo4j_graphrag/embeddings/mistral.py
oskarhane: Not sure why this slipped in. Removed.

src/neo4j_graphrag/experimental/pipeline/neo4j_viz.pyi
```diff
@@ -0,0 +1,39 @@
+from typing import Any, Dict, List, Optional, Union
```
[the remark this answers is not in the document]
oskarhane: Thanks

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -198,53 +200,132 @@ def show_as_dict(self) -> dict[str, Any]:
...
+                    rel = Relationship(  # type: ignore
+                        source=node_ids[source_output_node],
+                        target=node_ids[component_name],
+                        caption=param,
+                        properties={"type": "CONNECTS_TO"},
+                    )
+                    relationships.append(rel)
+
+        # Cast the constructor to Any, then cast the result back to VisualizationGraph
+        viz_graph = cast(Any, VisualizationGraph)(
```
stellasia, line 313: What's the mypy error without this cast?

stellasia: Also, isn't the `pyi` file supposed to help mypy here? Otherwise why do we have this file?

stellasia:
Just realized we can remove pygraphviz install from the GH actions (`pr-e2e-tests` and maybe others)

src/neo4j_graphrag/experimental/pipeline/pipeline.py
oskarhane: yep, you're right. Fixed in latest commit

stellasia (approved):
🕸️

PR #307, opened 2025-03-14
Add CypherRetriever for parameterized Cypher queries

oskarhane:
This PR adds a new CypherRetriever that enables direct database access through parameterized Cypher queries. Key features include:
- Type-safe parameter validation
- Support for optional parameters
- Custom result formatting
- Documentation and examples

# Description

> **Note**
>
>
>

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/retrievers/cypher.py
+++ b/src/neo4j_graphrag/retrievers/cypher.py
@@ -1,0 +1,268 @@
+#  Copyright (c) "Neo4j"
+#  Neo4j Sweden AB [https://neo4j.com]
+#  #
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  #
+#      https://www.apache.org/licenses/LICENSE-2.0
+#  #
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from __future__ import annotations
+
+import logging
+import re
+from typing import Any, Callable, Dict, List, Optional, Union
+
+import neo4j
+from neo4j.exceptions import CypherSyntaxError
+from pydantic import ValidationError
+
+from neo4j_graphrag.exceptions import RetrieverInitializationError, SearchValidationError
+from neo4j_graphrag.retrievers.base import Retriever
+from neo4j_graphrag.types import (
+    CypherParameterDefinition,
+    CypherParameterType,
+    CypherRetrieverModel,
+    CypherSearchModel,
+    Neo4jDriverModel,
+    RawSearchResult,
+    RetrieverResultItem,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class CypherRetriever(Retriever):
+    """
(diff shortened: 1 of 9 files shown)


src/neo4j_graphrag/retrievers/cypher.py
```diff
@@ -0,0 +1,379 @@
...
+            description = param_def.get("description", "")
+            required = param_def.get("required", True)
+
+            try:
+                param_definitions[param_name] = CypherParameterDefinition(
+                    type=param_type, description=description, required=required
+                )
+            except ValidationError as e:
+                raise RetrieverInitializationError(
+                    [
```
stellasia, line 112: I've never used this syntax but I think `from e` will show the same informations (loc, etc)? In such case we do not need `ErrorDestail`, which looks like an internal pydantic thing.

src/neo4j_graphrag/retrievers/cypher.py
```diff
@@ -0,0 +1,379 @@
...
+                            loc=("parameters", param_name),
+                            msg=f"Required parameter '{param_name}' is missing",
+                            type="value_error.missing",
+                            input=None,
+                        )
+                    ]
+                )
+
+        # Validate the type of each parameter
+        for param_name, param_value in parameters.items():
```
stellasia, line 242: Can this be merged into a single loop?

src/neo4j_graphrag/types.py
```diff
@@ -312,3 +312,37 @@ def validate_session_id(cls, v: Union[str, int]) -> Union[str, int]:
...
+    INTEGER = "integer"
+    BOOLEAN = "boolean"
+    ARRAY = "array"
+
+
+class CypherParameterDefinition(BaseModel):
+    """Definition of a Cypher query parameter."""
+
+    type: CypherParameterType
+    description: str
```
stellasia, line 331: Can the description be optional? We're not using it (yet)

src/neo4j_graphrag/retrievers/cypher.py
```diff
@@ -0,0 +1,379 @@
...
+
+    Raises:
+        RetrieverInitializationError: If validation of the input arguments fail.
+    """
+
+    def __init__(
+        self,
+        driver: neo4j.Driver,
+        query: str,
+        parameters: Dict[str, Dict[str, Any]],
```
stellasia, line 93: Maybe we can use `list[CypherParameterDefinition]` with the `@validate_call` decorator (see for instance [here](https://github.com/neo4j/neo4j-graphrag-python/blob/main/src/neo4j_graphrag/experimental/components/text_splitters/fixed_size_splitter.py#L94))

tests/unit/retrievers/test_cypher.py
```diff
@@ -0,0 +1,299 @@
...
+
+from neo4j_graphrag.exceptions import (
+    RetrieverInitializationError,
+    SearchValidationError,
+)
+from neo4j_graphrag.retrievers.cypher import CypherRetriever
+from neo4j_graphrag.types import RetrieverResultItem
+
+
+class TestCypherRetriever(unittest.TestCase):
```
stellasia, line 31: A bit nitpicking here but we do not use class-based tests in this code base, only functions. So unless there is a strong reason to use a class, I'd prefer we stick to using functions :)

PR #310, opened 2025-03-18
Add SpaCy Semantic match resolver for KG Builder

NathalieCharbel:
# Description

This PR introduces a new `SpaCySemanticMatchResolver` component that leverages spaCy embeddings to merge nodes based on textual similarity. It expands upon the existing resolver framework (which was previously limited to exact matching) to support semantic comparisons of a given set of textual properties (the "name" property is still the default choice) via cosine similarity of embeddings. The main motivation behind this is to provide a more robust entity resolution approach that avoid missing merges (e.g., Two `Person` nodes whose `name` properties are respectively "John Smith" and "Jonathan Smith" and `SSN` properties are the same, etc.), and false merges (e.g., Two `Person` nodes with exact `name` properties  and different `SSN` properties, etc.). 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist


examples/customize/build_graph/components/resolvers/spacy_entity_resolver_pre_filter.py
```diff
@@ -0,0 +1,36 @@
+"""The SpaCySemanticMatchResolver merges nodes with same label
+and similar textual properties (by default using the "name" property).
+
+If the resolution is intended to be applied only on some nodes, for instance nodes that
```
stellasia, line 4: Do we need both examples? Maybe the second one (with pre-filters) is enough? I don't remember what we did for the other resolver though.

src/neo4j_graphrag/experimental/components/resolver.py
```diff
@@ -137,3 +143,182 @@ async def run(self) -> ResolutionStats:
...
+        Attempt to load the specified spaCy model by name.
+        If not installed, automatically download and then load it.
+        """
+        try:
+            return spacy.load(model_name)
+        except OSError as e:
+            # The exact error message can differ slightly depending on spaCy version,
+            # so you may want to be broader or narrower with handling logic:
+            if "doesn't seem to be a Python package or a valid path" in str(e):
+                print(f"Model '{model_name}' not found. Downloading...")
```
stellasia, line 320: Please remove print or replace it with a log (it's easier to redirect for users).

src/neo4j_graphrag/experimental/components/resolver.py
```diff
@@ -137,3 +143,182 @@ async def run(self) -> ResolutionStats:
...
+    @staticmethod
+    def _load_or_download_spacy_model(model_name: str) -> Language:
+        """
+        Attempt to load the specified spaCy model by name.
+        If not installed, automatically download and then load it.
+        """
+        try:
+            return spacy.load(model_name)
+        except OSError as e:
+            # The exact error message can differ slightly depending on spaCy version,
```
stellasia, line 317: I don't get the message here, does that mean that the fallback does not work in some cases?

src/neo4j_graphrag/experimental/components/resolver.py
```diff
@@ -137,3 +143,182 @@ async def run(self) -> ResolutionStats:
...
+                        "WITH collect(n) AS nodes "
+                        "CALL apoc.refactor.mergeNodes(nodes, {properties: 'discard', mergeRels: true}) "
+                        "YIELD node RETURN id(node)"
+                    )
+                    result, _, _ = self.driver.execute_query(
+                        merge_query,
+                        {"ids": list(node_id_set)},
+                        database_=self.neo4j_database,
+                    )
+                    merged_count += len(result)
```
stellasia, line 271: If I follow, `len(result)` is always 1 here, right? (not meaning we need to change it, just checking my understanding)

examples/customize/build_graph/components/resolvers/spacy_entity_resolver_pre_filter.py
NathalieCharbel: You’re right. I initially added both because that’s how it is for the `SinglePropertyExactMatchResolver`. In this case, maybe it's better to leave only one example for each resolver.

src/neo4j_graphrag/experimental/components/resolver.py
NathalieCharbel: If the user would like to use the spaCy resolver two things are required:
1- installing the spaCy library
2- downloading the language model. 
The first one is handled in poetry. This function is to handle downloading the model. Maybe we can also trigger it from poetry? it's just that I found it easier this way. any preference?

src/neo4j_graphrag/experimental/components/resolver.py
NathalieCharbel: yes, whenever there is a merge!

src/neo4j_graphrag/experimental/components/resolver.py
NathalieCharbel: I added a log message.

NathalieCharbel:
Closing this PR as it was handled by [PR#319](https://github.com/neo4j/neo4j-graphrag-python/pull/319)

PR #296, opened 2025-03-06
Add optional schema enforcement for KG builder as a validation layer after entity and relation extraction

NathalieCharbel:
# Description

This PR adds an optional schema enforcement layer to validate the extracted entities, relations, and properties against the schema that is passed to the LLM as guidance for the extraction. This change introduces:
- A new `_clean_graph` function in the `LLMEntityRelationExtractor` class that filters out invalid nodes, relationships, and properties based on a provided `SchemaConfig` object:
- Support for different enforcement modes (e.g., NONE vs. STRICT). So far implemented using Enum.
- Unit tests covering various enforcement scenarios (no schema, invalid nodes, invalid relations, etc.).

The cleanup is performed at the chunk graphs level after the extraction and before post-processing, i.e.,before relationships to the chunks are created.

Cleanup logic:
- Remove invalid nodes (not conformant to the schema)
- Remove invalid node properties (not conformant to the schema)
- Remove nodes that are left with no properties
- Remove invalid relations (whose types or start/end nodes are not conformant to the schema)
- Remove relations whose start/end node was removed
- Remove invalid relations properties (not conformant to the schema).

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [ ] Project configuration change

## Complexity

> **Note**
>
> Please provide an estimated complexity of this PR of either Low, Medium or High
>
>

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [x] Documentation has been updated
- [x] Unit tests have been updated
- [ ] E2E tests have been updated
- [ ] Examples have been updated
- [ ] New files have copyright header
- [ ] CLA (https://neo4j.com/developer/cla/) has been signed
- [x] CHANGELOG.md updated if appropriate

PR #315, opened 2025-03-21
Expose schema enforcement via SimpleKGPipeline

NathalieCharbel:
# Description

This PR fixes an issue from [PR#296]( https://github.com/neo4j/neo4j-graphrag-python/pull/296), where the `enforce_schema` parameter was mistakenly omitted from `SimpleKGPipeline`, preventing the optional schema enforcement feature from being enabled. This update adds the missing parameter, allowing users to configure schema enforcement as intended.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/pipeline/kg_builder.py
+++ b/src/neo4j_graphrag/experimental/pipeline/kg_builder.py
@@ -25,8 +25,10 @@ from neo4j_graphrag.experimental.components.entity_relation_extractor import OnE
 from neo4j_graphrag.experimental.components.kg_writer import KGWriter
 from neo4j_graphrag.experimental.components.pdf_loader import DataLoader
 from neo4j_graphrag.experimental.components.text_splitters.base import TextSplitter
-from neo4j_graphrag.experimental.components.types import LexicalGraphConfig, \
-    SchemaEnforcementMode
+from neo4j_graphrag.experimental.components.types import (
+    LexicalGraphConfig,
+    SchemaEnforcementMode,
+)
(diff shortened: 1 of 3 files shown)


src/neo4j_graphrag/experimental/pipeline/kg_builder.py
```diff
@@ -81,6 +85,7 @@ def __init__(
         entities: Optional[Sequence[EntityInputType]] = None,
         relations: Optional[Sequence[RelationInputType]] = None,
         potential_schema: Optional[List[tuple[str, str, str]]] = None,
+        enforce_schema: str = "NONE",
```
willtai, line 88: Should this type be `Optional[SchemaEnforcementMode]` instead and default to `None`?

NathalieCharbel: Two things here: 
- we wanted `SimpleKGPipeline` to accept a string and convert it internally to the enum value (for consistency with `on_error` param)
- as for `Optional`, I don't mind. I initially made it similar to what we have with `on_error`, but I also thought having it like this will avoid extra conditions for `None` values. It is also "semantically" different (`SchemaEnforcementMode.NONE` means that we know the enforcement option chosen by the user -> no enforcement , `None` means the choice of schema enforcement is unset or unknown). But maybe we don't care? wdyt?

willtai (approved):

PR #322, opened 2025-04-04
Add tool calling to the LLM base class, implement in OpenAI

oskarhane:
# Description

> **Note**
>
> This PR adds tool calling functionality to the LLM base class with OpenAI implementation, enabling structured parameter extraction and function calling. The implementation includes both synchronous and asynchronous methods for tool invocation, with proper type hints and error handling. An example has been added to demonstrate how to use this new functionality.
>

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
> The implementation adds a new feature to the existing LLM interface but follows established patterns. It required careful handling of the OpenAI API's tool invocation parameters and proper type definitions.
>

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

- [x] CLA ([https://neo4j.com/developer/cla/)](https://neo4j.com/developer/cla/)) has been signed

### DIFF
--- a/examples/tool_calls/openai_tool_calls.py
+++ b/examples/tool_calls/openai_tool_calls.py
@@ -45,16 +45,22 @@ person_info_tool = Tool(
 TOOLS = [person_info_tool]
 
 
-def process_tool_call(response: ToolCallResponse) -> Dict[str, Any]:
-    """Process the tool call response and return the extracted parameters."""
+def process_tool_calls(response: ToolCallResponse) -> Dict[str, Any]:
+    """Process all tool calls in the response and return the extracted parameters."""
     if not response.tool_calls:
         raise ValueError("No tool calls found in response")
 
-    tool_call = response.tool_calls[0]
-    print(f"\nTool called: {tool_call.name}")
-    print(f"Arguments: {tool_call.arguments}")
+    print(f"\nNumber of tool calls: {len(response.tool_calls)}")
     print(f"Additional content: {response.content or 'None'}")
-    return tool_call.arguments
+    
+    results = []
+    for i, tool_call in enumerate(response.tool_calls):
+        print(f"\nTool call #{i+1}: {tool_call.name}")
+        print(f"Arguments: {tool_call.arguments}")
+        results.append(tool_call.arguments)
+    
+    # For backward compatibility, return the first tool call's arguments
+    return results[0] if results else {}
 
 
 async def main() -> None:
@@ -74,7 +80,7 @@ async def main() -> None:
         input=f"Extract information about the person from this text: {text}",
         tools=TOOLS,
     )
-    sync_result = process_tool_call(sync_response)
+    sync_result = process_tool_calls(sync_response)
     print("\n=== Synchronous Tool Call Result ===")
     print(json.dumps(sync_result, indent=2))
 
@@ -85,7 +91,7 @@ async def main() -> None:
         input=f"Extract information about the person from this text: {text2}",
         tools=TOOLS,
     )
-    async_result = process_tool_call(async_response)
+    async_result = process_tool_calls(async_response)
     print("\n=== Asynchronous Tool Call Result ===")
     print(json.dumps(async_result, indent=2))
 
--- a/examples/tool_calls/openai_tool_calls.py
+++ b/examples/tool_calls/openai_tool_calls.py
@@ -17,32 +17,32 @@ from dotenv import load_dotenv
 
 from neo4j_graphrag.llm import OpenAILLM
 from neo4j_graphrag.llm.types import ToolCallResponse
+from neo4j_graphrag.tool import Tool, ObjectParameter, StringParameter, IntegerParameter
 
 # Load environment variables from .env file
 load_dotenv()
 
-# Define a tool for extracting information from text
-TOOLS = [
-    {
-        "type": "function",
-        "function": {
-            "name": "extract_person_info",
-            "description": "Extract information about a person from text",
-            "parameters": {
-                "type": "object",
-                "properties": {
-                    "name": {"type": "string", "description": "The person's full name"},
-                    "age": {"type": "integer", "description": "The person's age"},
-                    "occupation": {
-                        "type": "string",
-                        "description": "The person's occupation",
-                    },
-                },
-                "required": ["name"],
-            },
-        },
-    }
-]
+
+# Create a custom Tool implementation for person info extraction
+parameters = ObjectParameter(
+    description="Parameters for extracting person information",
+    properties={
+        "name": StringParameter(description="The person's full name"),
+        "age": IntegerParameter(description="The person's age"),
+        "occupation": StringParameter(description="The person's occupation"),
+    },
+    required_properties=["name"],
+    additional_properties=False,
+)
+person_info_tool = Tool(
+    name="extract_person_info",
+    description="Extract information about a person from text",
+    parameters=parameters,
+    execute_func=lambda **kwargs: kwargs,
+)
+
+# Create the tool instance
+TOOLS = [person_info_tool]
 
 
 def process_tool_call(response: ToolCallResponse) -> Dict[str, Any]:
--- a/examples/tool_calls/openai_tool_calls.py
+++ b/examples/tool_calls/openai_tool_calls.py
@@ -1,0 +1,95 @@
+"""
+Example showing how to use OpenAI tool calls with parameter extraction.
+Both synchronous and asynchronous examples are provided.
+
+To run this example:
+1. Make sure you have the OpenAI API key in your .env file:
+   OPENAI_API_KEY=your-api-key
+2. Run: python examples/tool_calls/openai_tool_calls.py
+"""
+
+import asyncio
+import json
+import os
+from typing import Dict, Any
+
+from dotenv import load_dotenv
+
+from neo4j_graphrag.llm import OpenAILLM
+from neo4j_graphrag.llm.types import ToolCallResponse
+
+# Load environment variables from .env file
+load_dotenv()
+
+# Define a tool for extracting information from text
+TOOLS = [
+    {
+        "type": "function",
+        "function": {
+            "name": "extract_person_info",
+            "description": "Extract information about a person from text",
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string", "description": "The person's full name"},
+                    "age": {"type": "integer", "description": "The person's age"},
+                    "occupation": {
+                        "type": "string",
+                        "description": "The person's occupation",
+                    },
+                },
+                "required": ["name"],
+            },
+        },
+    }
+]
+
+
+def process_tool_call(response: ToolCallResponse) -> Dict[str, Any]:
+    """Process the tool call response and return the extracted parameters."""
+    if not response.tool_calls:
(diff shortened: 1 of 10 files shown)


src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -72,20 +77,51 @@ def get_messages(
...
+                        )
+                    )
+                elif role == "assistant":
+                    messages.append(
+                        ChatCompletionAssistantMessageParam(
+                            role="assistant", content=str(content)
+                        )
+                    )
+        messages.append(ChatCompletionUserMessageParam(role="user", content=str(input)))
+        return messages
```
oskarhane, line 124: Not sure about this change, but mypy kept complaining locally so I had to do this. Happy to undo.

src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -121,6 +125,79 @@ def invoke(
...
+                message_history = message_history.messages
+
+            params = self.model_params.copy() if self.model_params else {}
+            if "temperature" not in params:
+                params["temperature"] = 0.0
+
+            # Convert tools to OpenAI's expected type
+            openai_tools: List[ChatCompletionToolParam] = []
+            for tool in tools:
+                openai_tools.append(cast(ChatCompletionToolParam, tool))
```
stellasia, line 162: Can't we create the object with `ChatCompletionToolParam(**tool)` (or something similar)?

src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -121,6 +125,79 @@ def invoke(
...
+                tool_choice="auto",
+                **params,
+            )
+
+            message = response.choices[0].message
+
+            # If there's no tool call, return the content as a regular response
+            if not message.tool_calls or len(message.tool_calls) == 0:
+                return ToolCallResponse(
+                    tool_calls=[ToolCall(name="", arguments={})],
```
stellasia, line 177: I'd expect an empty list here if there was no tool matching the query, wdty?

src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -121,6 +125,79 @@ def invoke(
...
+                    content=message.content or "",
```
stellasia, line 178: Since `content` is optional in this model, can we get rid of the ` or ""`?

src/neo4j_graphrag/llm/openai_llm.py
```diff
@@ -155,6 +232,80 @@ async def ainvoke(
...
+
+            response = await self.async_client.chat.completions.create(
+                messages=self.get_messages(input, message_history, system_instruction),
+                model=self.model_name,
+                tools=openai_tools,
+                tool_choice="auto",
+                **params,
+            )
+
+            message = response.choices[0].message
```
stellasia, line 279: Not for now because this is also the case in the already existing methods, but we should probably try to reduce code duplication between the sync and async calls.

tests/unit/llm/test_openai_llm.py
```diff
@@ -13,16 +13,18 @@
...
 from neo4j_graphrag.exceptions import LLMGenerationError
 from neo4j_graphrag.llm import LLMResponse
 from neo4j_graphrag.llm.openai_llm import AzureOpenAILLM, OpenAILLM
+from neo4j_graphrag.llm.types import ToolCallResponse
 
 
-def get_mock_openai() -> MagicMock:
-    mock = MagicMock()
+def get_mock_openai() -> Any:
+    mock = cast(Any, MagicMock())
```
stellasia, line 27: That's so... mypy-ugly :) I hope there is another way to make mypy happy but I need to think more about it :)

src/neo4j_graphrag/llm/openai_llm.py
stellasia: Yes I guess this is because `openai_tools` is declared to be `List[ChatCompletionToolParam]` but `tool` is a dict. That's why I thought instantiating the `ChatCompletionToolParam` could fix mypy without having to use `cast`, but I don't know the model exactly so maybe it doesn't work.

examples/tool_calls/openai_tool_calls.py
```diff
@@ -0,0 +1,95 @@
...
+import os
+from typing import Dict, Any
+
+from dotenv import load_dotenv
+
+from neo4j_graphrag.llm import OpenAILLM
+from neo4j_graphrag.llm.types import ToolCallResponse
+
+# Load environment variables from .env file
+load_dotenv()
```
stellasia, line 22: Maybe specify that this is where to set the `OPENAI_API_KEY` ?

examples/tool_calls/openai_tool_calls.py
```diff
@@ -0,0 +1,95 @@
...
+    }
+]
+
+
+def process_tool_call(response: ToolCallResponse) -> Dict[str, Any]:
+    """Process the tool call response and return the extracted parameters."""
+    if not response.tool_calls:
+        raise ValueError("No tool calls found in response")
+
+    tool_call = response.tool_calls[0]
```
stellasia, line 53: Is there a reason to print only the first tool?

tests/unit/llm/test_openai_llm.py
```diff
@@ -65,10 +67,14 @@ def test_openai_llm_with_message_history_happy_path(mock_import: Mock) -> None:
     assert isinstance(res, LLMResponse)
     assert res.content == "openai chat response"
     message_history.append({"role": "user", "content": question})
-    llm.client.chat.completions.create.assert_called_once_with(  # type: ignore
-        messages=message_history,
-        model="gpt",
-    )
+    # Use assert_called_once() instead of assert_called_once_with() to avoid issues with overloaded functions
+    cast(Any, llm.client.chat.completions.create).assert_called_once()
```
stellasia, line 71: I thought these `cast(Any...)` were fixed?

src/neo4j_graphrag/tool.py
```diff
@@ -0,0 +1,224 @@
...
+        return result
+
+
+class ArrayParameter(ToolParameter):
+    """Array parameter for tools."""
+
+    def __init__(
+        self,
+        description: str,
+        items: ToolParameter,
```
stellasia, line 149: Shouldn't `items` be a `list[ToolParameter]`?

src/neo4j_graphrag/tool.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class BooleanParameter(ToolParameter):
+    """Boolean parameter for tools."""
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {"type": ParameterType.BOOLEAN, "description": self.description}
+
+
+class ObjectParameter(ToolParameter):
```
stellasia, line 107: Nice! It's maybe a bit verbose, would be great if users could provide tool definition as dict, wdyt? I think converting the `*Parameter` to Pydantic model would make this easily possible (and we could also get rid of the `to_dict` methods).

examples/tool_calls/openai_tool_calls.py
```diff
@@ -0,0 +1,95 @@
+"""
```
stellasia, line 1: Another small thing: maybe this file (or folder) can be moved under `customize/llms` and remember to add a link in the examples' README please.

src/neo4j_graphrag/tool.py
oskarhane: Actually not, the JSON schema definition of the array item types is a type. See https://json-schema.org/understanding-json-schema/reference/array#items

src/neo4j_graphrag/tool.py
oskarhane: Great idea!

tests/unit/llm/test_openai_llm.py
oskarhane: Yeah, me too. Removed and added mypy ignores in test file

src/neo4j_graphrag/tool.py
oskarhane: Pushed this change now

src/neo4j_graphrag/tool.py
stellasia: Ah ok, makes sense actually.

src/neo4j_graphrag/tool.py
```diff
@@ -0,0 +1,261 @@
...
+class ToolParameter(BaseModel):
+    """Base class for all tool parameters using Pydantic."""
+
+    description: str
+    required: bool = False
+    type: ClassVar[ParameterType]
+
+    def model_dump_tool(self) -> Dict[str, Any]:
+        """Convert the parameter to a dictionary format for tool usage."""
+        result = {"type": self.type, "description": self.description}
```
stellasia, line 27: We do not need `required` here?

src/neo4j_graphrag/tool.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class IntegerParameter(ToolParameter):
+    """Integer parameter for tools."""
+
+    type: ClassVar[ParameterType] = ParameterType.INTEGER
+    minimum: Optional[int] = None
+    maximum: Optional[int] = None
+
+    def model_dump_tool(self) -> Dict[str, Any]:
```
stellasia, line 75: Is `model_dump_tool` only used to discard null values? If so, there is an option in `model_dump` for this (`exclude_none=True` by memory)

stellasia: I know I'm a bit nitpicking here, but I've started to work on the implementation for Vertex AI, and some parameters need to be excluded, so if we can rely on Pydantic for this instead of manually re-implementing it, that would be great :) but if it's not possible, it's not possible.

src/neo4j_graphrag/tool.py
```diff
@@ -0,0 +1,261 @@
...
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "ToolParameter":
+        """Create a parameter from a dictionary."""
+        param_type = data.get("type")
+        if not param_type:
+            raise ValueError("Parameter type is required")
+
+        # Find the appropriate class based on the type
+        param_classes = {
```
stellasia, line 38: I feel like we could get rid of many of these by using either a field discriminator or some union, but I won't be strict on this point, experiments can be left for a cool down period.

src/neo4j_graphrag/tool.py
oskarhane: It does a bit more than that mapping some names "min_items" -> "minItems", but that could be removed by naming the parameters "minItems" directly. 
Your call, do you want me to change it?

src/neo4j_graphrag/tool.py
oskarhane: Sounds good, let's look into it later

src/neo4j_graphrag/tool.py
stellasia: Ok I see, no do not change anything now, we'll add improvements as we need them. They are unit tested so that's great :)

stellasia (approved):
🦸‍♂️

PR #332, opened 2025-05-08
Add ToolsRetriever class and Retriever.convert_to_tool() method

oskarhane:
# Description

In addition to the major parts of this PR, I've refactored and moved the `tool.py` file into its own directory to group relevant tool files together. This move touches a lot of files because of updated imports.

### `ToolsRetriever`
Now that we have tool support in the LLM interface, let's create a retriever that can use one or more tools to retrieve data. 

This retriever follows the Retriever interface so it can be used within the GraphRAG class to get the full e2e experience (see example file).

The ToolsRetriever uses an LLM to decide on what tools to use to find the relevant data.

```python
calendar_tool = CalendarTool()
weather_tool = WeatherTool()

llm = OpenAILLM(...)
driver = GraphDatabase.driver(...)

tools_retriever = ToolsRetriever(
    driver=driver,
    llm=llm,
    tools=[calendar_tool, weather_tool],
)

graphrag = GraphRAG(
    llm=llm,
    retriever=tools_retriever,
)

result = graphrag.search(query_text="Tell me about tomorrow", return_context=False)
```

### `convert_to_tool ()`
This method is a way to convert a `Retriever` to a `Tool` so it can be used within the `ToolsRetriever`. This is useful when you might want to have both a `VectorRetriever` and a `Text2CypherRetreiver` as a fallback.  
See new example files for usage.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


examples/retrieve/tools/retriever_to_tool_example.py
```diff
@@ -0,0 +1,121 @@
...
+        return RawSearchResult(records=records, metadata={"query": query_text})
+
+
+def main() -> None:
+    # Convert a StaticRetriever to a tool with specific parameters
+    static_retriever = StaticRetriever(driver=cast(Any, MagicMock()))
+
+    # Define parameters for the static retriever tool
+    static_parameters = ObjectParameter(
+        description="Parameters for the Neo4j information retriever",
```
stellasia, line 81: Is this description really required? It doesn't seem to have a real added-value.

examples/retrieve/tools/retriever_to_tool_example.py
```diff
@@ -0,0 +1,121 @@
...
+        properties={
+            "query_text": StringParameter(
+                description="The query about Neo4j (any query will return general Neo4j information)",
+                required=True,
+            ),
+        },
+    )
+
+    # Convert the retriever to a tool with specific parameters
+    static_tool = convert_retriever_to_tool(
```
stellasia, line 91: I'd say this should be a method from the `Retriever` class, and `parameters` should be encapsulated in this class as well, these parameters are bound to the `search` method and won't change from one instance to another. 

So something like:

```
class Retriever:
   def get_parameters(self) -> ObjectParameter:
       raise NotImplementedError()

   def convert_to_tool(self, name: str, description: Optional[str] = None) -> Tool:
       raise NotImplementedError()
```

src/neo4j_graphrag/retrievers/tools_retriever.py
```diff
@@ -0,0 +1,161 @@
...
+        neo4j_database (Optional[str], optional): Neo4j database name. Defaults to None.
+        system_instruction (Optional[str], optional): Custom system instruction for the LLM. Defaults to None.
+    """
+
+    # Disable Neo4j version verification since this retriever doesn't directly interact with Neo4j
+    VERIFY_NEO4J_VERSION = False
+
+    def __init__(
+        self,
+        driver: neo4j.Driver,
```
stellasia, line 48: Should we consider updating the `Retriever` interface if this once does not use the driver?

src/neo4j_graphrag/retrievers/tools_retriever.py
```diff
@@ -0,0 +1,161 @@
...
+
+                # Find the tool by name
+                selected_tool = next(
+                    (tool for tool in self._tools if tool.get_name() == tool_name), None
+                )
+                if selected_tool is not None:
+                    # Extract arguments from the tool call
+                    tool_args = tool_call.arguments or {}
+
+                    # Always include the query_text in the arguments for tools that might need it
```
stellasia, line 130: Shouldn't the LLM include the query if needed by the tool? It should be in the tool parameters in this case?

src/neo4j_graphrag/retrievers/tools_retriever.py
```diff
@@ -0,0 +1,161 @@
...
+                    tool_args.setdefault("query", query_text)
+
+                    # Execute the tool with the provided arguments
+                    tool_result = selected_tool.execute(**tool_args)
+                    # If the tool result is a RawSearchResult, extract its records
+                    if hasattr(tool_result, "records"):
```
stellasia, line 136: We can maybe enforce the fact that `Tool.execute` must return a list of something (or event string?), or do we really want to create fake records?

src/neo4j_graphrag/tools/tool.py
```diff
@@ -211,23 +211,28 @@ def validate_properties(self) -> "ObjectParameter":
 class Tool(ABC):
     """Abstract base class defining the interface for all tools in the neo4j-graphrag library."""
 
+    _name: str
```
stellasia, line 214: Do you remember why we need to have these declarations here? Mypy stuff?

src/neo4j_graphrag/tools/utils.py
```diff
@@ -0,0 +1,76 @@
...
+            If not provided, no parameters will be included in the tool.
+        name (Optional[str]): Custom name for the tool. If not provided,
+            an attempt will be made to infer it from the retriever or a default name will be used.
+
+    Returns:
+        RetrieverTool: A Tool object configured to use the retriever's search functionality.
+    """
+    # Use provided name or infer it from the retriever
+    if name is None:
+        name = getattr(retriever, "name", None) or getattr(
```
stellasia: If we expect retrievers to have a name attribute, maybe we should add this to the interface?

src/neo4j_graphrag/tools/utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Parameters can be None
+
+    # Define a function that matches the Callable[[str, ...], Any] signature
+    def execute_func(**kwargs: Any) -> Any:
+        # The retriever's get_search_results method is expected to handle
+        # arguments like query_text, top_k, etc., passed as keyword arguments.
+        # The Tool's 'parameters' definition (e.g., ObjectParameter) ensures
+        # that these arguments are provided in kwargs when Tool.execute is called.
+        return retriever.get_search_results(**kwargs)
```
stellasia, line 62: Why not using the `search` method?

oskarhane: I think dealing with the raw output is better in these cases, and let the ToolRetriever handle the formatting.

src/neo4j_graphrag/tools/tool.py
oskarhane: No idea why, but yes, mypy :)

src/neo4j_graphrag/retrievers/tools_retriever.py
oskarhane: I'm not sure tbh. I think the tools retriever is an exception so I think we can keep it for now.

examples/retrieve/tools/retriever_to_tool_example.py
oskarhane: Yeah, makes sense. I wanted to be as little intrusive as possible, bot I've updated to a method like this now.

src/neo4j_graphrag/retrievers/tools_retriever.py
oskarhane: Yeah, I'll think about this one a bit.

src/neo4j_graphrag/tools/utils.py
stellasia: Each retriever is likely to return results in different format, I mean it will "often" be a neo4j Record, but with different keys, especially if using the "*CypherRetriever", so the only formatting we can do here is stringify these records, right? But we can let the formatting question for a later discussion, it's probably part of a wider decision.

src/neo4j_graphrag/retrievers/base.py
```diff
@@ -175,6 +193,253 @@ def default_record_formatter(self, record: neo4j.Record) -> RetrieverResultItem:
...
+        Args:
+            parameter_descriptions (Optional[Dict[str, str]]): Custom descriptions for parameters.
+                Keys should match parameter names from get_search_results method.
+
+        Returns:
+            ObjectParameter: The parameter definition for this retriever
+        """
+        return self._infer_parameters_from_signature(parameter_descriptions or {})
+
+    def _infer_parameters_from_signature(
```
stellasia, line 212: Nice :)

src/neo4j_graphrag/retrievers/base.py
```diff
@@ -175,6 +193,253 @@ def default_record_formatter(self, record: neo4j.Record) -> RetrieverResultItem:
...
+                ),
+                required=is_required,
+            )
+
+        # Get the origin and args for generic types
+        origin = get_origin(type_annotation)
+        args = get_args(type_annotation)
+
+        # Handle Optional[T] and Union[T, None]
+        if origin is Union:
```
stellasia, line 295: `Optional[T]` converted to `Union[None, T]` internally?

src/neo4j_graphrag/tools/tool.py
```diff
@@ -211,23 +211,28 @@ def validate_properties(self) -> "ObjectParameter":
...
 
-        # Allow parameters to be provided as a dictionary
         if isinstance(parameters, dict):
             self._parameters = ObjectParameter.model_validate(parameters)
-        else:
+        elif isinstance(parameters, ObjectParameter):
             self._parameters = parameters
-
-        self._execute_func = execute_func
+        else:
```
stellasia: Would it be worth checking parameters is really None here? The user could provide a different type (a list of something for instance) and silently not using it would lead to unexpected behavior

src/neo4j_graphrag/retrievers/tools_retriever.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+
+    # Disable Neo4j version verification since this retriever doesn't directly interact with Neo4j
+    VERIFY_NEO4J_VERSION = False
+
+    def __init__(
+        self,
+        driver: neo4j.Driver,
+        llm: LLMInterface,
+        tools: Sequence[Tool],
```
stellasia, line 50: Are we checking somewhere that tool's name are unique? Or this is not a requirement for the LLM?

src/neo4j_graphrag/tools/utils.py
```diff
@@ -0,0 +1,76 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import Any, Dict, Optional, Union
+
+from neo4j_graphrag.tools.tool import Tool, ObjectParameter
+
+
+def convert_retriever_to_tool(
```
stellasia: Isn't this function replaced by `Retriever.convert_to_tool`?

src/neo4j_graphrag/retrievers/base.py
stellasia: Yes, just saw it in the test

src/neo4j_graphrag/tools/utils.py
oskarhane: Yeah, gone in last commit 👍

src/neo4j_graphrag/retrievers/tools_retriever.py
oskarhane: Good point, updated

src/neo4j_graphrag/tools/tool.py
oskarhane: Yeah, makes it more robust. Updated.

src/neo4j_graphrag/tools/utils.py
oskarhane: Hmm, yeah, I guess that's correct. I updated the we call the `.search()` method now instead and I do some result type checking. See if that look ok.
I think it 's better than it was at least :)

PR #342, opened 2025-05-20
Fix VertexAILLM

stellasia:
# Description

The main reason for this PR is that, even if Google `GenerativeModel` accepts a list of tools as parameter, this is not how we should provide multiple tools. We should instead [provide a single tool with multiple function declarations](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#python-dictionary). This PR addresses this behavior.

It also addresses a few other problems related to the way `GenerativeModel` is instantiated. The usage is:

```python
    # config for chat mode (aka invoke)
    generation_config = GenerationConfig(
        temperature=0.0,
        response_mime_type="application/json",
    )
    # config for tool mode (aka invoke_with_tools)
    # optional
    tool_config = ToolConfig(
        function_calling_config=ToolConfig.FunctionCallingConfig(
            mode=ToolConfig.FunctionCallingConfig.Mode.ANY
        ),
    )

    llm = VertexAILLM(
        model_name="gemini-2.0-flash-001",
        generation_config=generation_config,
        tool_config=tool_config,
    )
```

Then, internally:
- if `invoke` mode: we drop the tool config (it's not useful)
- if `invoke_with_tool` mode, we drop the generation config (otherwise the presence of fields like `response_mime_type`/`response_schema` raise errors)

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: ?

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/examples/customize/llms/vertexai_tool_calls.py
+++ b/examples/customize/llms/vertexai_tool_calls.py
@@ -97,8 +97,13 @@ async def main() -> None:
     # Initialize the VertexAI LLM
     generation_config = GenerationConfig(temperature=0.0)
     llm = VertexAILLM(
-        model_name="gemini-1.5-flash-001",
+        model_name="gemini-2.0-flash-001",
         generation_config=generation_config,
+        # tool_config=ToolConfig(
+        #     function_calling_config=ToolConfig.FunctionCallingConfig(
+        #         mode=ToolConfig.FunctionCallingConfig.Mode.ANY,
+        #         # allowed_function_names=["extract_person_info"],
+        #     ))
     )
 
     # Example text containing information about a company
--- a/examples/customize/llms/vertexai_tool_calls.py
+++ b/examples/customize/llms/vertexai_tool_calls.py
@@ -88,7 +88,7 @@ def process_tool_call(response: ToolCallResponse) -> str:
     if tool_call.name == "extract_person_info":
         return person_info_tool.execute(**tool_call.arguments)  # type: ignore[no-any-return]
     elif tool_call.name == "extract_company_info":
-        return company_info_tool.execute(**tool_call.arguments)
+        return str(company_info_tool.execute(**tool_call.arguments))
     else:
         raise ValueError("Unknown tool call")
 
--- a/examples/customize/llms/vertexai_tool_calls.py
+++ b/examples/customize/llms/vertexai_tool_calls.py
@@ -101,8 +101,7 @@ async def main() -> None:
         generation_config=generation_config,
     )
 
-    # Example text containing information about a person
-    # text = "Stella Hane is a 35-year-old software engineer who loves coding."
+    # Example text containing information about a company
     text1 = "Neo4j is a software company created in 2017"
 
     print("\n=== Synchronous Tool Call ===")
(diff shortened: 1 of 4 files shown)


src/neo4j_graphrag/llm/vertexai_llm.py
```diff
@@ -172,39 +170,45 @@ async def ainvoke(
...
-        return [self._to_vertexai_tool(tool) for tool in tools]
+        return [
+            VertexAITool(
+                function_declarations=[
+                    self._to_vertexai_function_declaration(tool) for tool in tools
+                ]
+            )
+        ]
+
+    def _get_options(self, tool_mode: bool = False) -> dict[str, Any]:
```
alexthomas93, line 203: Does this need to be its own method? Why not just put this directly in `_get_model`?

src/neo4j_graphrag/llm/vertexai_llm.py
```diff
@@ -137,20 +137,18 @@ def invoke(
...
+            system_instruction=system_instruction,
+            tools=None,
         )
         try:
             if isinstance(message_history, MessageHistory):
                 message_history = message_history.messages
             messages = self.get_messages(input, message_history)
-            response = self.model.generate_content(messages, **self.model_params)
-            return LLMResponse(content=response.text)
+            response = model.generate_content(messages)
```
alexthomas93, line 148: Why are we removing `model_params` here?

examples/customize/llms/vertexai_tool_calls.py
```diff
@@ -65,21 +101,21 @@ async def main() -> None:
         generation_config=generation_config,
     )
 
-    # Example text containing information about a person
-    text = "Stella Hane is a 35-year-old software engineer who loves coding."
+    # Example text containing information about a company
+    text1 = "Neo4j is a software company created in 2017"
```
alexthomas93, line 105: Nitpicky but I think it was created in 2007 😛

src/neo4j_graphrag/llm/vertexai_llm.py
stellasia: It used to be more complex in previous versions, that's why it was in a separate method, but moving it back to _get_model now.

examples/customize/llms/vertexai_tool_calls.py
stellasia: ahah, thank you :)

src/neo4j_graphrag/llm/vertexai_llm.py
stellasia: So, finally took time to think about this question, that was definitely a good one :) I think I messed up things again, due to the fact that "system_instruction" is only a "model" parameter (not an "invoke" parameter for GenerativeModel), and I thought it was the same for tools, which isn't: tools can be provided either in the init OR in the model call. I've made some related changes here: https://github.com/neo4j/neo4j-graphrag-python/pull/342/commits/c1db7569dd1a392fc4a0f0eba359b7611a640905

alexthomas93 (approved):
LGTM

PR #345, opened 2025-05-21
Fix schema validation

stellasia:
# Description

- One test was wrong: we still want users to be able to provide entity types as strings (ie only labels)
- While investigating, realized part of the code was not used, so removed a full `else` block
- Also, it seems that having `relationship_types=None` in the prompt is interpreted as "do not extract relationships" by the LLM, so `relationship_types` is now removed if null (but not if empty)

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/examples/build_graph/from_config_files/simple_kg_pipeline_config.json
+++ b/examples/build_graph/from_config_files/simple_kg_pipeline_config.json
@@ -42,65 +42,67 @@
         }
     },
     "from_pdf": false,
-    "entities": [
-        "Person",
-        {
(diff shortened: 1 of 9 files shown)


oskarhane (approved):
Nice catch

PR #347, opened 2025-05-23
New schema and pruning

stellasia:
# Description

Still to be tested manually on a bunch of examples, and update doc

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


docs/source/user_guide_kg_builder.rst
```diff
@@ -74,9 +74,8 @@ Graph Schema
 ------------
 
 It is possible to guide the LLM by supplying a list of node and relationship types,
```
NathalieCharbel: for clarity, maybe it is also good to mention "their property types" as well.

docs/source/user_guide_kg_builder.rst
```diff
@@ -1017,6 +984,46 @@ If more customization is needed, it is possible to subclass the `EntityRelationE
...
+
+By default, all extracted elements — including nodes, relationships, and properties — are retained in the constructed graph. This behavior can be configured using the following schema options:
+(see :ref:`graphschema`)
+
+
+Configuration Options
+---------------------
+
+- **Required Properties**
+  Required properties may be specified at the node or relationship type level. Any extracted node or relationship missing one or more of its required properties will be pruned from the graph.
```
NathalieCharbel, line 1001: is the last statement above true? let's say one node type was defined with 2 required properties and the LLM has extracted one of them only, will the node be pruned?

NathalieCharbel: I understand the logic but I feel it's a bit counter intuitive wrt the general logic followed elsewhere. For instance, node types that are included in the schema are all implicitly required. However, if the LLM does not extract all of them, we do not prune the whole graph.

docs/source/user_guide_kg_builder.rst
```diff
@@ -1017,6 +984,46 @@ If more customization is needed, it is possible to subclass the `EntityRelationE
...
+
+- **Additional Properties** *(default: False)*
+  This node- or relationship-level option determines whether extra properties not listed in the schema should be retained.
+
+   - If set to ``False`` (default), all extracted properties are retained.
+   - If set to ``True``, only the properties defined in the schema are preserved; all others are removed.
```
NathalieCharbel: it's the other way round isn't it?

docs/source/user_guide_kg_builder.rst
```diff
@@ -1017,6 +984,46 @@ If more customization is needed, it is possible to subclass the `EntityRelationE
...
+(see :ref:`graphschema`)
+
+
+Configuration Options
+---------------------
+
+- **Required Properties**
+  Required properties may be specified at the node or relationship type level. Any extracted node or relationship missing one or more of its required properties will be pruned from the graph.
+
+- **Additional Properties** *(default: False)*
```
NathalieCharbel: shouldn't default value be true?

src/neo4j_graphrag/experimental/components/graph_pruning.py
```diff
@@ -0,0 +1,259 @@
...
+            new_graph = self._clean_graph(graph, schema)
+        else:
+            new_graph = graph
+        return GraphPruningResult(
+            graph=new_graph,
+            metadata={
+                "stats": {
+                    "pruned_node_count": len(graph.nodes) - len(new_graph.nodes),
+                    "pruned_relationship_count": len(graph.relationships)
+                    - len(new_graph.relationships),
```
NathalieCharbel: no stats on pruned properties?

src/neo4j_graphrag/experimental/components/graph_pruning.py
```diff
@@ -0,0 +1,259 @@
...
+            if new_node:
+                valid_nodes.append(new_node)
+        return valid_nodes
+
+    def _validate_relationship(
+        self,
+        rel: Neo4jRelationship,
+        valid_nodes: dict[str, str],
+        relationship_type: Optional[RelationshipType],
+        additional_relationship_types: bool,
```
NathalieCharbel, line 248: should we set default value to true?

docs/source/user_guide_kg_builder.rst
```diff
@@ -1017,6 +984,46 @@ If more customization is needed, it is possible to subclass the `EntityRelationE
...
+   - If set to ``True`` (default), such relationships are retained.
+   - If set to ``False``, relationships with undefined types are removed.
+
+- **Additional Patterns** *(default: True)*
+  This schema-level option determines whether relationship patterns not explicitly listed in the schema are allowed.
+
+   - If set to ``True`` (default), all patterns are retained.
+   - If set to ``False``, only patterns defined in the schema are kept. **Note** `additional_relationship_types` must also be `False`.
+
+
```
NathalieCharbel, line 1033: I think documentation explains well the configuration options related to whether we keep or remove additional LLM-extracted info. However, no explanations on how invalid nodes, rels, props are pruned, i.e., cases where schema is enforced implicitly. like rules pruning invalid rels, nodes, etc.

src/neo4j_graphrag/experimental/components/graph_pruning.py
```diff
@@ -0,0 +1,259 @@
...
+        filtered_nodes = self._enforce_nodes(graph.nodes, schema)
+        if not filtered_nodes:
+            logger.warning(
+                "PRUNING: all nodes were pruned, resulting graph will be empty. Check logs for details."
+            )
+            return Neo4jGraph()
+
+        # enforce relationships (remove those referencing invalid nodes or with invalid
+        # types or with start/end nodes not conforming to the schema, and strip invalid
+        # properties)
```
NathalieCharbel: For clarity, it is good to mention above that rel directions are corrected.

src/neo4j_graphrag/experimental/components/graph_pruning.py
```diff
@@ -0,0 +1,259 @@
...
+            )
+            if new_rel:
+                valid_rels.append(new_rel)
+        return valid_rels
+
+    def _enforce_properties(
+        self,
+        properties: dict[str, Any],
+        valid_properties: list[PropertyType],
+        additional_properties: bool,
```
NathalieCharbel, line 390: should we set default value to true?

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -119,38 +157,46 @@ class GraphSchema(DataModel):
...
                     )
 
         return self
 
+    @model_validator(mode="after")
+    def validate_additional_parameters(self) -> Self:
+        if self.additional_patterns and not self.additional_relationship_types:
+            raise ValueError(
+                "`additional_relationship_types` must be set to True when using `additional_patterns=True`"
+            )
```
NathalieCharbel, line 200: shouldn't the restriction be applied only when `additional_patterns` is set to `False` and `additional_relationship_types` is set to `True`? (doc above also mentions that).
what if the user wants to have additional patterns with schema-defined relationships only? I think in this case, `additional_patterns=True` and `additional_relationship_types=False` makes sense.

NathalieCharbel (commented):
Overall, very nice schema and pruning improvement, much better now! 
I have a few comments though.
PS: my review does not include tests yet.

docs/source/user_guide_kg_builder.rst
stellasia: To me, node types are not "required" in the sense that they might not be present in all the documents the user is importing. It is the same for properties which are not required: maybe some piece of information are in some part of the dataset and some other information in another document. 
However, it is possible to have a certain node type that doesn't make sense without a certain property. This "required" property  will be important for ER. 

For instance, if we find in document A Person "Alice" with age 20, and in document B Person "Alice" with occupation "nurse", maybe ER will be able to say they are the same person. But if one of them is missing "name" property, it will be merely impossible and then this node might be useless (or not, that's why it's configurable by the user)

src/neo4j_graphrag/experimental/components/graph_pruning.py
stellasia: Which kind of stats would you like to see? I'm not sure the number of pruned properties is useful, is it?

src/neo4j_graphrag/experimental/components/graph_pruning.py
stellasia: I'm not sure I see why we should as it's a private method?

src/neo4j_graphrag/experimental/components/schema.py
stellasia: Yes you're right, I messed up the condition here.

docs/source/user_guide_kg_builder.rst
stellasia: I added a note about empty nodes being pruned, is there anything else that needs to be mentionned? I don't get what you mean by "rules pruning" and I thought invalid rels was covered in the additional relationship/pattern sections?

src/neo4j_graphrag/experimental/components/graph_pruning.py
stellasia: Removed comment  entirely, otherwise it's a duplicate of the method's docstring

src/neo4j_graphrag/experimental/components/graph_pruning.py
stellasia: Same as above

docs/source/user_guide_kg_builder.rst
stellasia: Right, that's what happen when you ask ChatGPT to rephrase doc and do not check carefully what it did... :/ Fixed now, thanks.

docs/source/user_guide_kg_builder.rst
NathalieCharbel: I see, you convinced me. thanks for the clarification.

src/neo4j_graphrag/experimental/components/graph_pruning.py
NathalieCharbel: I was looking for the same kind of stats as we have for nodes and relationships, specifically, the number of pruned properties. Why do you think that wouldn’t be useful? The user might not want to allow extra properties initially, but they could still be interested in seeing what was suggested by the LLM and ultimately excluded. That information could be helpful for refining the schema and generating an improved graph in the next iteration.

src/neo4j_graphrag/experimental/components/graph_pruning.py
NathalieCharbel: Oh I see. So, we provide default values for these kind of flags on the creation of an instance of `GraphSchema`, right? If so, following this kind of logic should we remove the default value from `additional_node_types` in the `_validate_node` above?

docs/source/user_guide_kg_builder.rst
NathalieCharbel: I mean the rules that prune invalid rels, nodes, and props that are not conformant to the schema in general, independently of these flags. let's say a rel type that was defined in the schema but having wrong start node and end node. I think those rules are mentioned in docstring of graph pruning functions but not in documentation.

src/neo4j_graphrag/experimental/components/graph_pruning.py
NathalieCharbel: same as my reply [here](https://github.com/neo4j/neo4j-graphrag-python/pull/347#discussion_r2137419388)

docs/source/user_guide_kg_builder.rst
NathalieCharbel: so in short I mean the "Enforcement rules". this section was completely omitted, which I think is a bad idea.

tests/unit/experimental/components/test_graph_pruning.py
```diff
@@ -0,0 +1,392 @@
...
+        ),
+    ],
+)
+def test_graph_pruning_enforce_properties(
+    properties: dict[str, Any],
+    valid_properties: list[PropertyType],
+    additional_properties: bool,
+    expected_filtered_properties: dict[str, Any],
+) -> None:
+    prunner = GraphPruning()
```
NathalieCharbel: miner typo, `pruner`

tests/unit/experimental/components/test_graph_pruning.py
```diff
@@ -0,0 +1,392 @@
...
+def test_graph_pruning_validate_node(
+    node: Neo4jNode,
+    entity: str,
+    additional_node_types: bool,
+    expected_node: Neo4jNode,
+    request: pytest.FixtureRequest,
+) -> None:
+    e = request.getfixturevalue(entity) if entity else None
+
+    prunner = GraphPruning()
```
NathalieCharbel: same as above

src/neo4j_graphrag/experimental/components/graph_pruning.py
stellasia: Added more info about pruning and explanations in https://github.com/neo4j/neo4j-graphrag-python/pull/347/commits/d8f194992617fc358294de259d712b0702eec1b9

src/neo4j_graphrag/experimental/components/graph_pruning.py
stellasia: Deal, removed the default value for the _validate_node method ;)

src/neo4j_graphrag/experimental/components/graph_pruning.py
NathalieCharbel: Nice! :)

docs/source/user_guide_kg_builder.rst
stellasia: Added a section here: https://github.com/neo4j/neo4j-graphrag-python/pull/347/commits/8c5b8a29f2f05e8dc6f6fa741586ed1e34730469. LMK if you think of anything else that's still missing

PR #352, opened 2025-06-04
Pipeline state dump and load

NathalieCharbel:
# Description

This PR introduces state management capabilities to the `Pipeline` class, enabling:
- Checkpointing pipeline execution at specific components using `run_until`
- Resuming pipeline execution from saved states using `resume_from`
- Saving/loading pipeline state to/from a json file or or passing it directly between pipeline runs.

The state includes pipeline configuration, execution results, and final results from previous runs. This feature is particularly useful for:
- Debugging long-running pipelines
- Recovering from failures
- Comparing component implementations with deterministic inputs

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
low
>
>

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/docs/source/user_guide_pipeline.rst
+++ b/docs/source/user_guide_pipeline.rst
@@ -133,6 +133,36 @@ can be added to the visualization by setting `hide_unused_outputs` to `False`:
     webbrowser.open("pipeline_full.html")
 
 
+*************************
+Pipeline State Management
+*************************
+
+Pipelines support checkpointing and resumption through state management features:
+
+.. code:: python
+
+    # Run pipeline until a specific component
+    state = await pipeline.run_until(data, stop_after="component_name", state_file="state.json")
+
+    # Resume pipeline from a specific component
+    result = await pipeline.resume_from(state, data, start_from="component_name")
+
+    # Alternatively, load state from file
+    result = await pipeline.resume_from(None, data, start_from="component_name", state_file="state.json")
+
+The state contains:
+- Pipeline configuration (parameter mappings between components and validation state)
+- Execution results (outputs from completed components stored in the ResultStore)
+- Final pipeline results from previous runs
+- Component-specific state (interface available but not yet implemented by components)
+
+This enables:
+- Checkpointing long-running pipelines
+- Debugging pipeline execution
+- Resuming failed pipelines from the last successful component
+- Comparing different component implementations with deterministic inputs by saving the state before the component and reusing it, avoiding non-deterministic results from preceding components
+
+
 ************************
 Adding an Event Callback
 ************************
--- a/src/neo4j_graphrag/experimental/pipeline/pipeline.py
+++ b/src/neo4j_graphrag/experimental/pipeline/pipeline.py
@@ -18,8 +18,9 @@ import logging
 import warnings
 from collections import defaultdict
 from timeit import default_timer
-from typing import Any, Optional, AsyncGenerator, Dict
+from typing import Any, Optional, AsyncGenerator, Dict, Union
 import asyncio
+import json
 
 from neo4j_graphrag.utils.logging import prettify
 
@@ -580,14 +581,17 @@ class Pipeline(PipelineGraph[TaskPipelineNode, PipelineEdge]):
             result=await self.get_final_results(orchestrator.run_id),
         )
 
-    def dump_state(self) -> Dict[str, Any]:
-        """Dump the current state of the pipeline and its components
-        to a serializable dictionary.
+    def dump_state(self, run_id: str) -> Dict[str, Any]:
+        """Dump the current state of the pipeline and its components to a serializable dictionary.
+
+        Args:
+            run_id: The run_id that was used when the pipeline was executed
 
         Returns:
             dict[str, Any]: A serializable dictionary containing the pipeline state
         """
-        pipeline_state = {
+        pipeline_state: Dict[str, Any] = {
+            "run_id": run_id,
             "components": {},
             "store": self.store.dump() if hasattr(self.store, "dump") else {},
             "final_results": self.final_results.dump()
@@ -636,8 +640,20 @@ class Pipeline(PipelineGraph[TaskPipelineNode, PipelineEdge]):
         if "final_results" in state and hasattr(self.final_results, "load"):
             self.final_results.load(state["final_results"])
 
-    async def run_until(self, data: Dict[str, Any], stop_after: str) -> Dict[str, Any]:
-        """Run the pipeline until a specific component and return the state."""
+    async def run_until(
+        self, data: Dict[str, Any], stop_after: str, state_file: Optional[str] = None
+    ) -> Dict[str, Any]:
+        """
+        Run the pipeline until a specific component and return the state.
+
+        Args:
+            data (Dict[str, Any]): The input data for the pipeline.
+            stop_after (str): The name of the component to stop after.
+            state_file (Optional[str]): If provided, save the state to this file as JSON.
+
+        Returns:
+            Dict[str, Any]: The serialized state of the pipeline after execution.
+        """
         logger.debug("PIPELINE START (RUN UNTIL)")
         start_time = default_timer()
         self.invalidate()
@@ -649,17 +665,50 @@ class Pipeline(PipelineGraph[TaskPipelineNode, PipelineEdge]):
         logger.debug(
             f"PIPELINE FINISHED (RUN UNTIL) {orchestrator.run_id} in {end_time - start_time}s"
         )
-        return self.dump_state()
+        state = self.dump_state(orchestrator.run_id)
+        if state_file:
+            with open(state_file, "w", encoding="utf-8") as f:
+                json.dump(state, f, ensure_ascii=False, indent=2)
+        return state
 
     async def resume_from(
-        self, state: Dict[str, Any], data: Dict[str, Any], start_from: str
-    ) -> "PipelineResult":
-        """Resume pipeline execution from a specific component using a saved state."""
+        self,
+        state: Optional[
+            Dict[str, Any]
+        ],  # Required but can be None if state_file is provided
+        data: Dict[str, Any],
+        start_from: str,
+        state_file: Optional[str] = None,
+    ) -> PipelineResult:
+        """
+        Resume pipeline execution from a specific component using a saved state.
+
+        Args:
+            state (Optional[Dict[str, Any]]): The serialized pipeline state. Required, but can be None if state_file is provided.
+            data (Dict[str, Any]): Additional input data for the pipeline.
+            start_from (str): The name of the component to start execution from.
+            state_file (Optional[str]): If provided, load the state from this file as JSON. Required if state is None.
+
+        Returns:
+            PipelineResult: The result of the pipeline execution.
+
+        Raises:
+            ValueError: If neither state nor state_file is provided.
+        """
+        if state_file:
+            with open(state_file, "r", encoding="utf-8") as f:
+                state = json.load(f)
+        if state is None:
+            raise ValueError("No state provided for resume_from.")
         self.load_state(state)
+        run_id = state.get("run_id")
+        if not run_id:
+            raise ValueError("No run_id found in state. Cannot resume execution.")
         logger.debug("PIPELINE START (RESUME FROM)")
         start_time = default_timer()
         self.validate_input_data(data)
         orchestrator = Orchestrator(self, start_from=start_from)
+        orchestrator.run_id = run_id  # Use the original run_id
         logger.debug(f"PIPELINE ORCHESTRATOR: {orchestrator.run_id}")
         await orchestrator.run(data)
         end_time = default_timer()
--- a/src/neo4j_graphrag/experimental/pipeline/pipeline.py
+++ b/src/neo4j_graphrag/experimental/pipeline/pipeline.py
@@ -18,7 +18,7 @@ import logging
 import warnings
 from collections import defaultdict
 from timeit import default_timer
-from typing import Any, Optional, AsyncGenerator
+from typing import Any, Optional, AsyncGenerator, Dict
 import asyncio
 
 from neo4j_graphrag.utils.logging import prettify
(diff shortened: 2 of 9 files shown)


docs/source/user_guide_pipeline.rst
```diff
@@ -133,6 +133,36 @@ can be added to the visualization by setting `hide_unused_outputs` to `False`:
...
+*************************
+Pipeline State Management
+*************************
+
+Pipelines support checkpointing and resumption through state management features:
+
+.. code:: python
+
+    # Run pipeline until a specific component
+    state = await pipeline.run_until(data, stop_after="component_name", state_file="state.json")
```
stellasia: If we are returning the state, I think it would be cleaner let to the user save it to file or not, I'm not sure adding the `state_file` option is helpful.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -579,3 +580,142 @@ async def run(self, data: dict[str, Any]) -> PipelineResult:
...
+        Args:
+            run_id: The run_id that was used when the pipeline was executed
+
+        Returns:
+            dict[str, Any]: A serializable dictionary containing the pipeline state
+        """
+        pipeline_state: Dict[str, Any] = {
+            "run_id": run_id,
+            "components": {},
+            "store": self.store.dump() if hasattr(self.store, "dump") else {},
```
stellasia: The store must obey to a certain interface so we can assume `dump` will always be there.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -579,3 +580,142 @@ async def run(self, data: dict[str, Any]) -> PipelineResult:
...
+                state = json.load(f)
+        if state is None:
+            raise ValueError("No state provided for resume_from.")
+        self.load_state(state)
+        run_id = state.get("run_id")
+        if not run_id:
+            raise ValueError("No run_id found in state. Cannot resume execution.")
+        logger.debug("PIPELINE START (RESUME FROM)")
+        start_time = default_timer()
+        self.validate_input_data(data)
```
stellasia: Here we revalidate the pipeline, which is also where the `param_mapping` and `missing_input` fields are set, so I think these parameters do not need to be saved in the state.

tests/unit/experimental/pipeline/components.py
```diff
@@ -63,3 +64,18 @@ def __init__(self, sleep: float = 1.0) -> None:
     async def run(self, number1: int, number2: int = 2) -> IntResultModel:
         await asyncio.sleep(self.sleep)
         return IntResultModel(result=number1 * number2)
+
+
+class StatefulComponent(Component):
```
stellasia: So far, we considered components as "stateless" because a given component can be run concurrently between different pipeline.run calls. Unless there is a strong reason to change that, I'd prefer not to open that door now (we can rediscuss it of course, but this PR is already quite complex)

src/neo4j_graphrag/experimental/pipeline/pipeline.py
stellasia: Btw, shouldn't `ResultStore` have a `dump` method and an implementation of it for the `InMemoryStore`?

docs/source/user_guide_pipeline.rst
```diff
@@ -133,6 +133,36 @@ can be added to the visualization by setting `hide_unused_outputs` to `False`:
...
+
+    # Resume pipeline from a specific component
+    result = await pipeline.resume_from(state, data, start_from="component_name")
+
+    # Alternatively, load state from file
+    result = await pipeline.resume_from(None, data, start_from="component_name", state_file="state.json")
```
stellasia: Also because of that, `None` as first argument is not super nice IMHO :)

docs/source/user_guide_pipeline.rst
```diff
@@ -133,6 +133,36 @@ can be added to the visualization by setting `hide_unused_outputs` to `False`:
...
+
+The state contains:
+- Pipeline configuration (parameter mappings between components and validation state)
+- Execution results (outputs from completed components stored in the ResultStore)
+- Final pipeline results from previous runs
```
stellasia: Why do we need `final_results`  in the state?

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -579,3 +580,142 @@ async def run(self, data: dict[str, Any]) -> PipelineResult:
...
+        logger.debug(
+            f"PIPELINE FINISHED (RUN UNTIL) {orchestrator.run_id} in {end_time - start_time}s"
+        )
+        state = self.dump_state(orchestrator.run_id)
+        if state_file:
+            with open(state_file, "w", encoding="utf-8") as f:
+                json.dump(state, f, ensure_ascii=False, indent=2)
+        return state
+
+    async def resume_from(
```
stellasia: What happens if the user loads a state that does not correspond to this pipeline? (not saying we should check, just wondering if the behavior will be clear enough)

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -579,3 +580,142 @@ async def run(self, data: dict[str, Any]) -> PipelineResult:
...
+        # load store data if store has load method
+        if "store" in state and hasattr(self.store, "load"):
+            self.store.load(state["store"])
+
+        # load final results if it has load method
+        if "final_results" in state and hasattr(self.final_results, "load"):
+            self.final_results.load(state["final_results"])
+
+    async def run_until(
+        self, data: Dict[str, Any], stop_after: str, state_file: Optional[str] = None
```
stellasia: What is the reason not to use the same interface as the orchestrator (almost), ie changing the run method signature to:

```
pipeline.run(..., from : str | None = "", until : str | None = "")
```

The process could be:
```
pipeline.run(..., until="b")
state = pipeline.dump_state()

# ...
pipeline.load_state(state)
pipeline.run(..., from="b")
```

(not saying it's better, though it would maybe prevent some code duplication, but asking mainly because I see this is what's being used in the orchestrator :) )

NathalieCharbel: I think the main reason for that was that for pipeline I wanted initially to make partial execution and state handling explicit operations and separate concerns from running it normally. It was btw the same case for the orchestrator until I realised that the same interface can be reused and the different behaviour can be handled internally and is basically just adding conditions controlling logic flow (and it is not user-facing method). 
That being said, you are right that using the same interface for pipeline is simpler and avoids code duplication, but there is more to add your proposed signature:
1- adding `state` parameter to `run()` for cases loading for a specific checkpoint.
2- handling different return types?? (`PipelineResult` vs state `Dict`)

NathalieCharbel: for my 2nd point, if you originally meant that "run until" actually only stops the execution until a certain component and not even save the state (this being handled by a different function), then I think this is easier to implement and apart from the arguments above, we need to add state as well.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
NathalieCharbel: it was the case until we decided not to store the state in memory. is this still needed in your opinion?

tests/unit/experimental/pipeline/components.py
NathalieCharbel: Yea I think dump and load state at the component level is a separate thing. the corresponding methods were anyway abstract. I'll remove this as there was no specific reason for introducing them initially.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
NathalieCharbel: if we are dumping the pipeline state in one run and resuming it or loading its state in another run, where will these param come from if we haven't saved them initially in the state?

docs/source/user_guide_pipeline.rst
NathalieCharbel: see my reply [here](https://github.com/neo4j/neo4j-graphrag-python/pull/352/files/9a57046ea23403f09c0b148fb01e093918ce3971#r2138035703). same applies to `final_results` or am I missing something?

src/neo4j_graphrag/experimental/pipeline/pipeline.py
NathalieCharbel: I think the input data will not be validated at some point of the execution, when the `validate_input_data` is called right?

src/neo4j_graphrag/experimental/pipeline/pipeline.py
stellasia: I think there is a confusion here: the component results within the pipeline are already stored in memory (through the InMemoryResultStore). 

My understanding regarding the in memory state discussion is that we do not want to store the state ONLY in memory, so that users are able to dump/load a state - and actually, returning the state from the `run_until` method covers this use case.

I've tried this: in the pipeline in `examples/customize/build_graph/pipeline/lexical_graph_builder_from_text.py`, if I change the run to:

```
state = await pipe.run_until(pipe_inputs, stop_after="splitter")
print(state)
```

then the store is an empty dict `'store': {}` while it should contain the splitter result? And I think it all related to the fact that there is no `dump` method in the stores, so we fallback on an empty store.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
stellasia: From the pipeline validation. Assuming we will rerun the pipeline with the same input data (ie even inputs for components that have already run). Otherwise we'll have to change the validation.

docs/source/user_guide_pipeline.rst
stellasia: - final results potentially contains results for multiple runs, I think we do not care about the results for other runs
- at the moment, the pipeline only returns the result of the last component actually, that's why I'm not sure about the usefulness of this parameter at the moment, but in the future we can imagine configuring what is being returned, and in that case we might need to access previous results that are already stored in the final result store, so maybe we want to keep it for this use case.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
stellasia: I don't think so, `validate_input_data` checks that the mandatory input parameters for each component, that are not propagated through a connection, are provided in the input data. But if we load the wrong state, some components that should have results in the store won't, and that's a different story, it will fail later:

```
  File "neo4j-graphrag-python/examples/customize/build_graph/pipeline/lexical_graph_builder_from_text.py", line 78, in main
    state = await pipe.resume_from({"run_id": "something", "store": {}}, pipe_inputs, start_from="writer")
  File "neo4j-graphrag-python/src/neo4j_graphrag/experimental/pipeline/pipeline.py", line 713, in resume_from
    await orchestrator.run(data)
  File "neo4j-graphrag-python/src/neo4j_graphrag/experimental/pipeline/orchestrator.py", line 291, in run
    await asyncio.gather(*tasks)
  File "neo4j-graphrag-python/src/neo4j_graphrag/experimental/pipeline/orchestrator.py", line 85, in run_task
    inputs = await self.get_component_inputs(task.name, param_mapping, data)
  File "neo4j-graphrag-python/src/neo4j_graphrag/experimental/pipeline/orchestrator.py", line 242, in get_component_inputs
    value = component_result.get(output_param)
AttributeError: 'NoneType' object has no attribute 'get'
```

src/neo4j_graphrag/experimental/pipeline/pipeline.py
NathalieCharbel: I see!

src/neo4j_graphrag/experimental/pipeline/stores.py
```diff
@@ -115,3 +133,19 @@ def all(self) -> dict[str, Any]:
 
     def empty(self) -> None:
         self._data = {}
+
+    def dump(self) -> dict[str, Any]:
```
stellasia: Should we dump/load state only for a specific run_id?

NathalieCharbel: I think so! But we should also ensure that when a pipeline is run again from a resumed state, it uses the same `run_id`. I’m not sure the current changes support that, I’ll look into the details.

src/neo4j_graphrag/experimental/pipeline/stores.py
```diff
@@ -134,18 +138,38 @@ def all(self) -> dict[str, Any]:
...
+            run_id (str): The run_id to load data for
             state (dict[str, Any]): A serializable dictionary containing the store state
         """
-        self._data = state.copy()
+        # clear existing data for this run_id first
+        run_id_prefix = f"{run_id}:"
+        keys_to_remove = [
+            key for key in self._data.keys() if key.startswith(run_id_prefix)
+        ]
+        for key in keys_to_remove:
```
stellasia: So here we are removing all results from a previous run with this run_id, right?

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -140,6 +140,7 @@ def __init__(
         }
         """
         self.missing_inputs: dict[str, list[str]] = defaultdict()
+        self._current_run_id: Optional[str] = None
```
stellasia: This can not be saved in the Pipeline instance, since concurrent runs will override it.

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -600,20 +650,23 @@ async def run(
             result=await self.get_final_results(orchestrator.run_id),
         )
 
-    def dump_state(self, run_id: str) -> Dict[str, Any]:
+    def dump_state(self) -> Dict[str, Any]:
```
stellasia: I guess we will need the run_id as parameter here as a consequence of my first comment

src/neo4j_graphrag/experimental/pipeline/pipeline.py
```diff
@@ -622,15 +675,66 @@ def load_state(self, state: Dict[str, Any]) -> None:
...
+            if len(parts) >= 2:
+                component_name = parts[1]
+                stored_components.add(component_name)
+
+        # get current pipeline component names
+        current_components = set(self._nodes.keys())
+
+        # check if stored components are a subset of current components
+        # this allows for the pipeline to have additional components, but not missing ones
+        missing_components = stored_components - current_components
```
stellasia, line 730: If I get it right, missing components will contain all components that have not been run yet?

stellasia: Nope, I got it wrong :) looks good!

src/neo4j_graphrag/experimental/pipeline/stores.py
NathalieCharbel: yes!

src/neo4j_graphrag/experimental/pipeline/pipeline.py
NathalieCharbel: I will move it back to `dump()` function. I think we should maintain creating different `run_id`s even after resuming the same pipeline and dump the state based on previous ones. we could keep track of `run_id`s of the same pipeline in the state. This should resolve the concurrency issue, right?

PR #371, opened 2025-06-25
Add API rate limit handler

NathalieCharbel:
# Description

- New `RateLimitHandler` interface
- Default behavior: retry and exponential backoff using Tenacity
- Provider-specific errors are converted to `RateLimitError`
- Basic users get rate limiting out-of-the-box
- Advanced users can implement custom rate limiting strategies via the `RateLimitHandler` interface

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


docs/source/user_guide_rag.rst
```diff
@@ -294,6 +294,71 @@ Here's an example using the Python Ollama client:
...
+
+Default Rate Limit Handler
+--------------------------
+
+Rate limiting is enabled by default for all LLM instances with the following configuration:
+
+- **Max attempts**: 3
+- **Min wait**: 1.0 seconds  
+- **Max wait**: 60.0 seconds
+- **Multiplier**: 2.0 (exponential backoff)
```
stellasia, line 310: Can we add an example here that demonstrates how to change these default parameters?

If I understand properly, something like:

```
llm = OpenAILLM(
   # ...
   rate_limit_handler=RetryRateLimitHandler(max_attempts=10)
)
```

src/neo4j_graphrag/llm/rate_limit.py
```diff
@@ -0,0 +1,299 @@
...
+    Returns:
+        A decorator function.
+    """
+
+    def decorator(func: F) -> F:
+        @functools.wraps(func)
+        def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
+            # Use instance handler if available, otherwise use provided handler or default
+            instance_handler = (
+                getattr(self, "_rate_limit_handler", None)
```
stellasia: Isn't `getattr(..., None)` already handling the cases where the attribute is not there (ie is `hasattr` needed)

src/neo4j_graphrag/llm/rate_limit.py
```diff
@@ -0,0 +1,299 @@
...
+    from tenacity import (
+        retry,
+        stop_after_attempt,
+        wait_exponential,
+        retry_if_exception_type,
+        before_sleep_log,
+    )
+
+    TENACITY_AVAILABLE = True
+except ImportError:
```
stellasia: Since this is a default behavior, I think we can add tenacity as a mandatory dependency, and assume it is always there.

src/neo4j_graphrag/llm/rate_limit.py
```diff
@@ -0,0 +1,299 @@
...
+
+            def inner_func() -> Any:
+                try:
+                    return func(self, *args, **kwargs)
+                except Exception as e:
+                    if is_rate_limit_error(e):
+                        raise convert_to_rate_limit_error(e)
+                    raise
+
+            return active_handler.handle_sync(inner_func)()
```
stellasia: Do we need to call the returned function here?

src/neo4j_graphrag/llm/cohere_llm.py
```diff
@@ -69,7 +71,7 @@ def __init__(
                 """Could not import cohere python client.
                 Please install it with `pip install "neo4j-graphrag[cohere]"`."""
             )
-        super().__init__(model_name, model_params)
+        super().__init__(model_name, model_params, rate_limit_handler)
         self.cohere = cohere
         self.cohere_api_error = cohere.core.api_error.ApiError
```
stellasia, line 81: We don't need to repeat the decorator in the subclasses? I mean having:

```
 @rate_limit_handler()
 def invoke(
```

everywhere.

src/neo4j_graphrag/llm/rate_limit.py
```diff
@@ -0,0 +1,299 @@
...
+            stop=stop_after_attempt(self.max_attempts),
+            wait=wait_exponential(
+                multiplier=self.multiplier,
+                min=self.min_wait,
+                max=self.max_wait,
+            ),
+            before_sleep=before_sleep_log(logger, logging.WARNING),
+        )
+        @functools.wraps(func)
+        async def wrapper(*args: Any, **kwargs: Any) -> Any:
```
stellasia: Did you manage to test it in async mode? I'm worried we'll have some weird behaviour with the concurrent mode if retries are using `await asyncio.sleep(..)`, that would allow other calls to be "enqueued" and almost all of the following calls will have to wait, which will still fail pretty quickly since they are all waiting and retrying at the same time.

docs/source/user_guide_rag.rst
```diff
@@ -294,6 +294,71 @@ Here's an example using the Python Ollama client:
...
+You can customize the rate limiting behavior by creating your own rate limit handler:
+
+.. code:: python
+
+    from neo4j_graphrag.llm import AnthropicLLM
+    from neo4j_graphrag.llm.rate_limit import RateLimitHandler
+    
+    class CustomRateLimitHandler(RateLimitHandler):
+        """Implement your custom rate limiting strategy."""
+        # Implement required methods: handle_sync, handle_async
```
stellasia, line 354: Which parameters are these methods getting as input?

src/neo4j_graphrag/llm/rate_limit.py
stellasia: Which is what we do in the pyproject.toml btw, tenacity is not an optional dependency, so no risk to have the same issue we had with spacy ;)

docs/source/user_guide_rag.rst
NathalieCharbel: Done

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: yes 😮!
I don't even think that `rate_limit_handler` will be `None` anyway since we always fallback to default one in the `LLMInterface` constructor. 
I also further simplified the structure of the decorator.

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: yes :) I believe this was mistakenly added.

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: I am not so familiar with defining decorators. But don't we still need this to return the retry version of the inner function?

src/neo4j_graphrag/llm/cohere_llm.py
NathalieCharbel: You're right I thought defining them in abstract method will lead to propagating them in subclasses but I was completely wrong.

NathalieCharbel: I added them in subclasses methods

docs/source/user_guide_rag.rst
NathalieCharbel: `func` representing sync/async functions that need rate limit handling applied to them as it is the case for current `RetryRateLimitHandler` (the default one)

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: I havn't thought about that but we should definitely avoid this. One solution could be adding randomisation to delay retries. something like the below:
```
  def handle_async(self, func: AF) -> AF:
        @retry(
            retry=retry_if_exception_type(RateLimitError),
            stop=stop_after_attempt(self.max_attempts),
            wait=wait_exponential(
                multiplier=self.multiplier,
                min=self.min_wait,
                max=self.max_wait,
            ) + (wait_random(0, 1) if self.jitter else wait_fixed(0)),  # add jitter
            before_sleep=before_sleep_log(logger, logging.WARNING),
        )
```

wdyt?

stellasia: To be tested ;)

docs/source/user_guide_rag.rst
stellasia: Do you think this interface could be used to limit the number of requests per second for instance?

src/neo4j_graphrag/llm/rate_limit.py
stellasia: The decorator should return a function, that's why I tend to think the final `()` are not needed, but I haven't tested.

docs/source/user_guide_rag.rst
NathalieCharbel: If we would like to extend our strategy for rate limit handling and do throttling, we can define a new subclass:

```
class ThrottlingRateLimitHandler(RateLimitHandler):
    
   def __init__(self, requests_per_window: int = 10, window_size: float = 1.0):
        self.requests_per_window = requests_per_window
        self.window_size = window_size
        ....

    
    def handle_sync(self, func: F) -> F:
        @functools.wraps(func)
        def wrapper(*args: Any, **kwargs: Any) -> Any:
            # handle throttling here
            return func(*args, **kwargs)
        return wrapper
        
     .....
```
then we can basically use the same decorators with the invoke functions of the LLM classes, we just need to ensure we pass the right rate limit handler in their constructors.

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: looks like Tenacity already has a function that does exponential backoff + jitter, so I ended up using it [8cfacb4](https://github.com/neo4j/neo4j-graphrag-python/pull/371/commits/8cfacb426a8c48c2bde1acbf1f6b19034fe165e6)

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: If we omit the trailing (), the decorator would return the wrapped function object instead of the function’s result.

stellasia: Ok, then shouldn't the return type of `rate_limit_handler` by `Any` instead of `F`?

docs/source/user_guide_rag.rst
stellasia: Maybe at some point we'll need to pass more arguments to the handle_sync method, like the input query for instance, if people want to use some token-based throttling. But no need to overcomplicate things for now.

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: but doesn't using `F` preserve more type safety since at least at call sites the method has the same type as it has before decoration?

stellasia: It would if we were returning a function, but based on the above discussion I understood that we're returning the result of the method? Also, what happens when removing the `type ignore` comment?

stellasia:
Just thinking as I was writing the release notes (sorry!): do we need to update [this example](https://github.com/neo4j/neo4j-graphrag-python/blob/main/examples/customize/llms/custom_llm.py) as well to show how to use rate limiter for custom implementation?

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: The inner function and the wrapper are returning the result of the original function, but the decorator is returning a function. 
I am following the below pattern:
```
def my_decorator(func: Callable) -> Callable:
    def wrapper(*args, **kwargs) -> Any:
        def inner(): -> Any
            print("Inner function logic")
            return func(*args, **kwargs)
        return inner()
    return wrapper
```
- `my_decorator`: takes a function `func` and returns a new function (`wrapper`)
- `wrapper`: actual replacement for `func`. It takes arguments, defines an inner function, and returns the result of inner().
- inner: helper inside wrapper that encapsulates additional logic, then calls and returns the result of `func`.

obviously mypy will complain about incompatible return types, thus type ignore is added. Otherwise your suggestion of changing return type of the decorator to `Any` will pass the mypy checks but the thing here is that the decorator could be misused afterwards. 

I don't mind changing it, but I'd like to understand what's the best solution in this case or am I following the wrong pattern?

PR #374, opened 2025-07-01
Explicit configuration of schema extraction

stellasia:
# Description

`schema_` parameter can be configured with a string "NO_EXTRACTION" or "AUTO_EXTRACTION". Internally, this is changed to either an empty schema or None, so that the rest of the process is unchanged.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/docs/source/user_guide_kg_builder.rst
+++ b/docs/source/user_guide_kg_builder.rst
@@ -131,8 +131,19 @@ This schema information can be provided to the `SimpleKGBuilder` as demonstrated
         # ...
     )
 
-.. note::
-   By default, if no schema is provided to the SimpleKGPipeline, automatic schema extraction will be performed using the LLM (See the :ref:`Automatic Schema Extraction`).
+
+Schema Parameter Behavior
+-------------------------
+
+The `schema` parameter controls how entity and relation extraction is performed:
+
+* **AUTO_EXTRACTION**: ``schema="AUTO_EXTRACTION"`` or (``schema=None``)
+  The schema is automatically extracted from the input text once. This guiding schema is then used to structure entity and relation extraction for all chunks. This guarantees all chunks have the same guiding schema.
+  (See :ref:`Automatic Schema Extraction`)
+
+* **NO_EXTRACTION**: ``schema="NO_EXTRACTION"`` or empty schema (``{"node_types": ()}``)
+  No schema extraction is performed. Entity and relation extraction proceed without a predefined or derived schema, resulting in unguided extraction.
+
 
 Extra configurations
 --------------------
--- a/docs/source/user_guide_kg_builder.rst
+++ b/docs/source/user_guide_kg_builder.rst
@@ -137,12 +137,12 @@ Schema Parameter Behavior
 
 The `schema` parameter controls how entity and relation extraction is performed:
 
-* **AUTO_EXTRACTION**: ``schema="AUTO_EXTRACTION"`` or (``schema=None``)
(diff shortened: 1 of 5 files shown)


docs/source/user_guide_kg_builder.rst
```diff
@@ -131,8 +131,19 @@ This schema information can be provided to the `SimpleKGBuilder` as demonstrated
...
+Schema Parameter Behavior
+-------------------------
+
+The `schema` parameter controls how entity and relation extraction is performed:
+
+* **AUTO_EXTRACTION**: ``schema="AUTO_EXTRACTION"`` or (``schema=None``)
+  The schema is automatically extracted from the input text once. This guiding schema is then used to structure entity and relation extraction for all chunks. This guarantees all chunks have the same guiding schema.
+  (See :ref:`Automatic Schema Extraction`)
+
+* **NO_EXTRACTION**: ``schema="NO_EXTRACTION"`` or empty schema (``{"node_types": ()}``)
```
NathalieCharbel: just double checking, there is no way `schema` can be empty `{}` i.e. without even `node_types` because it won't be valid right?

docs/source/user_guide_kg_builder.rst
```diff
@@ -131,8 +131,19 @@ This schema information can be provided to the `SimpleKGBuilder` as demonstrated
...
+  No schema extraction is performed. Entity and relation extraction proceed without a predefined or derived schema, resulting in unguided extraction.
+
```
NathalieCharbel, line 146: related to my comment below (about the name of "NO_EXTRACTION", if you keep the name as is maybe saying explicitly that this used to bypass auto schema extraction would clarify its purpose?

tests/unit/experimental/pipeline/config/template_pipeline/test_simple_kg_builder.py
```diff
@@ -138,6 +138,14 @@ def test_simple_kg_pipeline_config_manual_schema() -> None:
     assert isinstance(config._get_schema(), SchemaBuilder)
 
 
+def test_simple_kg_pipeline_config_literal_schema_validation() -> None:
+    config = SimpleKGPipelineConfig(schema="NO_EXTRACTION")  # type: ignore
```
NathalieCharbel: looking at it here, I am wondering if "NO_EXTRACTION" is the best term to use? when putting it next to "AUTO_EXTRACTION" it makes sense but here it is a bit ambiguous? would "NO_SCHEMA" or "NO_GUIDING_SCHEMA" be better? as it does not only cover not being extracted but not provided at all? I am not strongly opinionated about it though as it is well-documented...

docs/source/user_guide_kg_builder.rst
stellasia: Yes, totally empty dict will fail pydantic validation because `node_types` is not optional.

docs/source/user_guide_kg_builder.rst
stellasia: Yes, added a line even after renaming.

tests/unit/experimental/pipeline/config/template_pipeline/test_simple_kg_builder.py
stellasia: Updated after private discussion

PR #398, opened 2025-08-01
Function to visualize a GraphSchema object or dict

stellasia:
# Description

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/utils/schema.py
+++ b/src/neo4j_graphrag/experimental/utils/schema.py
@@ -1,0 +1,101 @@
+from typing import Any, Union
+
+try:
+    from neo4j_viz import VisualizationGraph, Node, Relationship
+except ImportError:
+    VisualizationGraph = Node = Relationship = None  # type: ignore
+
+from neo4j_graphrag.experimental.components.schema import (
+    GraphSchema,
+    NodeType,
+    PropertyType,
+)
+
+
+def schema_visualization(
+    schema: Union[dict[str, Any], GraphSchema],
+) -> VisualizationGraph:
+    """Helper function to visualize a GraphSchema using the neo4j-viz library.
+
+    Usage:
+
+    .. code:: python
+
+        VG = schema_visualization(schema)
+        html = VG.render()
+
+        # in Jupyter:
+        display(html)
+
+        # to save the generated HTML
+        with open("my_schema.html", "w") as f:
+            f.write(html.data)
+    """
+    if VisualizationGraph is None:
+        raise ImportError(
+            "Please install neo4j-viz to use the graph schema visualization feature: pip install neo4j-viz"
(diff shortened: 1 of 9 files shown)


tests/unit/experimental/utils/test_schema.py
```diff
@@ -0,0 +1,107 @@
...
+            (
+                "Person",
+                "BORN_IN",
+                "Location",
+            ),  # invalid pattern, "Location" node type not defined
+        ],
+    }
+
+
+@patch("neo4j_graphrag.experimental.utils.schema.neo4j_viz", None)
```
NathalieCharbel: should patch `VisualizationGraph` instead?

src/neo4j_graphrag/experimental/utils/schema.py
```diff
@@ -0,0 +1,115 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from typing import Any, Union
+
+try:
+    from neo4j_viz import VisualizationGraph, Node, Relationship
+except ImportError:
+    VisualizationGraph = Node = Relationship = None  # type: ignore
```
NathalieCharbel, line 20: should we use `TYPE_CHECKING` instead? what should be a rule of thumb for cases like this to avoid handling them differently across the package?

NathalieCharbel (commented):
minor comments, otherwise looks great and a nice feature to have :)

tests/unit/experimental/utils/test_schema.py
stellasia: Yes, good catch!

src/neo4j_graphrag/experimental/utils/schema.py
stellasia: They serve different purposes: try/except is needed to avoid import errors if dep is not installed, which should be used by default. `TYPE_CHECKING` should only be used if some dependencies are required during `mypy` checks but can not be imported "normally" (due to circular imports. However, looking more closely at how we use `TYPE_CHECKING`, it seems some of them are actually not required, I'll create a card so that we can clean them.

NathalieCharbel: I couldn't find the card about this cleanup. so I just create this [one](https://trello.com/c/HNS6fkSC/1572-typechecking-cleanup).

PR #403, opened 2025-08-04
Fix cleanup ops for inferred schema

NathalieCharbel:
# Description

Fixes `AttributeError` that occurs during schema cleanup in `SchemaFromTextExtractor` when LLMs return node/relationship types as strings instead of the expected dictionary format.

Changes:
- Added defensive type checking in `_filter_nodes_without_labels` and `_filter_relationships_without_labels` methods
- Updated unit tests 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/components/schema.py
+++ b/src/neo4j_graphrag/experimental/components/schema.py
@@ -498,55 +498,38 @@ class SchemaFromTextExtractor(Component):
 
         return filtered_patterns
 
+    def _filter_items_without_labels(
+        self, items: List[Dict[str, Any]], item_type: str
+    ) -> List[Dict[str, Any]]:
+        """Filter out items that have no valid labels."""
+        filtered_items = []
+        for item in items:
+            if isinstance(item, str):
+                if item and " " not in item and not item.startswith("{"):
+                    filtered_items.append({"label": item})
+                elif item:
+                    logging.info(
+                        f"Filtering out {item_type} with invalid label: {item}"
+                    )
+            elif isinstance(item, dict) and item.get("label"):
+                filtered_items.append(item)
+            else:
+                logging.info(f"Filtering out {item_type} with missing label: {item}")
+        return filtered_items
+
     def _filter_nodes_without_labels(
         self, node_types: List[Dict[str, Any]]
     ) -> List[Dict[str, Any]]:
-        """
-        Filter out node types that have no labels.
-
-        Args:
-            node_types: List of node type definitions.
-
-        Returns:
-            Filtered list of node types containing only those with valid labels.
-        """
-        filtered_nodes = []
-        for node_type in node_types:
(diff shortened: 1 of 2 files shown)


src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -512,7 +512,10 @@ def _filter_nodes_without_labels(
         """
         filtered_nodes = []
         for node_type in node_types:
-            if node_type.get("label"):
+            if isinstance(node_type, str):
```
stellasia: Are we confident the string is a label? Or can it be something more complex?

tests/unit/experimental/components/test_schema.py
```diff
@@ -794,12 +794,9 @@ def schema_json_with_nodes_without_labels() -> str:
...
             },
-            {
-                "label": "Organization",
-                "properties": [
-                    {"name": "name", "type": "STRING"}
-                ]
-            }
+            "Organization",
+            "",
+            "Company"
```
stellasia: Do not change anything but just our of curiosity: are the "Organization" and "Company" test cases actually testing different scenarios ?

src/neo4j_graphrag/experimental/components/schema.py
NathalieCharbel: does it matter for the purpose of this function?  (i.e. removing empty string or empty labels). The response will be anyway processed for validation in a later step.

tests/unit/experimental/components/test_schema.py
NathalieCharbel: no I think there were added for the below pattern. but does not matter for testing nodes without labels, we could have only one here.

src/neo4j_graphrag/experimental/components/schema.py
stellasia: If it's a string, it will be validated, we do not have much more validation regarding label, I was more thinking about cases where the string will be another json, or a description of the node/relationship rather than a real label.

NathalieCharbel: I added more rules to cater for these cases. [2fe7800](https://github.com/neo4j/neo4j-graphrag-python/pull/403/commits/2fe7800500d046bce414404b998b41f7d125ddc4)

stellasia (approved):
I was a bit afraid that we will fix things without knowing what is broken exactly. But in practice, we don't know what the LLM will choose as a label even if the JSON structure is correct, so I convinced myself that this fix is good enough, or at least consistent with the rest of the component :) 

Thanks Nathalie!

PR #416, opened 2025-08-26
Update LLMInterface to restore LC compatibility

stellasia:
# Description

Restore LangChain chat models compatibility in the LLMInterface:

- No breaking change.
- The `input` parameter for all `invoke` method now accepts `Union[str, list[LLMMessage]]`, where `LLMMessage` is a `TypedDict` with role/content keys.
- This signature, including `system_instructions` and `message_history` is translated into a `list[LLMMessage]` that is being passed down to the corresponding `_invoke` method.
- Subclasses now only need to implement the private `_invoke` methods.
    - Advantage: no need to repeat the parameter conversion nor the rate limit decorators
- Updated unit tests:
    - Many tests from each implementation are removed because the logic has been moved elsewhere (either in utils or the base class)

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low (many changes, but just refactoring in the end)

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


examples/customize/llms/custom_llm.py
```diff
@@ -18,38 +17,26 @@ def __init__(
...
 
-llm = CustomLLM(
-    ""
-)  # if rate_limit_handler and async_rate_limit_handler decorators are used, the default rate limit handler will be applied automatically (retry with exponential backoff)
+llm = CustomLLM("")
 res: LLMResponse = llm.invoke("text")
 print(res.content)
 
-# If rate_limit_handler and async_rate_limit_handler decorators are used and you want to use a custom rate limit handler
-# Type variables for function signatures used in rate limit handlers
```
NathalieCharbel, line 52: the above comments on how to customise a rate limit handler are still valid no?

stellasia: Yes, actually only the first "If" is not needed anymore, as the rate limit handler is applied to all LLMs now. If reintroduced the rest of the comment.

src/neo4j_graphrag/llm/base.py
```diff
@@ -124,6 +146,17 @@ def invoke_with_tools(
...
+        except ValidationError as e:
+            raise LLMGenerationError("Input validation failed") from e
+        return self._invoke_with_tools(messages, tools)
+
+    def _invoke_with_tools(
+        self, inputs: list[LLMMessage], tools: Sequence[Tool]
+    ) -> ToolCallResponse:
         raise NotImplementedError("This LLM provider does not support tool calling.")
 
     async def ainvoke_with_tools(
```
NathalieCharbel, line 163: shouldn't be preceded by the decorator  `@async_rate_limit_handler`

src/neo4j_graphrag/llm/mistralai_llm.py
```diff
@@ -75,38 +73,30 @@ def __init__(
...
+        for m in input:
+            if m["role"] == "system":
+                messages.append(SystemMessage(content=m["content"]))
+                continue
+            if m["role"] == "user":
+                messages.append(UserMessage(content=m["content"]))
+                continue
+            if m["role"] == "assistant":
+                messages.append(AssistantMessage(content=m["content"]))
+                continue
```
NathalieCharbel, line 88: should we handle cases where role is unknown? same as you do in the openai interface? same questions apply to other provider interfaces as well.

src/neo4j_graphrag/llm/base.py
```diff
@@ -55,20 +61,30 @@ def __init__(
         else:
             self._rate_limit_handler = DEFAULT_RATE_LIMIT_HANDLER
 
-    @abstractmethod
+    @rate_limit_handler
     def invoke(
         self,
-        input: str,
+        input: Union[str, List[LLMMessage], MessageHistory],
```
NathalieCharbel, line 67: so here we keep support for old input and we handle new input in the form of `List[LLMMessage]` for LC compatibility right? 
what is the plan for the future? should we deprecate the old input and only keep LLMMessage? If so should we warn the user about that now?

src/neo4j_graphrag/llm/base.py
stellasia: Yeeesss, thanks!

src/neo4j_graphrag/llm/base.py
stellasia: I was actually unsure we want to add a deprecation for now, I think we can decide on this later on.

PR #425, opened 2025-09-20
Add rate limit handling to embedders

NathalieCharbel:
# Description

This PR extends the rate limiting functionality (previously available only for LLMs) to all embedding providers, and ensures consistent error handling

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Low

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/embeddings/base.py
+++ b/src/neo4j_graphrag/embeddings/base.py
@@ -15,13 +15,28 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from typing import Any, Optional
+
+from neo4j_graphrag.llm.rate_limit import (
+    DEFAULT_RATE_LIMIT_HANDLER,
+    RateLimitHandler,
+)
 
 
 class Embedder(ABC):
     """
     Interface for embedding models.
     An embedder passed into a retriever must implement this interface.
+
+    Args:
+        rate_limit_handler (Optional[RateLimitHandler]): Handler for rate limiting. Defaults to retry with exponential backoff.
     """
+
+    def __init__(self, rate_limit_handler: Optional[RateLimitHandler] = None):
+        if rate_limit_handler is not None:
+            self._rate_limit_handler = rate_limit_handler
+        else:
+            self._rate_limit_handler = DEFAULT_RATE_LIMIT_HANDLER
 
     @abstractmethod
     def embed_query(self, text: str) -> list[float]:
--- a/src/neo4j_graphrag/embeddings/base.py
+++ b/src/neo4j_graphrag/embeddings/base.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any, Optional
+from typing import Optional
 
 from neo4j_graphrag.llm.rate_limit import (
     DEFAULT_RATE_LIMIT_HANDLER,
--- a/src/neo4j_graphrag/embeddings/base.py
+++ b/src/neo4j_graphrag/embeddings/base.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from typing import Optional
 
-from neo4j_graphrag.llm.rate_limit import (
+from neo4j_graphrag.utils.rate_limit import (
     DEFAULT_RATE_LIMIT_HANDLER,
     RateLimitHandler,
     rate_limit_handler,
--- a/src/neo4j_graphrag/embeddings/base.py
+++ b/src/neo4j_graphrag/embeddings/base.py
@@ -20,6 +20,7 @@ from typing import Optional
 from neo4j_graphrag.llm.rate_limit import (
     DEFAULT_RATE_LIMIT_HANDLER,
     RateLimitHandler,
+    rate_limit_handler,
 )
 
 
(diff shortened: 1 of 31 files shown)


src/neo4j_graphrag/llm/rate_limit.py
```diff
@@ -147,7 +147,7 @@ def handle_async(self, func: AF) -> AF:
 
 
 def is_rate_limit_error(exception: Exception) -> bool:
-    """Check if an exception is a rate limit error from any LLM provider.
+    """Check if an exception is a rate limit error from any LLM provider or embedder.
```
stellasia: Do you think we can move this file somewhere else now that it's not only used for LLMs?

tests/unit/embeddings/test_openai_embedder.py
```diff
@@ -92,3 +93,18 @@ def test_azure_openai_embedder_does_not_call_openai_client() -> None:
...
+def test_openai_embedder_error_handling(mock_import: Mock) -> None:
+    mock_openai = get_mock_openai()
+    mock_import.return_value = mock_openai
+
+    mock_openai.OpenAI.return_value.embeddings.create.side_effect = Exception(
+        "API Error"
+    )
+    embedder = OpenAIEmbeddings(api_key="my key")
+    with pytest.raises(
+        EmbeddingsGenerationError, match="Failed to generate embedding with OpenAI"
```
stellasia, line 111: I don't understand this test, shouldn't this error be retried? I'm not sure this test captures it? I would add a check to make sure the function is called the expected number of times.

src/neo4j_graphrag/embeddings/base.py
```diff
@@ -15,14 +15,29 @@
...
+    Args:
+        rate_limit_handler (Optional[RateLimitHandler]): Handler for rate limiting. Defaults to retry with exponential backoff.
     """
 
+    def __init__(self, rate_limit_handler: Optional[RateLimitHandler] = None):
+        if rate_limit_handler is not None:
+            self._rate_limit_handler = rate_limit_handler
+        else:
+            self._rate_limit_handler = DEFAULT_RATE_LIMIT_HANDLER
+
```
stellasia, line 41: We could consider having the `embed_query` method in the base class, that deals with the common logic of retries, and calls an `_embed_query` (or another name) in the subclasses. In this way, the implementation in the subclasses doesn't have to change (just update the method name):

In short:

base class:

```
@rate_limit_handler
def embed_query(self, query):
     try:
         return self._embed_query(query)
     except Exception as e:
         raise EmbeddingGenerationError() from e

@abstractmethod
def _embed_query(self, query):
    ...
```

and in the children:

```
# no need to add the decorator
def _embed_query(self, query):
   # rest of the code
```

(up to you if you want to do it this way, it's not a request for change now)

stellasia (commented):
Just worried about the tests, if you have some time to explain. The other comments are marginal and non blocking.

src/neo4j_graphrag/llm/rate_limit.py
NathalieCharbel: sure, I thought about that, but it would be a breaking change. unless we believe most people aren’t calling the rate limit handler directly and are just relying on the default one. I’m fine with moving it, but I wanted to point out the potential drawback.

tests/unit/embeddings/test_openai_embedder.py
NathalieCharbel: this wasn't meant to test retryable errors, which will be wrapped in RateLimitError.
But you're right, I should definitely add one to cover it.

NathalieCharbel: Done [4bc0ef5](https://github.com/neo4j/neo4j-graphrag-python/pull/425/commits/4bc0ef51184da834bc0391d7d8e3b9904323ae73)

src/neo4j_graphrag/embeddings/base.py
NathalieCharbel: Done [a245f73](https://github.com/neo4j/neo4j-graphrag-python/pull/425/commits/a245f73dd153e733b699c084154482880e92702f). 
However, I left the error handling part at the subclasses level since they are a bit different.

src/neo4j_graphrag/llm/rate_limit.py
stellasia: Good point on the breaking change! We could use something similar to this:
https://github.com/neo4j/neo4j-graphrag-python/blob/4541de2b39f55edcd19c577150eb4e0d3a72e126/src/neo4j_graphrag/llm/types.py#L9
(ie still making the import from llm work, but with a deprecation warning). Do you think it's worth it?

NathalieCharbel: Great! I'll go ahead and make this change asap. I agree it's worth it, having it in the right place definitely makes things cleaner.

NathalieCharbel: Done

src/neo4j_graphrag/llm/__init__.py
```diff
@@ -18,7 +18,7 @@
 from .mistralai_llm import MistralAILLM
 from .ollama_llm import OllamaLLM
 from .openai_llm import AzureOpenAILLM, OpenAILLM
-from .rate_limit import (
+from neo4j_graphrag.utils.rate_limit import (
```
stellasia: Should this import be deprecated as well? (I mean importing from the llm module)

NathalieCharbel: [c1bb8c9](https://github.com/neo4j/neo4j-graphrag-python/pull/425/commits/c1bb8c9a19fd58c70ef465885a2c14145b8c555c)

PR #428, opened 2025-10-08
Fixes for the new LLMInterface

stellasia:
# Description

- Add deprecation warnings as discussed internally
- Remove abstractmethod for `_invoke` and `_ainvoke` methods, which would force users to implement these methods, even if they are already implementing the `invoke/ainvoke` method (ie would be a breaking change)

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/docs/source/user_guide_rag.rst
+++ b/docs/source/user_guide_rag.rst
@@ -265,21 +265,17 @@ Here's an example using the Python Ollama client:
 
     import ollama
     from neo4j_graphrag.llm import LLMInterface, LLMResponse
+    from neo4j_graphrag.types import LLMMessage
 
     class OllamaLLM(LLMInterface):
 
-        def invoke(self, input: str) -> LLMResponse:
-            response = ollama.chat(model=self.model_name, messages=[
-              {
-                'role': 'user',
-                'content': input,
-              },
-            ])
+        def _invoke(self, input: list[LLMMessage]) -> LLMResponse:
+            response = ollama.chat(model=self.model_name, messages=input)
             return LLMResponse(
                 content=response["message"]["content"]
             )
 
-        async def ainvoke(self, input: str) -> LLMResponse:
+        async def _ainvoke(self, input: list[LLMMessage]) -> LLMResponse:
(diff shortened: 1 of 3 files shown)


docs/source/user_guide_rag.rst
```diff
@@ -265,21 +269,17 @@ Here's an example using the Python Ollama client:
...
     class OllamaLLM(LLMInterface):
 
-        def invoke(self, input: str) -> LLMResponse:
-            response = ollama.chat(model=self.model_name, messages=[
-              {
-                'role': 'user',
-                'content': input,
-              },
-            ])
+        def _invoke(self, input: list[LLMMessage]) -> LLMResponse:
```
NathalieCharbel, line 276: I am not sure I get it. Before removing abstractmethod and inheritance from abc, the user couldn't also skip the invoke method and only implement the _invoke() method which was the abstract method? so what's the difference?

src/neo4j_graphrag/embeddings/base.py
```diff
@@ -24,7 +23,7 @@
 )
 
 
-class Embedder(ABC):
+class Embedder:
```
NathalieCharbel, line 26: Do we care about things lost when we remove inheritance from ABC? less safety and early error handling?

docs/source/user_guide_rag.rst
stellasia: With the `abstractmethod` decorator, users are forced to implement the `_invoke` method, and this is breaking change for users who may have custom interfaces.

NathalieCharbel: I see 👍🏻

src/neo4j_graphrag/embeddings/base.py
stellasia: Well, the only other solution I can think of is to create a `LLMInterfaceV2`. If you prefer this option, I'm fine with it.

Issue #383, opened 2025-07-14
langchain is not available


heishen6:
code:

```
import neo4j
from langchain_openai.chat_models import ChatOpenAI
from langchain_openai.embeddings import OpenAIEmbeddings
from neo4j_graphrag.generation import GraphRAG
from neo4j_graphrag.retrievers import VectorCypherRetriever
import os



# Define database credentials
URI  = "neo4j://x.x.x.x:19009"
AUTH = ("neo4j", "xxxx")
DATABASE = "neo4j"
INDEX = "xxx_embeddings"


driver = neo4j.GraphDatabase.driver(
    URI,
    auth=AUTH,
)

embedder = OpenAIEmbeddings(model="text-embedding-ada-002")

retriever = VectorCypherRetriever(
    driver,
    index_name=INDEX,
    retrieval_query="WITH node, score RETURN node.title as title, node.plot as plot",
    embedder=embedder,  # type: ignore[arg-type, unused-ignore]
    neo4j_database=DATABASE,
)

llm = ChatOpenAI(model="gpt-4o", temperature=0)

rag = GraphRAG(
    retriever=retriever,
    llm=llm,  # type: ignore[arg-type, unused-ignore]
)

result = rag.search(
    "have movie?",
    return_context=False,
)
print(result.answer)

driver.close()
```

```
Traceback (most recent call last):
  File "/Users/heishen/Desktop/work_Mac_mini/Developer/GraphRAG_/langchain 测试.py", line 49, in <module>
    result = rag.search(
             ^^^^^^^^^^^
  File "/opt/anaconda3/envs/test/lib/python3.11/site-packages/neo4j_graphrag/generation/graphrag.py", line 150, in search
    llm_response = self.llm.invoke(
                   ^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/test/lib/python3.11/site-packages/langchain_core/language_models/chat_models.py", line 378, in invoke
    self.generate_prompt(
  File "/opt/anaconda3/envs/test/lib/python3.11/site-packages/langchain_core/language_models/chat_models.py", line 963, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #442, opened 2025-10-30
LangChain Compability Restored

serengil:
# Description

The additional arguments we previously introduced to the invoke function in LLMInterface broke compatibility with LangChain. This change introduces a new `LLMInterfaceV2` contract that is fully compatible with LangChain while maintaining backward compatibility. If any users have implemented the old LLMInterface manually, their implementations will continue to work without modification.

Resolves https://github.com/neo4j/neo4j-graphrag-python/issues/383

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Moderate. Even though a new contract provided, the old contact is still valid to offer backward compability.

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -104,24 +113,30 @@ def search(
 
         Args:
             query_text (str): The user question.
-            message_history (Optional[Union[List[LLMMessage], MessageHistory]]): A collection previous messages,
-                with each message having a specific role assigned.
+            message_history (Optional[Union[List[LLMMessage], MessageHistory]]): A collection
+                previous messages, with each message having a specific role assigned.
```
stellasia:
```suggestion
                of previous messages, with each message having a specific role assigned.
```
?

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -104,24 +113,30 @@ def search(
...
         """
         if return_context is None:
-            warnings.warn(
-                "The default value of 'return_context' will change from 'False' to 'True' in a future version.",
-                DeprecationWarning,
-            )
-            return_context = False
+            if isinstance(self.llm, LLMInterface):
+                warnings.warn(
+                    "The default value of 'return_context' will change from 'False'"
```
stellasia: These changes seem to be related to line length only, are there made on purpose? (asking because I know Nathalie had some issues about this in the past)

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -145,13 +160,32 @@ def search(
...
+                    prompt=prompt,
+                    message_history=message_history,
+                    system_instruction=self.prompt_template.system_instructions,
+                )
+
+                # langchain chat model compatible invoke
+                llm_response = self.llm.invoke(
+                    input=messages,
+                )
+            elif isinstance(self.llm, LLMInterface):
```
stellasia, line 177: What about:
```
if isinstance(llm, LLMInterface):
   # ... call to invoke with input, message_history...
else:
   # we assume the interface is compatible with our LLMInterfaceV2
```
?

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -163,15 +197,33 @@ def _build_query(
...
+            "You are a summarization assistant. "
+            "Summarize the given text in no more than 300 words."
+        )
         if message_history:
             summarization_prompt = self._chat_summary_prompt(
                 message_history=message_history
             )
-            summary = self.llm.invoke(
-                input=summarization_prompt,
+            messages = legacy_inputs_to_messages(
```
stellasia: Nitpicking: I'd move this line inside the `if` block if it's not used elsewhere.

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -82,25 +84,67 @@ def __init__(
...
+            return await self.__legacy_ainvoke(
+                input, message_history, system_instruction
+            )
+        elif isinstance(input, list):
+            return await self.__brand_new_ainvoke(input)
+        else:
+            raise ValueError(f"Invalid input type for ainvoke method - {type(input)}")
+
+    # implementaions
+    @rate_limit_handler
```
stellasia: Don't we need to keep this decorator on the `invoke` method (or all of the `overload`?) so that it's applied for all calls?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -82,25 +84,67 @@ def __init__(
...
+    ) -> LLMResponse: ...
 
-    @rate_limit_handler
+    @overload
     def invoke(
+        self,
+        input: List[LLMMessage],
+    ) -> LLMResponse: ...
+
+    @overload  # type: ignore[no-overload-impl]
```
stellasia, line 110: Do you know why we need these two type ignore comments: `# type: ignore[no-overload-impl]` and later `# type: ignore[no-redef]`?

stellasia (commented):
Nice work, I haven't read everything yet, but left a few comments/questions.

src/neo4j_graphrag/generation/graphrag.py
serengil: Our LLM models (e.g. src/neo4j_graphrag/llm/openai_llm.py) inherit from LLMInterface and LLMInterfaceV2. In other words, when we check its type, then both isinstance(llm, LLMInterface) and isinstance(llm, LLMInterfaceV2) will return true. If we check that it is type of LLMInterface first, then we will run langchain incompatible branch all the time. That's why, I checked that it is type of LLMInterfaceV2 firstly.

src/neo4j_graphrag/generation/graphrag.py
serengil: True, pylint was warning about that line is too long. Autolinter in my ide splitted this line to many lines. I can revert them back if you think it may cause any issue.

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: seems like a mypy bug. alternatively we can move overload definitions to same file with pyi extension but it may make repo complexer.

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: You're right, we should keep this in the actual implementation of invoke function, but not in overload's level because overload definitions are used purely for type checking and are ignored at runtime.

src/neo4j_graphrag/generation/graphrag.py
stellasia: It's just that the next person pushing to the repo will have to make the opposite change, so we should agree on a convention. Is this something we should add to a config file in the repo?

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -163,18 +195,40 @@ def _build_query(
...
+                ).content
+            else:
+                raise ValueError(f"Type {type(self.llm)} of LLM is not supported.")
+
             return self.conversation_prompt(summary=summary, current_query=query_text)
         return query_text
 
+    def is_langchain_compatible(self) -> bool:
+        """Checks if the LLM is compatible with LangChain."""
+        return isinstance(self.llm, LLMInterfaceV2) or self.llm.__module__.startswith(
```
serengil: I should do something prettier than string check here.

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -136,8 +204,31 @@ def invoke(
         except self.anthropic.APIError as e:
             raise LLMGenerationError(e)
 
-    @async_rate_limit_handler
-    async def ainvoke(
+    def __brand_new_invoke(
```
NathalieCharbel: should we keep rate limit handler decorator for all `__brand_new_invoke` so we also ensure rate limit handling for the LLMInterfaceV2 instances?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -35,15 +35,20 @@
...
     """Interface for large language models on Anthropic
 
     Args:
         model_name (str, optional): Name of the LLM to use. Defaults to "gemini-1.5-flash-001".
-        model_params (Optional[dict], optional): Additional parameters passed to the model when text is sent to it. Defaults to None.
+        model_params (Optional[dict], optional): Additional parameters for LLMInterface(V1) passed to the model when text is sent to it. Defaults to None.
         system_instruction: Optional[str], optional): Additional instructions for setting the behavior and context for the model in a conversation. Defaults to None.
+        rate_limit_handler (Optional[RateLimitHandler], optional): Handler for managing rate limits for LLMInterface(V1). Defaults to None.
+        model_kwargs (Optional[dict], optional): Additional parameters for LLMInterfaceV2 passed to the model when text is sent to it. Defaults to None.
+        rate_limiter (Optional[RateLimitHandler], optional): Handler for managing rate limits for LLMInterfaceV2. Defaults to None.
```
NathalieCharbel: should we also add deprecation warnings for all v1 params (`rate_limit_handler`, `model_params`) and ask users to start using the new params?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -77,30 +84,91 @@ def __init__(
                 """Could not import Anthropic Python client.
                 Please install it with `pip install "neo4j-graphrag[anthropic]"`."""
             )
-        super().__init__(model_name, model_params, rate_limit_handler)
+        if isinstance(self, LLMInterfaceV2):
```
stellasia: Isn't this condition always true?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -77,30 +84,91 @@ def __init__(
...
+
+    # switching logics to LLMInterface or LLMInterfaceV2
+    def invoke(  # type: ignore[no-redef]
+        self,
+        input: Union[str, List[LLMMessage]],
+        message_history: Optional[Union[List[LLMMessage], MessageHistory]] = None,
+        system_instruction: Optional[str] = None,
+        **kwargs: Any,
+    ) -> LLMResponse:
+        if isinstance(input, str):
```
stellasia, line 133: `input` can be a string in LC too (https://reference.langchain.com/python/integrations/langchain_openai/ChatOpenAI/#langchain_core.language_models.BaseChatModel.astream_events). Maybe we can use:

```
if message_history or system_instruction:
   ...
```

instead?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -172,3 +263,72 @@ async def ainvoke(
             return LLMResponse(content=text)
         except self.anthropic.APIError as e:
             raise LLMGenerationError(e)
+
+    async def __brand_new_ainvoke(
```
stellasia: Naming question: should we use `__ainvoke_v2` instead of `brand_new`? (same for other occurrences of `brand_new`)

src/neo4j_graphrag/llm/base.py
```diff
@@ -152,3 +154,109 @@ async def ainvoke_with_tools(
...
+    def invoke(
+        self,
+        input: List[LLMMessage],
+        **kwargs: Any,
+    ) -> LLMResponse:
+        """Sends a text input to the LLM and retrieves a response.
+
+        Args:
+            input (str): Text sent to the LLM.
+            message_history (Optional[Union[List[LLMMessage], MessageHistory]]): A collection previous messages,
```
stellasia: Remember to update docstrings ;)

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: if smo creates their own class inherif from LLMInterface, then it's false.

For users using our schemas, both V2 and V1 are true, that's why we added isinstance(self, LLMInterfaceV2) check first.

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: yeah good point, if we have v3 in the future, then they won't be brand new :)

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: in src/neo4j_graphrag/generation/graphrag.py, we are converting it to List[LLMMessage] if it's langchain object.

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: Added a warning while initializing LLMInterface(V1) - https://github.com/neo4j/neo4j-graphrag-python/pull/442/commits/4728be93dfc3113111ade3e5e6ee19ad50024bec

src/neo4j_graphrag/llm/anthropic_llm.py
stellasia: Here it is inside our Anttorpic implementation, that inherits from both V1 and V2? Or am I missing something?

src/neo4j_graphrag/llm/anthropic_llm.py
stellasia: Yes, we only support `list[LLMMessage]` in our interface for now , which is compatible with LC, and in our code we'll force ourselves to always pass the LLM input as such. My point here is just that we could support simple string input in the future (which is also still LC compatible), and then this condition would fail.

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -104,24 +113,30 @@ def search(
...
             RagResultModel: The LLM-generated answer.
 
         """
         if return_context is None:
-            warnings.warn(
-                "The default value of 'return_context' will change from 'False' to 'True' in a future version.",
-                DeprecationWarning,
-            )
-            return_context = False
+            if self.is_langchain_compatible():
```
stellasia: I missed that on first review, why changing this behavior here?

src/neo4j_graphrag/generation/graphrag.py
```diff
@@ -163,18 +195,47 @@ def _build_query(
...
+                    input=summarization_prompt,
+                    system_instruction=summary_system_message,
+                ).content
+            else:
+                raise ValueError(f"Type {type(self.llm)} of LLM is not supported.")
+
             return self.conversation_prompt(summary=summary, current_query=query_text)
         return query_text
 
+    def is_langchain_compatible(self) -> bool:
```
stellasia, line 225: Note for later: we need to call the LLM in other places in this package (e.g. in Text2CypherRetriever or components from SimpleKGPIpeline). We need to find a common place for this function and conditionned call.

src/neo4j_graphrag/llm/base.py
```diff
@@ -152,3 +157,105 @@ async def ainvoke_with_tools(
...
+
+        Raises:
+            LLMGenerationError: If anything goes wrong.
+        """
+        raise NotImplementedError("This LLM provider does not support tool calling.")
+
+    async def ainvoke_with_tools(
+        self,
+        input: list[LLMMessage],
+        tools: Sequence[Tool],  # Tools definition as a sequence of Tool objects
```
stellasia: We can address this in a separate PR, but have we checked that our tools are LC compatible?

stellasia: Before we release v2, we should also drop these `invoke_with_tools` methods.

serengil: dropped (a)invoke_with_tools for v2

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: now, creating LLMInterfaceV2 by default in llm implementations

src/neo4j_graphrag/generation/graphrag.py
serengil: it was saying that `The default value of 'return_context' will change from 'False'  to 'True' in a future version.`. So, I make this true in the new release. I can revert this we want to wait.

src/neo4j_graphrag/llm/anthropic_llm.py
serengil: i see what you mean. here, message_history and system_instruction are optional.

agreed to keep this as is.

src/neo4j_graphrag/generation/graphrag.py
serengil: Reverted the behaviour back for return_context

src/neo4j_graphrag/generation/graphrag.py
serengil: resolved this because it is a note for later

PR #488, opened 2026-03-13
Add Cypher 25 SEARCH clause for vector retrievers

oskarhane:
## Summary
- Auto-detect Neo4j 2026.01+ and route vector/hybrid queries through the new Cypher 25 `SEARCH ... WHERE` clause for in-index filtered ANN search
- Classify filters as SEARCH-compatible (simple AND comparisons) vs incompatible (`$or`, `$in`, `$like`) with automatic fallback to procedure path + warning log
- Gate SEARCH path on the index's `filterable_properties`: fetch them in `_fetch_index_infos` and fall back to the procedure path (with a warning) when any filtered property isn't declared filterable
- Apply `effective_search_ratio` inside the SEARCH clause (`LIMIT $top_k * $effective_search_ratio` then outer `LIMIT $top_k`) for parity with the procedure path
- Add `filterable_properties` param to `create_vector_index()` for `WITH [n.prop]` clause
- All 4 retrievers supported: VectorRetriever, VectorCypherRetriever, HybridRetriever, HybridCypherRetriever

Closes #485
Closes GENKGB-1243

## Test plan
- [x] Unit tests pass
- [x] 6 E2E tests pass against Neo4j 2026.02.2-enterprise in Docker
- [x] mypy strict, ruff lint/format clean
- [x] CI passes


docs/source/user_guide_rag.rst
```diff
@@ -716,6 +720,44 @@ The currently supported operators are:
 - `$like`: LIKE operator case-sensitive.
 - `$ilike`: LIKE operator case-insensitive.
 
+.. _search-clause-filtering:
+
+In-Index Filtering with the SEARCH Clause (Neo4j 2026.01+)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
stellasia: Total detail, but I think we're using `=` or `-` separator in the rest of the file, is `^` for another level?

src/neo4j_graphrag/retrievers/hybrid.py
```diff
@@ -120,6 +126,7 @@ def __init__(
             else None
         )
         self.result_formatter = validated_data.result_formatter
+        self._node_label = None
```
stellasia, line 129: Not clear to me where this value is actually filled?

docs/source/user_guide_rag.rst
```diff
@@ -716,6 +720,44 @@ The currently supported operators are:
...
+
+On Neo4j 2026.01 and later, the library automatically detects the server version and
+uses the Cypher ``SEARCH`` clause for vector queries. This enables in-index filtering,
+where compatible filter predicates are evaluated inside the vector index itself rather
+than via post-hoc brute-force scanning.
+
+**Requirements:**
+
+1. Neo4j server version 2026.01 or later.
+2. The vector index must be created with ``filterable_properties`` (see :ref:`filterable-index-creation`).
```
stellasia: Is it something we could check in `supports_search_clause` using the index name?

src/neo4j_graphrag/neo4j_queries.py
```diff
@@ -289,6 +292,139 @@ def _get_filtered_vector_query(
...
+        where_part = f" WHERE {filter_classification.cypher_where_clause}"
+        if filter_classification.params:
+            params.update(filter_classification.params)
+
+    query = (
+        f"MATCH (node:`{node_label}`) "
+        f"SEARCH node IN (VECTOR INDEX `{index_name}` "
+        f"FOR $query_vector"
+        f"{where_part} "
+        f"LIMIT $top_k) "
```
stellasia: No need to use the effective search ratio here?

tests/e2e/docker-compose.neo4j2026.yml
```diff
@@ -0,0 +1,12 @@
+---
```
stellasia, line 1: How is this file used with respect to the other docker compose file?

docs/source/user_guide_rag.rst
oskarhane: Good catch — switched to ~~~ in 175aa746.

src/neo4j_graphrag/retrievers/hybrid.py
oskarhane: Set by _fetch_index_infos() (base.py) called 3 lines below — same pattern as _embedding_node_property/_embedding_dimension above. The None is the pre-fetch default.

docs/source/user_guide_rag.rst
oskarhane: Done in b0444979. Fetch filterable_properties in _fetch_index_infos and gate SEARCH path on it: if any filtered field is missing from the index's filterable_properties, fall back to procedure path with a warning. Doc requirement updated to reflect that filterable_properties is only needed for filtered queries.

src/neo4j_graphrag/neo4j_queries.py
oskarhane: Fixed in 8699a76b. Inner SEARCH LIMIT now $top_k * $effective_search_ratio with an outer WITH ... LIMIT $top_k, mirroring the procedure path.

tests/e2e/docker-compose.neo4j2026.yml
oskarhane: Pins Neo4j 2026.02.2 (needed for the SEARCH clause); same 7687/7474 ports as the default file so they can't run together. Added a README note in e7f31059 clarifying when to use which.

stellasia (approved):
LGTM! Nice!

PR #489, opened 2026-03-17
feat: Add Cypher 25 SEARCH clause support for vector retrievers

ADunfield:
## Summary

Adds support for the Cypher 25 `SEARCH` clause with in-index filtering for vector retrievers, as proposed in #485.

When connected to Neo4j 2026.01+, compatible filtered vector queries now use the new `MATCH ... SEARCH node IN (VECTOR INDEX ... FOR ... WHERE ...)` syntax, pushing predicates directly into the index scan instead of brute-force exact KNN. This significantly improves performance for filtered vector search workloads.

### Changes

- **Version detection**: `has_search_clause_support()` detects Neo4j 2026.01+ (calendar versioning)
- **Filter compatibility**: `is_search_compatible_filter()` identifies filters that can be pushed into the SEARCH clause (simple comparisons with AND combinations only; `$or`, `$in`, `$nin`, `$like`, `$ilike`, `$ne`, `$between` fall back to exact KNN)
- **SEARCH query generation**: `_get_search_vector_query()` and `_get_search_vector_query_no_filter()` build the new Cypher 25 syntax
- **Automatic routing**: `get_search_query()` transparently routes through SEARCH when the version supports it and filters are compatible
- **Index creation**: `filter_properties` parameter on `create_vector_index()` generates the `WITH [n.prop1, n.prop2]` clause for declaring filterable properties
- **Retriever integration**: Base retriever detects SEARCH support at init; `VectorRetriever` and `VectorCypherRetriever` pass the flag through

### Design decisions

1. **Zero breaking changes**: `neo4j_version_supports_search_clause` defaults to `False` everywhere, preserving existing behavior for all current users
2. **Automatic fallback**: Incompatible filters silently fall back to exact KNN — no user action required
3. **No filter on unfiltered search**: When SEARCH is available and `node_label` is known, uses SEARCH clause without WHERE (still leverages ANN index efficiently)

### Example generated queries

**With filters (Neo4j 2026.01+):**
```cypher
MATCH (node:`Movie`)
SEARCH node IN (
  VECTOR INDEX $vector_index_name
  FOR $query_vector
  WHERE node.rating >= $param_0 AND node.year > $param_1
  LIMIT $top_k
) SCORE AS score
```

**Index creation with filterable properties:**
```cypher
CREATE VECTOR INDEX $name IF NOT EXISTS FOR (n:Movie) ON n.embedding
WITH [n.releaseDate, n.rating]
OPTIONS { indexConfig: { `vector.dimensions`: toInteger($dimensions), `vector.similarity_function`: $similarity_fn } }
```

## Test plan

- [x] 33 new unit tests covering version detection, filter compatibility, SEARCH query generation, and routing logic
- [x] All 777 existing unit tests continue to pass (20 pre-existing failures from optional dependencies unrelated to this change)
- [ ] Integration testing against Neo4j 2026.01 instance (requires access to pre-release or calendar-versioned Neo4j)

Closes #485

### DIFF
--- a/src/neo4j_graphrag/filters.py
+++ b/src/neo4j_graphrag/filters.py
@@ -367,3 +367,75 @@ def get_metadata_filter(
     param_store = ParameterStore()
     query = _construct_metadata_filter(filter, param_store, node_alias=node_alias)
     return query, param_store.params
+
+
+# Operators compatible with Cypher 25 SEARCH clause in-index filtering.
+# Only simple comparison operators (and AND combinations of them) can be
+# pushed down into the vector index scan.
+_SEARCH_COMPATIBLE_OPERATORS = {
+    OPERATOR_EQ,
+    OPERATOR_LT,
+    OPERATOR_LTE,
+    OPERATOR_GT,
+    OPERATOR_GTE,
+}
+
+
+def is_search_compatible_filter(filter: Any) -> bool:
+    """Check whether a filter dict can be evaluated inside a SEARCH clause.
+
+    The Cypher 25 SEARCH clause supports a limited subset of predicates
+    inside ``FOR ... WHERE``: simple comparisons (=, <, <=, >, >=) combined
+    with AND.  Anything else ($or, $in, $nin, $like, $ilike, $ne, $between)
+    requires the brute-force exact-KNN fallback.
+
+    Args:
+        filter: The filter dictionary to check.
+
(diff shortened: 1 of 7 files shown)


stellasia:
Similar work done in https://github.com/neo4j/neo4j-graphrag-python/pull/488

PR #498, opened 2026-03-30
Add close/aclose methods to LLMBase

stellasia:
# Description

This PR adds `close` and `aclose` methods to the `LLMBase` base class in order to avoid having to access the internal `async_client` to clean up ressources properly. 

In practice, only the `aclose` method needs to be implemented. The `close` method tries to call `aclose`. If an event loop is already running, it raises an error saying to call `await llm.aclose()` instead.

Also added the context management methods, even if not mandatory, it allows to use this pattern: 

```
with OpenAILLM(model_name="") as llm:
    pass
```
and clients will be closed gracefully.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

> **Note**
>
>
>

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/examples/customize/build_graph/components/schema_builders/schema_from_text.py
+++ b/examples/customize/build_graph/components/schema_builders/schema_from_text.py
@@ -61,7 +61,7 @@ async def extract_and_save_schema() -> None:
 
     # Define LLM parameters
     llm_model_params = {
-        "max_tokens": 2000,
+        "max_completion_tokens": 2000,
         "response_format": {"type": "json_object"},
         "temperature": 0,  # Lower temperature for more consistent output
     }
@@ -104,7 +104,7 @@ async def extract_and_save_schema() -> None:
 
     finally:
         # Close the LLM client
-        await llm.async_client.close()
+        await llm.aclose()
 
 
 async def main() -> None:
--- a/examples/customize/build_graph/components/schema_builders/schema_from_text.py
+++ b/examples/customize/build_graph/components/schema_builders/schema_from_text.py
@@ -104,7 +104,8 @@ async def extract_and_save_schema() -> None:
 
     finally:
         # Close the LLM client
-        await llm.aclose()
+        if hasattr(llm, "aclose"):
+            await llm.aclose()
 
 
 async def main() -> None:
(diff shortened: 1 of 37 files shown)


src/neo4j_graphrag/llm/cohere_llm.py
```diff
@@ -353,6 +353,9 @@ async def __ainvoke_v2(
             usage=usage,
         )
 
+    async def aclose(self) -> None:
+        pass
```
adamnsch: Do we not need do to anything here? (same question for ollama). If not we don't have to implement it since it's not an `@abstractmethod` right?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -299,6 +299,10 @@ async def __ainvoke_v2(
         except self.anthropic.APIError as e:
             raise LLMGenerationError(e)
 
+    async def aclose(self) -> None:
+        self.client.close()
+        await self.async_client.close()
```
adamnsch, line 304: Could this be a default implementation of aclose?

examples/customize/build_graph/components/schema_builders/schema_from_text.py
```diff
@@ -104,7 +104,8 @@ async def extract_and_save_schema() -> None:
 
     finally:
         # Close the LLM client
-        await llm.async_client.close()
+        if hasattr(llm, "aclose"):
```
adamnsch, line 107: Do we need these `hasattr` when `LLMBase` has `aclose`?

adamnsch (approved):
LGTM :) Just have some minor questions/comments

src/neo4j_graphrag/llm/anthropic_llm.py
stellasia: Well in practice all implementations look similar, but it's not guaranteed. And the base class does not have the clients attributes, so I'm a bit afraid pyright will complain about that.

examples/customize/build_graph/components/schema_builders/schema_from_text.py
stellasia: I added this check because we had a hard time making the `LLMInterfaceV2` compatible with langchain LLM chat classes, and they do not have this `close` method.

PR #500, opened 2026-03-30
fix: build explicit PyArrow schema in Neo4jGraphParquetFormatter

matteomedioli:
# Description

  `pa.Table.from_pylist(rows)` infers the Parquet schema from the first row it encounters. When the first node or relationship row has no `embedding_properties`, the embedding column is never created and all embedding values from subsequent rows are silently discarded. According to here: https://github.com/neo4j/neo4j-graphrag-python/blob/3d092f6078926bde38d233018528a8fb6fdc79d4/src/neo4j_graphrag/experimental/components/lexical_graph.py#L138 in case of null or empty embeddings, the default value of `embedding_properties` is empty dict `{}`.                                                                                                                              
   
  **Minimal reproduction code:**                                                                                                                                                                  
                                                            
  ```python
  import pyarrow.parquet as pq
  from io import BytesIO
  from neo4j_graphrag.experimental.components.parquet_formatter import Neo4jGraphParquetFormatter                                                                                            
  from neo4j_graphrag.experimental.components.types import LexicalGraphConfig, Neo4jGraph, Neo4jNode
                                                                                                                                                                                             
  formatter = Neo4jGraphParquetFormatter()                  
  config = LexicalGraphConfig()                                                                                                                                                              
                                                            
  # batch 1 FAILED → no embedding_properties → first in node list
  # batch 2 SUCCEEDED → embedding_properties present
  nodes = [
      Neo4jNode(id='c0', label='Chunk', properties={'text': 'a'}, embedding_properties={}),
      Neo4jNode(id='c1', label='Chunk', properties={'text': 'b'}, embedding_properties={}),                                                                                                  
      Neo4jNode(id='c2', label='Chunk', properties={'text': 'c'}, embedding_properties={'embedding': [0.1, 0.2, 0.3]}),                                                                      
      Neo4jNode(id='c3', label='Chunk', properties={'text': 'd'}, embedding_properties={'embedding': [0.4, 0.5, 0.6]}),                                                                      
  ]                                                                                                                                                                                          
  graph = Neo4jGraph(nodes=nodes, relationships=[])         
  data, _, _ = formatter.format_graph(graph, config)                                                                                                                                         
  table = pq.read_table(BytesIO(list(data['nodes'].values())[0]))                                                                                                                            
  print(table.schema.names)  # ['__id__', 'labels', 'text'] — 'embedding' silently dropped
  print(table.to_pylist())   # c2 and c3 have NO embedding — successful embeddings lost                                                                                                      
```
                                                                                                                                                                                          
The fix builds an explicit `pa.schema` from the union of all keys across every row before calling `from_pylist`. Missing values  are filled with null, and float-list columns (e.g. embeddings) are normalised to `list<float32>` (default type for python float list) regardless of which row provides the first non-null sample.                                                                                                                                                                 
   
## Type of Change                                                                                                                                                                         
  - Bug fix                                                                                                                                                                             
   
## Complexity: 
- Low                                                                                                                                                                            
                                                            
## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/components/parquet_formatter.py
+++ b/src/neo4j_graphrag/experimental/components/parquet_formatter.py
@@ -504,7 +504,7 @@ class Neo4jGraphParquetFormatter:
                 if k in sample:
                     t: Any = pa.infer_type([sample[k]])
                     if pa.types.is_list(t) and pa.types.is_floating(t.value_type):
-                        t = pa.list_(pa.float64())
+                        t = pa.list_(pa.float32())
                 elif k in has_empty_list:
                     # Only empty lists seen — use list<null> so [] values round-trip correctly.
                     t = pa.list_(pa.null())
--- a/src/neo4j_graphrag/experimental/components/parquet_formatter.py
+++ b/src/neo4j_graphrag/experimental/components/parquet_formatter.py
@@ -486,11 +486,17 @@ class Neo4jGraphParquetFormatter:
             # silent column drops when the first row lacks some keys (e.g. embeddings).
             # Dict preserves first-seen insertion order for deterministic column ordering.
             all_keys: dict[str, None] = {k: None for row in rows for k in row}
-            # First non-null value for each key, used to infer the column type.
+            # Collect the first non-null, non-empty-list value per key for type inference.
+            # Also track keys that carry at least one empty list, so we can fall back to
+            # list<null> when no richer sample exists (rather than pa.null(), which cannot
+            # hold [] values).
             sample: dict[str, Any] = {}
+            has_empty_list: set[str] = set()
             for row in rows:
                 for k, v in row.items():
-                    if k not in sample and v is not None:
+                    if v == []:
+                        has_empty_list.add(k)
+                    elif k not in sample and v is not None:
                         sample[k] = v
 
             fields: list[Any] = []
@@ -498,7 +504,10 @@ class Neo4jGraphParquetFormatter:
                 if k in sample:
                     t: Any = pa.infer_type([sample[k]])
                     if pa.types.is_list(t) and pa.types.is_floating(t.value_type):
-                        t = pa.list_(pa.float32())
+                        t = pa.list_(pa.float64())
+                elif k in has_empty_list:
+                    # Only empty lists seen — use list<null> so [] values round-trip correctly.
+                    t = pa.list_(pa.null())
                 else:
                     t = pa.null()
                 fields.append(pa.field(k, t))
--- a/src/neo4j_graphrag/experimental/components/parquet_formatter.py
+++ b/src/neo4j_graphrag/experimental/components/parquet_formatter.py
@@ -481,13 +481,38 @@ class Neo4jGraphParquetFormatter:
             import pyarrow.parquet as pq
 
             self._normalize_column_types(rows)
-            table = pa.Table.from_pylist(rows)
-            # Write to BytesIO buffer
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/neo4j_graphrag/experimental/components/parquet_formatter.py [hunk 1]
```diff
@@ -481,13 +481,47 @@ def format_parquet(
...
+                    if v == []:
+                        has_empty_list.add(k)
+                    elif k not in sample and v is not None:
+                        sample[k] = v
+
+            fields: list[Any] = []
+            for k in all_keys:
+                if k in sample:
+                    t: Any = pa.infer_type([sample[k]])
+                    if pa.types.is_list(t) and pa.types.is_floating(t.value_type):
```
ali-sedaghatbaf, line 506: What if embeddings happen to be all-integer? they'll be inferred as `list<int64>` I guess and stay that way.

src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -481,13 +481,47 @@ def format_parquet(
...
+                        t = pa.list_(pa.float32())
+                elif k in has_empty_list:
+                    # Only empty lists seen — use list<null> so [] values round-trip correctly.
+                    t = pa.list_(pa.null())
```
ali-sedaghatbaf, line 510: I think it's worth documenting this `list<null>` as a limitation since consumers might use DockDB, Spark, etct which are not happy with this.

src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -481,13 +481,47 @@ def format_parquet(
...
+            sample: dict[str, Any] = {}
+            has_empty_list: set[str] = set()
+            for row in rows:
+                for k, v in row.items():
+                    if v == []:
+                        has_empty_list.add(k)
+                    elif k not in sample and v is not None:
+                        sample[k] = v
+
+            fields: list[Any] = []
```
ali-sedaghatbaf, line 502: `list[pa.Field] `?

tests/unit/experimental/components/test_kg_writer.py
```diff
@@ -767,3 +768,286 @@ async def test_parquet_writer_mixed_property_types() -> None:
...
+    assert (
+        "embedding" in schema.names
+    ), "Embedding column should be present even when all rows have an empty list"
+
+    table = pq.read_table(BytesIO(parquet_bytes))
+    assert table.num_rows == 2
+    assert "embedding" in table.column_names
+    for row in table.to_pylist():
+        assert (
+            row["embedding"] == [] or row["embedding"] is None
```
ali-sedaghatbaf, line 1006: Shouldn't we focus only on empty lists as the name of the test implies?

ali-sedaghatbaf (approved):
Nice work!

Approved with non-blocking comments.

src/neo4j_graphrag/experimental/components/parquet_formatter.py
matteomedioli: Sure, thanks!

src/neo4j_graphrag/experimental/components/parquet_formatter.py
matteomedioli: Great catch! While embeddings shouldn't be integers by design, this is a solid guardrail in case that assumption changes or becomes invalid in the future.

PR #510, opened 2026-04-13
Introduce Neo4j-style EXISTENCE constraints for KG schemas

adamnsch:
# Description

## Summary

Aligns the KG schema model with how **Neo4j** separates concerns: **`required` (existence)** and **UNIQUENESS** are different constraint concepts and should not be represented the same way on node/relationship properties. This change makes **`GraphSchema`** express that distinction explicitly—**`EXISTENCE`** constraints for mandatory properties, **`UNIQUENESS`** for unique keys—and deprecates **`PropertyType.required`** with migration on load.

Structured schema extraction is also updated so **Vertex AI** can consume the response JSON Schema (wire-only constraint types without nullable fields that break protobuf conversion).

## Why

Previously, **`required`** lived on individual properties while **UNIQUENESS** lived in **`constraints`**, and the two were easy to conflate in prompts, extraction, and pruning. Neo4j treats **property existence** and **uniqueness** as separate constraint kinds; mirroring that in the library removes ambiguity and avoids implying that uniqueness enforces presence (or vice versa).

### Runtime schema (`GraphSchema` / `ConstraintType`)

- Introduces **`GraphConstraintType`**: **`UNIQUENESS`** (node properties only in this API) and **`EXISTENCE`** (node **or** relationship property).
- **`PropertyType.required`** is deprecated; loading a schema migrates legacy `required` flags into **`EXISTENCE`** constraints where appropriate.
- **Graph pruning**, **schema visualization**, and **schema-from-existing-graph** flows updated to use **EXISTENCE** (and no longer treat uniqueness as a stand-in for required).

### Schema-from-text structured output (OpenAI / Vertex)

- **`GraphSchemaExtractionOutput.constraints`** use **`ExtractedConstraintType`**: plain string fields and **`relationship_type: ""`** instead of **`Optional` / null**, so generated JSON Schema stays compatible with Vertex’s protobuf parser.
- **`wire_extraction_constraints_for_graph_schema`** maps wire values (`""` / legacy `null`) to **`None`** for runtime **`ConstraintType`**.
- **`SchemaExtractionTemplate`** examples updated to use **`""`** instead of **`null`** where a constraint is not relationship-scoped.

### Tests & docs

- Unit/contract tests for extraction, constraints, pruning, and deprecation; **CHANGELOG** (`## Next`) updated.

## Type of Change
- [x] New feature (more like new syntax)
- [ ] Bug fix
- [x] Breaking change (but not really since we're only deprecating)
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/components/schema.py
+++ b/src/neo4j_graphrag/experimental/components/schema.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 from __future__ import annotations
 
+import enum
 import json
 import logging
 import re
@@ -93,6 +94,17 @@ Neo4jPropertyTypeName: TypeAlias = Literal[
 _DUNDER_RE = re.compile(r"^__|__$")
 
 
+class GraphConstraintType(str, enum.Enum):
+    """Constraint kinds for :class:`ConstraintType`.
+
+    ``UNIQUENESS`` is for node properties only in this API. ``EXISTENCE`` marks a mandatory
+    (non-null) node or relationship property, analogous to Neo4j property existence constraints.
+    """
+
+    UNIQUENESS = "UNIQUENESS"
+    EXISTENCE = "EXISTENCE"
+
+
 def _reject_dunder_label(label: str, kind: str) -> str:
     """Raise ValueError if *label* starts or ends with double underscores."""
     if _DUNDER_RE.search(label):
@@ -112,7 +124,13 @@ class PropertyType(BaseModel):
     # See https://neo4j.com/docs/cypher-manual/current/values-and-types/property-structural-constructed/#property-types
     type: Neo4jPropertyTypeName
     description: str = ""
-    required: bool = False
+    required: bool = Field(
+        default=False,
+        deprecated=(
+            "Use GraphSchema.constraints with type EXISTENCE instead of PropertyType.required. "
+            "Uniqueness does not imply existence; model existence explicitly with EXISTENCE constraints."
+        ),
+    )
     model_config = ConfigDict(
         frozen=True,
     )
@@ -237,18 +255,42 @@ class RelationshipType(BaseModel):
 
 class ConstraintType(BaseModel):
     """
-    Represents a constraint on a node in the graph.
+    Represents a schema-level constraint (uniqueness or existence) on a node or
+    (for EXISTENCE only) relationship property.
     """
 
-    type: Literal[
-        "UNIQUENESS"
-    ]  # TODO: add other constraint types ["propertyExistence", "propertyType", "key"]
-    node_type: str
+    type: GraphConstraintType
     property_name: str
+    node_type: str = ""
+    relationship_type: Optional[str] = None
 
     model_config = ConfigDict(
         frozen=True,
+        use_enum_values=True,
     )
+
+    @model_validator(mode="after")
+    def validate_constraint_shape(self) -> Self:
+        if self.type == GraphConstraintType.UNIQUENESS:
+            if not (self.node_type and self.node_type.strip()):
+                raise ValueError(
+                    "UNIQUENESS constraint requires a non-empty node_type; "
+                    "relationship uniqueness is not supported on GraphSchema constraints."
+                )
+            if self.relationship_type is not None and self.relationship_type.strip():
+                raise ValueError(
+                    "UNIQUENESS constraint must not set relationship_type; "
+                    "only node-level UNIQUENESS is supported."
+                )
+        elif self.type == GraphConstraintType.EXISTENCE:
+            has_node = bool(self.node_type and self.node_type.strip())
+            has_rel = bool(self.relationship_type and self.relationship_type.strip())
+            if has_node == has_rel:
+                raise ValueError(
+                    "EXISTENCE constraint requires exactly one of node_type or relationship_type "
+                    "(non-empty), not both and not neither."
+                )
+        return self
 
 
 class Pattern(BaseModel):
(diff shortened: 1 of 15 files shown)


src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -1268,52 +1484,63 @@ def __init__(
...
+                    continue
+                seen.add(dedupe_key)
+                result.append(
+                    {
+                        "type": GraphConstraintType.EXISTENCE.value,
+                        "node_type": "",
+                        "property_name": prop,
+                        "relationship_type": lab,
+                    }
+                )
```
NathalieCharbel, line 1553: In cypher, a node key constraint is the intersection of existence and uniqueness (it requires the property to exist AND be unique). When `SchemaFromExistingGraphExtractor` encounters a `NODE_KEY`, it only emits an `EXISTENCE` constraint. Should it also emit a `UNIQUENESS` constraint for the property Similarly for `RELATIONSHIP_KEY` but since the API currently only supports node-level `UNIQUENESS`, at minimum should we emit a `UNIQUENESS` for `NODE_KEY`

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -344,6 +386,95 @@ def convert_tuple_patterns(cls, data: Any) -> Any:
             data["patterns"] = converted
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def migrate_deprecated_required_to_existence_constraints(cls, data: Any) -> Any:
+        """Convert legacy ``PropertyType.required`` to ``EXISTENCE`` constraints and clear flags."""
+        if not isinstance(data, dict):
+            return data
```
NathalieCharbel, line 394: for existing users, code that does the programmatic-construction path with `required=True`
e.g.
```
nt = NodeType(
    label="Person",
    properties=[PropertyType(name="name", type="STRING", required=True)]
)
schema = GraphSchema(node_types=(nt,))
```
where `nt` is a `NodeType` instance, not a dict. The `isinstance(node, dict)` check is False, the `required=True` flag will never be inspected and hence No EXISTENCE constraint will be created?

src/neo4j_graphrag/generation/prompts.py
```diff
@@ -218,18 +218,18 @@ class SchemaExtractionTemplate(PromptTemplate):
...
 8.2 Only use properties that seem to not have too many missing values in the sample.
 8.3 Constraints reference node_types by label and specify which property is unique.
 8.4 If a property appears in a uniqueness constraint it MUST also appear in the corresponding node_type as a property.
-9. REQUIRED PROPERTIES:
-9.1 Mark a property as "required": true if every instance of that node/relationship type MUST have this property (non-nullable).
-9.2 Mark a property as "required": false if the property is optional and may be absent on some instances.
-9.3 Properties that are identifiers, names, or essential characteristics are typically required.
-9.4 Properties that are supplementary information (phone numbers, descriptions, metadata) are typically optional.
-9.5 When uncertain, default to "required": false.
-9.6 If a property has a UNIQUENESS constraint, it MUST be marked as "required": true.
```
NathalieCharbel, line 232: since we're removing the `required` from the prompt, should we also clean up all the post processing filtering operations involving it, eg `_filter_properties_required_field()` in `schema.py`?

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -93,6 +94,17 @@
 _DUNDER_RE = re.compile(r"^__|__$")
 
 
+class GraphConstraintType(str, enum.Enum):
+    """Constraint kinds for :class:`ConstraintType`.
+
+    ``UNIQUENESS`` is for node properties only in this API. ``EXISTENCE`` marks a mandatory
```
NathalieCharbel, line 100: Since `ConstraintType` already has `relationship_type` and the `EXISTENCE` validation/filtering handles both node and relationship scopes, would it be worth extending `UNIQUENESS` to also support relationship properties at this stage?

NathalieCharbel (commented):
Nice work! It's good to see the semantics of constraints being handled properly now.

My main concern/questions:
- previously, the "required" flag and UNIQUENESS constraints were effectively treated together as a KEY constraint. Given that the old behaviour already modelled the combination, wouldn't introducing a KEY constraint type have been the more natural evolution?  Having said that, two separate constraints is arguably more flexible.
- we previously handled UNIQUENESS for node types only, with these changes in this PR would it be also feasible to add them for relationship types at this stage? instead of adding additional guards?
- should graph pruning enforce UNIQUENESS? previously the pruner enforced required (now EXISTENCE) on extracted properties. With UNIQUENESS now properly modelled, should we also handle the case where the LLM extracts duplicate values for a property marked as unique? This is of course best enforced at the database level, but since we already prune for one constraint type, could it feel inconsistent to users that the other constraint type is silently ignored during pruning?

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: Very good observation! I agree that this is not actually correct, but we're not changing anything in terms of the logic. Before NODE_KEY => required flag, and now NODE_KEY => EXISTENCE constraint. There was (and still isn't) UNIQUENESS constraint on this code path, though there probably should be. The idea is to add key constraints in the next PR though, and then that would be used for NODE_KEY instead. Should we wait for that PR to make a change, or do you want me to add a UNIQUENESS constraint here as well for now?

src/neo4j_graphrag/generation/prompts.py
adamnsch: Yes, I think you're right. Good catch! I will remove it

adamnsch: Done :)

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: Good catch!

src/neo4j_graphrag/experimental/components/schema.py
NathalieCharbel: yea I see np. It doesn't matter if it is in this PR or the next one. I just wanted to make sure we do it correctly in the end 👍🏻

adamnsch: Ok, makes sense! I think I'll leave for the next one then to keep this as simple as possible

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: This is addressed now :)

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: Yes, I think this is a good idea! But I don't want to make this PR bigger, so created triage card for it: https://linear.app/neo4j/issue/7e1f5f95-2cec-4bfb-93b5-576132463f5f. Hope that makes sense

adamnsch:
Thanks for a great review! 🙏 

> * previously, the "required" flag and UNIQUENESS constraints were effectively treated together as a KEY constraint. Given that the old behaviour already modelled the combination, wouldn't introducing a KEY constraint type have been the more natural evolution?  Having said that, two separate constraints is arguably more flexible.

As discussed in the comments, the next PR will introduce KEY constraints. This PR was kind of the set up for that, that also remove the `required` syntax to keep or constraints syntax consistent. So it will be addressed in the next PR :)

> * we previously handled UNIQUENESS for node types only, with these changes in this PR would it be also feasible to add them for relationship types at this stage? instead of adding additional guards?

Great idea, and also addressed in the comments. New triage card created: https://linear.app/neo4j/issue/7e1f5f95-2cec-4bfb-93b5-576132463f5f

> * should graph pruning enforce UNIQUENESS? previously the pruner enforced required (now EXISTENCE) on extracted properties. With UNIQUENESS now properly modelled, should we also handle the case where the LLM extracts duplicate values for a property marked as unique? This is of course best enforced at the database level, but since we already prune for one constraint type, could it feel inconsistent to users that the other constraint type is silently ignored during pruning?

I think this is an interesting idea. I suppose it can feel a bit inconsistent. But it's a much harder constraint to enforce in the pruner, since it's "global" (need to track all node properties) instead of "local" like EXISTENCE that only has to look at the current node. So I'm not sure. I agree with you it's something that the DB is designed to handle efficiently. And I think for the end user, they might not notice whether the constraint was enforce in the pruner or the DB?

NathalieCharbel (approved):
Thank you for applying the requested changes and for the nice work 🚀

PR #511, opened 2026-04-15
Add support for KEY constraints to KG builder schema

adamnsch:
# Description

## Summary

Adds **`GraphConstraintType.KEY`** to the experimental `GraphSchema` (single property, node or relationship scope), aligned with Neo4j key constraints for **mandatory presence + identity** semantics. Pruning enforces **presence** for KEY (like EXISTENCE); **uniqueness** is still not validated in-process.

**Note:** Support for **composite constraints** will be added in an upcoming PR.

## Breaking change: `ParquetWriter` / `KGWriterModel` metadata

`metadata["files"][].columns` entries now include:

- **`is_unique`** — `true` for **UNIQUENESS**-backed properties only  
- **`is_primary_key`** — `true` for **KEY**-backed properties, synthetic **`__id__`** when no KEY, and relationship **`from`** / **`to`** columns  

`UNIQUENESS` no longer sets `is_primary_key`; consumers that relied on that must use **`is_unique`**.

## Other changes

- **`UNIQUENESS` + `KEY`** on the same node property is rejected at schema validation (Neo4j rule).
- Helpers: `key_property_names_for_*`, `uniqueness_property_names_for_node`, `mandatory_property_names_for_*`.
- Schema-from-text: `ExtractedConstraintType` and extraction filter support **`KEY`**; prompt updated.
- **`SchemaFromExistingGraphExtractor`**: Neo4j **`NODE_KEY`** / **`RELATIONSHIP_KEY`** metadata maps to **`KEY`** (existence-only constraints unchanged).
- **`CHANGELOG`** updated under **Next**.

## Tests

Unit tests cover schema validation, conflict case, pruning (node KEY), extraction, and Parquet column metadata (`is_primary_key` / `is_unique`).

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist


src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -75,38 +75,67 @@ def sanitize_parquet_filestem(name: str) -> str:
...
-                    unique_properties.append(property_name)
+def get_unique_properties_for_node_type(
+    schema: Optional[dict[str, Any]], node_label: str
+) -> list[str]:
+    """Deprecated synonym for :func:`get_primary_key_column_names_for_node_type`.
 
-    return unique_properties or default
+    Kept for backward compatibility; prefer explicit helpers for UNIQUENESS vs KEY.
+    """
+    return get_primary_key_column_names_for_node_type(schema, node_label)
```
NathalieCharbel, line 151: would be nice here to have a deprecation warning to avoid this silent semantic change (before, the function was returning uniqueness constraint properties and now key properties) and alerting the users even though we suspect this is not yet already widely used. wdyt?

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -656,6 +664,67 @@ def validate_constraints_against_node_types(self) -> Self:
...
+                            f"Constraint references undefined node type: {constraint.node_type}"
+                        )
+                    node_type = self._node_type_index[constraint.node_type]
+                    valid_property_names = {p.name for p in node_type.properties}
+                    if constraint.property_name not in valid_property_names:
+                        raise SchemaValidationError(
+                            f"KEY constraint references undefined property "
+                            f"'{constraint.property_name}' on node type '{constraint.node_type}'. "
+                            f"Valid properties: {valid_property_names}"
+                        )
```
NathalieCharbel: I know that the schema module in general is heavy in code and needs for sure some refactoring n the future (not related to changes introduced by this PR) but the added block here on key constraint seems to be almost exactly same as the one for EXISTENCE for instance. Is it worth (to avoid more code duplication) to have a helper method that does the entire check? If you look at the code, the only difference seems to be the error message prefix.

src/neo4j_graphrag/generation/prompts.py
```diff
@@ -230,7 +230,12 @@ class SchemaExtractionTemplate(PromptTemplate):
 9.3 For a relationship property, add {{"type": "EXISTENCE", "node_type": "", "property_name": "<name>", "relationship_type": "<REL_TYPE>"}}.
 9.4 Each EXISTENCE constraint must reference exactly one of node_type or relationship_type (non-empty), never both.
 9.5 Do not infer EXISTENCE from UNIQUENESS; they are independent (as in Neo4j Cypher constraints).
-10. Never use double underscores (__) as a prefix or suffix in node labels or relationship types (e.g. __Person__ or __KNOWS__ are forbidden).
+10. KEY CONSTRAINTS (optional, Neo4j NODE KEY / RELATIONSHIP KEY for a single property):
```
NathalieCharbel, line 233: now that we added key constraints maybe we should update the examples above in the prompt showing an example of key constraint (not sure why we did not have one for uniqueness in the past though 🤔) I guess this doesn't matter for structured output path but for old V1 path maybe!

tests/unit/experimental/components/test_graph_pruning.py
```diff
@@ -351,6 +351,50 @@ def test_graph_pruning_enforce_relationships_lexical_graph_with_pruned_nodes(
...
+    filtered_nodes = pruner._enforce_nodes(
+        nodes=nodes,
+        schema=schema,
+        lexical_graph_config=lexical_graph_config,
+        pruning_stats=pruning_stats,
+    )
+    assert len(filtered_nodes) == 1
+    assert filtered_nodes[0].id == "chunk-1"
+    assert pruning_stats.number_of_pruned_nodes == 1
+
```
NathalieCharbel, line 397: should we extend a bit the coverage of these tests and ensure that key properties are enforced for rels?

tests/unit/experimental/components/test_schema.py
```diff
@@ -377,6 +377,75 @@ def test_schema_constraint_validation_missing_property_name() -> None:
...
+                "node_type": "Person",
+                "property_name": "id",
+                "relationship_type": None,
+            },
+        ],
+    }
+    with pytest.raises(SchemaValidationError, match="UNIQUENESS and KEY"):
+        GraphSchema.model_validate(schema_dict)
+
+
```
NathalieCharbel, line 448: should we also add a test for `SchemaFromExistingGraphExtractor._extract_graph_constraints_from_metadata` with `NODE_KEY` or `RELATIONSHIP_KEY` type metadata input? since the change from mapping `NODE_KEY→EXISTENCE` to `NODE_KEY→KEY` happens there

NathalieCharbel (approved):
Overall looks great 🚀

A few non-blocking suggestions for improvement

src/neo4j_graphrag/experimental/components/parquet_formatter.py
adamnsch: Good idea! Done

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: Yeah, I definitely agree :)

src/neo4j_graphrag/generation/prompts.py
adamnsch: Good idea! We do have:
```
    {{
      "type": "UNIQUENESS",
      "node_type": "Person",
      "property_name": "name",
      "relationship_type": ""
    }},
```
actually but I agree it makes sense to add something for KEY constraints too :)

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: Done!

tests/unit/experimental/components/test_graph_pruning.py
adamnsch: Sure, that makes sense :)

src/neo4j_graphrag/generation/prompts.py
adamnsch: ✅

tests/unit/experimental/components/test_graph_pruning.py
adamnsch: Done!

tests/unit/experimental/components/test_schema.py
adamnsch: Good catch! Yes, we should. Done :)

PR #512, opened 2026-04-16
Support composite constraints in KG builder schemas

adamnsch:
# Description

## Summary                                                                 
                                                                         
  - Add composite (multi-property) constraint support for UNIQUENESS and KEY constraint types
  - `ConstraintType` gains a `property_names: Tuple[str, ...]` field; the old `property_name: str` field is deprecated but still accepted via a before-validator migration                                                 
  - EXISTENCE constraints are validated to exactly one property (Neo4j doesn't support composite existence)                
  - Parquet metadata now includes a structured `constraints` list alongside per-column `is_primary_key`/`is_unique` booleans, preserving composite grouping for downstream consumers
  - LLM extraction prompt updated to use `property_names` (list) format only 
  - All existing tests migrated from `property_name` to `property_names`; backward-compat test explicitly asserts the deprecation warning            

  ## Breaking changes

  None. The old `property_name` field is still accepted in input (dicts, JSON) and migrated automatically. Accessing `constraint.property_name` on an instance emits a `DeprecationWarning`.                                  
                                                      
  ## Testing

14 new unit tests cover: composite KEY/UNIQUENESS creation and validation, EXISTENCE rejection of composite, backward-compat migration, property validation, exclusivity checks, ParquetWriter metadata output with composite constraints, LLM extraction round-trip for composite KEY/UNIQUENESS, and extraction filtering (rejects composite EXISTENCE, drops composites referencing nonexistent properties). Existing tests updated to use `property_names` throughout.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


examples/customize/build_graph/components/schema_builders/schema_from_text_with_structured_output.py
```diff
@@ -81,7 +81,7 @@ def print_schema_summary(schema: GraphSchema, title: str) -> None:
         print(f"\nConstraints ({len(schema.constraints)}):")
         for constraint in schema.constraints:
             print(
-                f"  - {constraint.type} on {constraint.node_type}.{constraint.property_name}"
+                f"  - {constraint.type} on {constraint.node_type}.{list(constraint.property_names)}"
```
stellasia, line 84: Nit: why do we need to cast to `list` here?

src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -81,10 +81,19 @@ def _constraint_relationship_type_unset(constraint: dict[str, Any]) -> bool:
     return rt is None or (isinstance(rt, str) and rt.strip() == "")
 
 
+def _resolve_constraint_property_names(constraint: dict[str, Any]) -> list[str]:
+    """Resolve property names from a constraint dict (``property_names`` or ``property_name``)."""
+    pns = constraint.get("property_names") or ()
+    if pns:
+        return list(pns)
+    pn = constraint.get("property_name", "")
```
stellasia, line 89: Is this really useful since we have the `before` validator in the schema that will populate the `property_names` field?

src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -113,9 +120,51 @@ def get_key_property_names_for_node_type(
...
             continue
-        pn = constraint.get("property_name", "")
-        if pn:
-            out.append(pn)
+        out.extend(_resolve_constraint_property_names(constraint))
+    return out
+
+
+def get_key_constraints_for_node_type(
+    schema: Optional[dict[str, Any]], node_label: str
```
stellasia, line 128: Can you remind me why the schema is a dict here, and not a `GraphSchema` objet?

src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -113,9 +120,51 @@ def get_key_property_names_for_node_type(
...
+) -> list[tuple[str, ...]]:
+    """KEY constraints for a node label, preserving composite grouping.
+
+    Returns a list of tuples, each containing the property names for one KEY constraint.
+    """
+    if not schema:
+        return []
+    out: list[tuple[str, ...]] = []
+    for constraint in schema.get("constraints", ()) or ():
+        if constraint.get("type") != "KEY":
```
stellasia, line 138: I guess having an object as input would allow us to use the GraphConstraintType enum here, even if it's not blocking :)

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -262,10 +263,25 @@ class ConstraintType(BaseModel):
...
     type: GraphConstraintType
-    property_name: str
+    property_name: str = Field(
+        default="",
+        deprecated=(
+            "Use property_names (a tuple of one or more names) instead of "
+            "property_name. property_name is kept for backward compatibility."
+        ),
+    )
+    property_names: Tuple[str, ...] = ()
```
stellasia: Nit: we can include the `min_length` constraint here:

```
property_names: Tuple[str, ...] = Field(default_factory=tuple, min_length=1)
```

https://pydantic.dev/docs/validation/latest/api/pydantic/standard_library_types#constraints-10

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -1624,66 +1703,70 @@ def _extract_graph_constraints_from_metadata(
...
+        seen: set[tuple[str, ...]] = set()
 
         for constraint in schema_metadata.get("constraint", []):
             ctype = constraint.get("type")
             properties = constraint.get("properties") or []
             labels = constraint.get("labelsOrTypes") or []
             if not properties or not labels:
                 continue
-            prop = properties[0]
+            props_tuple = tuple(properties)
```
stellasia, line 1710: Nice 🥳

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -453,13 +501,16 @@ def _constraint_identity(c: Any) -> tuple[str, str, str, str]:
             if isinstance(c, ConstraintType):
                 nt = c.node_type or ""
                 rt = c.relationship_type or ""
-                return (str(c.type), nt, rt, c.property_name)
+                pn = c.property_names[0] if c.property_names else c.property_name
```
stellasia, line 500: Aren't we loosing any information by using only the first property?

stellasia (commented):
Very nice work! Left a few comments mainly to understand (I'm a few PRs behind on this topic :) )

examples/customize/build_graph/components/schema_builders/schema_from_text_with_structured_output.py
adamnsch: You're right, we don’t strictly need `list()` here. Tuples format fine in f-strings too 👍  `list()` was only for square-bracket display. We could remove it or switch to ', '.join(...) for example

src/neo4j_graphrag/experimental/components/parquet_formatter.py
adamnsch: Very good point. My first instinct was also to not have it, and maybe it's the best choice.
But while we in practice only really use `ParquetWriter` as part of a pipeline that processes a `GraphSchema` (that converts `property_name` -> `property_names`), the `ParquetWriter.run` method still takes a `schema: Optional[dict[str, Any]]`. So I guess it just feels more safe and protects against issues from unconventional usages, but maybe in that case it's better if it fails to that users update their schemas 🤷 I'm not sure. What do you prefer?

adamnsch: I suppose we might also get better typing with dataflow programming, so the `schema` in `ParquetWriter.run` might be typed anyway?

adamnsch: I also see you commented on the type below now 😂

src/neo4j_graphrag/experimental/components/parquet_formatter.py
adamnsch: Good question :) Looking at the code, I honestly cannot tell why that choice was made to have `ParquetWriter.run` take a `dict` instead of `GraphSchema`. Maybe "purer" unit testing?
I think changing this would be a nice improvement :) I'll make a card for it

adamnsch: https://linear.app/neo4j/issue/GENKGB-1183/take-graphschema-instead-of-dict-as-parquetwriterrun-input

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: Good idea :) I changed it to
```
property_names: Tuple[str, ...] = Field(min_length=1)
```
since that default doesn't actually make sense 😄 
So now we can remove 
```
        if not self.property_names:
            raise ValueError(
                "Constraint must specify at least one property in property_names."
            )
```
below

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: This code path only deals with existence constraint, and these cannot be composite so must only be for one property (it deals with deprecated `required` flags). Below it's used like this:
```
for c in constraints:
            t, nt, rt, pn = _constraint_identity(c)
            if t in (GraphConstraintType.EXISTENCE.value, "EXISTENCE"):
                seen_existence.add((nt or "", rt or "", pn))
```
Hope that makes sense :)

stellasia (approved):

PR #519, opened 2026-04-21
Ensure there is always a single-property key constraint

adamnsch:
# Description

## Summary

- In the `ParquetWriter` output, make sure that each node type has at least one one property with a single-property KEY constraints. Use the internally generated `__id__` property if there aren't any in the schema.
- Also make sure that relationships are constructed by referencing node properties that have single-property KEY constraints.

## Motivation

The Neo4j data importer (consumer of `ParquetWriter` output) needs to have node properties with single-property KEY constraints when connecting nodes with relationships. Prior to this change, it was not always the case that 1) node properties with such constraints were used for from / to node property references, and that 2) node types even necessarily had node properties with single-property KEY constraints.

## What changed

- **Node Parquet metadata:** For each node label, exported `files[].constraints` includes at least one **single-property** `KEY`. If the schema has none (no `KEY`, or only composite `KEY`s), a synthetic `KEY` on `__id__` is **added alongside** existing composite `KEY` entries.
- **Relationship Parquet:** `from`/`to` values and `start_node_primary_keys` / `end_node_primary_keys` use the **first single-property** node `KEY` for that label, or **`__id__`** when there is no such `KEY`. `UNIQUENESS` is **no longer** used to choose endpoint properties.
- **New helpers** in `parquet_formatter.py`:
  - `schema_has_single_property_key_for_node_type`
  - `enrich_key_constraints_for_node_type` — appends a synthetic `("__id__",)` KEY group when the schema has no single-property KEY for that label (composite KEYs are kept; synthetic KEY is **in addition** when needed)
  - `flatten_key_constraint_property_names`
  - `get_relationship_join_key_property_name` — first **single-property** node `KEY` for the label, else `"__id__"`
- **Docs / tests:** `KGWriterModel` docstring clarifies node `constraints` and relationship endpoint lists; `CHANGELOG.md` and `test_kg_writer.py` updated accordingly.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -168,14 +168,65 @@ def get_uniqueness_constraints_for_node_type(
...
+
+
 def get_primary_key_column_names_for_node_type(
     schema: Optional[dict[str, Any]], node_label: str
 ) -> list[str]:
-    """Column names flagged as primary key in KG writer metadata: KEY properties, else ``__id__``."""
-    keys = get_key_property_names_for_node_type(schema, node_label)
-    if keys:
-        return keys
-    return ["__id__"]
```
ali-sedaghatbaf, line 178: Used in several places, let's centralize it: 
`INTERNAL_ID_PROPERTY = "__id__"`

src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -168,14 +168,65 @@ def get_uniqueness_constraints_for_node_type(
...
+    """Property used in relationship Parquet ``from``/``to`` and endpoint metadata.
+
+    Returns the first **single-property** KEY from the schema for this label; if none,
+    returns ``__id__`` (internal node id in row values). Independent of
+    :func:`enrich_key_constraints_for_node_type` ordering, but aligned with it:
+    when the schema has no single-property KEY, the join key is ``__id__`` and
+    enrichment adds a KEY on ``__id__``.
+    """
+    for t in get_key_constraints_for_node_type(schema, node_label):
+        if len(t) == 1:
```
ali-sedaghatbaf: Nit: repeated in multiple places and if we change reqs in the future, tthen we need to update all of them.

src/neo4j_graphrag/experimental/components/parquet_formatter.py
```diff
@@ -168,14 +168,65 @@ def get_uniqueness_constraints_for_node_type(
     return out
 
 
+def schema_has_single_property_key_for_node_type(
```
ali-sedaghatbaf: It seems `get_key_constraints_for_node_type` is called 3 times per call path. 
Can we do a bit of refactoring by passing constraints explicitly here?

```
def has_single_property_key(key_constraints: list[tuple[str, ...]]) -> bool:
    return any(len(t) == 1 for t in key_constraints)
```

and then use it like this:

```
def enrich_key_constraints_for_node_type(
    schema: Optional[dict[str, Any]], node_label: str
) -> list[tuple[str, ...]]:
    base = get_key_constraints_for_node_type(schema, node_label)

    if has_single_property_key(base):
        return base

    return [*base, ("__id__",)]
```

ali-sedaghatbaf (approved):
Good work!

Just a few non-blocking comments.

src/neo4j_graphrag/experimental/components/parquet_formatter.py
adamnsch: Good idea! Done

src/neo4j_graphrag/experimental/components/parquet_formatter.py
adamnsch: Good point. Have added a helper method for this now that is reused multiple times

src/neo4j_graphrag/experimental/components/parquet_formatter.py
adamnsch: Good points. I did a refactor to address this and simplify the code a bit now. Thanks!

PR #529, opened 2026-05-18
Add relationship-scope support for UNIQUENESS, EXISTENCE, and KEY constraints

AmirLayegh:
# Description
Extends the three constraint kinds (`EXISTENCE`, `KEY`, and `UNIQUENESS`) to relationship types, mirroring how Neo4j already supports them at the database level. Previously these constraints could only be scoped to node labels; this change makes `GraphSchema` express relationship-scoped constraints as first-class citizens.

## Why
`EXISTENCE` and `KEY` constraints could already be scoped to relationship types, but `UNIQUENESS` could not. Any `relationship_type` field on a `UNIQUENESS` constraint was silently stripped. Beyond that, all three constraint types shared validation and extraction logic written under the assumption of a node scope, so relationship-scoped constraints were not properly validated.

# Changes

- Schema model: The constraint model now treats node and relationship scope symmetrically. A constraint can now be scoped to either a node label or a relationship type and exactly one must be set. Empty-string sentinels from wire formats are normalised to `None` on load so callers always get a clean value. Validation now rejects schemas where both `UNIQUENESS` and `KEY` target the same relationship type and property set.
- Schema validation: Property-existence checks (every property in a constraint must be declared on the referenced type) now run for `UNIQUENESS` as well as `EXISTENCE` and `KEY`. The `UNIQUENESS`/`KEY` exclusivity rule (the two cannot apply to the same property set) is extended to relationship types, matching the existing rule for nodes.
- LLM extraction: The adapter that normalizes raw LLM output no longer drops relationship_type from `UNIQUENESS` constraints and it only strips empty or null values. The prompt updated to instruct the model to emit relationship-scoped constraints. 
- Parquet / KG writer: Relationship file entries in the parquet manifest now carry a constraints list (`KEY`, `UNIQUENESS`, `EXISTENCE`), parallel to what node entries already produced. The key is absent when there are none, keeping the format backward-compatible with existing consumers.
- Tests: New unit tests cover the schema side (relationship-scoped `UNIQUENESS` and `EXISTENCE` accepted; `UNIQUENESS`+`KEY` on the same relationship and properties rejected), the parquet writer side (all three constraint types appear in relationship file entries; the key is absent when no relationship constraints exist; node and relationship constraints don't bleed into each other's entries), and the extraction side (relationship-scoped `UNIQUENESS` preserved end-to-end through the LLM output adapter).

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Experimental: `EXISTENCE`, `KEY`, and `UNIQUENESS` constraints can now be scoped to relationship types in `GraphSchema`. `ConstraintType` accepts a `relationship_type` field (mutually exclusive with `node_type`); validation rejects schemas where both `UNIQUENESS` and `KEY` target the same relationship type and property set. `ParquetWriter` relationship file entries now include a `constraints` list when the schema defines any for that relationship type.
 - Experimental: `SchemaFromExistingGraphExtractor._extract_graph_constraints_from_metadata` now maps `NODE_PROPERTY_UNIQUENESS` and `RELATIONSHIP_PROPERTY_UNIQUENESS` rows from `SHOW CONSTRAINTS` to the corresponding node-scoped and relationship-scoped `UNIQUENESS` constraints in `GraphSchema`.
 
+### Fixed
+
+- Experimental: `GraphPruning` now drops relationships whose `KEY`- or `EXISTENCE`-constrained properties are null, matching the existing behaviour for nodes. Previously such relationships were logged as pruned but still passed through with empty properties.
+
 ## 1.16.0
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+### Added
+
+- Experimental: `EXISTENCE`, `KEY`, and `UNIQUENESS` constraints can now be scoped to relationship types in `GraphSchema`. `ConstraintType` accepts a `relationship_type` field (mutually exclusive with `node_type`); validation rejects schemas where both `UNIQUENESS` and `KEY` target the same relationship type and property set. `ParquetWriter` relationship file entries now include a `constraints` list when the schema defines any for that relationship type.
+
 ### Fixed
 
 - Experimental: `wire_extraction_constraints_for_graph_schema` now removes `relationship_type` whenever `type == "UNIQUENESS"`, preventing a `validate_constraint_shape` `ValueError` when an LLM emits a stray `relationship_type` alongside a UNIQUENESS constraint (only node-level UNIQUENESS is supported).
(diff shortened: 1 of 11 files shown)


tests/unit/experimental/components/test_graph_schema_extraction.py
```diff
@@ -29,7 +29,7 @@
 
 
 def test_wire_extraction_constraints_drops_relationship_type_for_uniqueness() -> None:
```
adamnsch: I think we need to update this test name now :)

tests/unit/experimental/components/test_schema.py
adamnsch, line 1: Might be nice to also have a unit test for UNIQUENESS on a relationship referencing an undefined property (we have that pattern for nodes)

src/neo4j_graphrag/experimental/components/schema.py
adamnsch: We should also change the name of this one since it also handles uniqueness now. Maybe to something more generic?

src/neo4j_graphrag/experimental/components/schema.py
```diff
@@ -365,7 +358,7 @@ def _validate_existence_or_key_property_defined(
         for pn in constraint.property_names:
             if pn not in valid_property_names:
                 raise SchemaValidationError(
-                    f"{kind} constraint references undefined property "
+                    f"{kind.value} constraint references undefined property "
```
adamnsch, line 361: Nitpick: Using `kind.value` instead of `kind` below for relationship is a mismatch I think

adamnsch (commented):
Looks really good! 🚀  

The main thing that's missing for an approval is that we also need to add relationship property UNIQUENESS constraints to `SchemaFromExistingGraphExtractor._extract_graph_constraints_from_metadata`

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Next
 
+### Added
+
+- Experimental: `EXISTENCE`, `KEY`, and `UNIQUENESS` constraints can now be scoped to relationship types in `GraphSchema`. `ConstraintType` accepts a `relationship_type` field (mutually exclusive with `node_type`); validation rejects schemas where both `UNIQUENESS` and `KEY` target the same relationship type and property set. `ParquetWriter` relationship file entries now include a `constraints` list when the schema defines any for that relationship type.
+
 ### Fixed
 
 - Experimental: `wire_extraction_constraints_for_graph_schema` now removes `relationship_type` whenever `type == "UNIQUENESS"`, preventing a `validate_constraint_shape` `ValueError` when an LLM emits a stray `relationship_type` alongside a UNIQUENESS constraint (only node-level UNIQUENESS is supported).
```
adamnsch: Actually I think we also need to reconcile these changelog entries, since they contradict each other 😄

AmirLayegh: Ah you're absolutely right! Removed the entry for dropping relationship `UNIQUENESS` constraint and folded the change requested to the added section.

src/neo4j_graphrag/experimental/components/graph_pruning.py
```diff
@@ -340,6 +340,12 @@ def _validate_relationship(
                 pruning_stats,
                 prune_empty=False,
             )
+            if not filtered_props and schema.mandatory_property_names_for_relationship(
```
adamnsch, line 343: Nice! I missed this too :O

adamnsch (approved):
LGTM :) Nice work!

PR #532, opened 2026-05-26
Add support for existing graph nodes and relationsships to the entity and relationship extractor

ali-sedaghatbaf:
# Description

This pull request enhances the entity and relation extraction process by allowing the system to incorporate information about existing nodes and relationships in the knowledge graph. When provided, these existing entities are included in the prompt to the language model, instructing it to reuse IDs for matching entities instead of creating duplicates. The changes also ensure that sensitive or unnecessary properties (like embeddings) are excluded from the prompt and add comprehensive tests for the new functionality.

* The `LLMEntityRelationExtractor` now accepts an optional `existing_graph` parameter in its main methods (`run`, `run_for_chunk`, and `extract_for_chunk`). When supplied, the extractor serializes existing nodes and relationships (excluding `embedding_properties`) and includes them in the prompt to guide the language model to reuse entity IDs. 
* The `ERExtractionTemplate` prompt now conditionally adds a section listing existing nodes and relationships, formatted as JSON, and provides clear instructions to the LLM about reusing IDs for existing entities. If no existing entities are provided, this section is omitted.
* New unit tests verify that:
  - Existing entities are included in the prompt when provided.
  - `embedding_properties` are excluded from the prompt.
  - The "Existing graph entities" section is omitted when no existing entities are provided.
  - The prompt template formats the existing entity information correctly under various scenarios. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
+++ b/src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
@@ -228,13 +228,30 @@ class LLMEntityRelationExtractor(EntityRelationExtractor):
         self.prompt_template = template
 
     async def extract_for_chunk(
-        self, schema: GraphSchema, examples: str, chunk: TextChunk
+        self,
+        schema: GraphSchema,
+        examples: str,
+        chunk: TextChunk,
+        existing_graph: Optional[Neo4jGraph] = None,
     ) -> Neo4jGraph:
         """Run entity extraction for a given text chunk."""
+        existing_nodes = None
+        existing_rels = None
+        if existing_graph:
+            existing_nodes = [
+                n.model_dump(exclude={"embedding_properties"}, exclude_none=True)
+                for n in existing_graph.nodes
+            ]
+            existing_rels = [
+                r.model_dump(exclude={"embedding_properties"}, exclude_none=True)
+                for r in existing_graph.relationships
+            ]
         prompt = self.prompt_template.format(
             text=chunk.text,
             schema=schema.model_dump(exclude_none=True),
             examples=examples,
+            existing_nodes=existing_nodes,
+            existing_rels=existing_rels,
         )
 
         # Use structured output (V2) if enabled
@@ -322,10 +339,13 @@ class LLMEntityRelationExtractor(EntityRelationExtractor):
         schema: GraphSchema,
         examples: str,
         lexical_graph_builder: Optional[LexicalGraphBuilder] = None,
+        existing_graph: Optional[Neo4jGraph] = None,
     ) -> Neo4jGraph:
         """Run extraction, validation and post processing for a single chunk"""
         async with sem:
-            chunk_graph = await self.extract_for_chunk(schema, examples, chunk)
+            chunk_graph = await self.extract_for_chunk(
+                schema, examples, chunk, existing_graph
+            )
             # final_chunk_graph = self.validate_chunk(chunk_graph, schema)
             await self.post_process_chunk(
                 chunk_graph,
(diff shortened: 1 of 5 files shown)


src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
```diff
@@ -381,8 +408,9 @@ async def run(
                 schema,
                 examples,
                 lexical_graph_builder,
+                existing_graphs[i] if existing_graphs else None,
```
NathalieCharbel, line 411: so each chunk is seeing only its caller-supplied `existing_graphs[i]`.  Do we suppose we're accumulating the graphs across previous runs and that's the responsibility of the caller? But anyway, if the same new entity appears in chunk 0 and chunk 5 of the same run, two distinct IDs will be created - we are only trying to deduplicate against externally-provided (prior/persisted) state, never against the current batch? I am just trying to understand the approach and the possible drawbacks. This could be legitimate design choice but I don't think it is clear in the docstring.

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
```diff
@@ -357,7 +378,13 @@ async def run(
             lexical_graph_config (Optional[LexicalGraphConfig], optional): Lexical graph configuration to customize node labels and relationship types in the lexical graph.
             schema (GraphSchema | None): Definition of the schema to guide the LLM in its extraction.
             examples (str): Examples for few-shot learning in the prompt.
+            existing_graphs (Optional[list[Neo4jGraph]]): One subgraph per chunk, each containing nodes and relationships already in the knowledge graph that are relevant to that chunk. When provided, the LLM is instructed to reuse their IDs for matching entities instead of creating new ones. Must have the same length as chunks.
```
NathalieCharbel, line 381: would this be negatively impacted by nodes IDs being rewritten by `update_id` (where node ids are re-written to ensure their uniqueness across chunks)?

src/neo4j_graphrag/generation/prompts.py
```diff
@@ -178,6 +179,7 @@ class ERExtractionTemplate(PromptTemplate):
 Do respect the source and target node types for relationship and
 the relationship direction.
 
+{existing_entities}
```
NathalieCharbel, line 182: how would this affect prompt size?

tests/unit/test_prompt_template.py
```diff
@@ -79,3 +81,54 @@ class MyTemplate(PromptTemplate):
...
+def test_er_extraction_template_with_empty_existing_lists() -> None:
+    template = ERExtractionTemplate()
+    prompt = template.format(
+        schema={},
+        examples="",
+        text="some text",
+        existing_nodes=[],
+        existing_rels=[],
+    )
+    assert "Existing graph entities" not in prompt
```
NathalieCharbel, line 134: while all of these unit tests are great to have, it feels a bit difficult to prove the approach and that the LLM is doing the right thing in picking the existing nodes/rels without testing it with real LLM calls on a very small dataset?

NathalieCharbel (commented):
Great stuff!

I'd like to further discuss the following points that are still unclear to me:
- how do we accumulate existing graphs from previous runs or what is the intention regarding how the caller deals with the existing graph?
- how is the approach impacted by node IDs being prefixed in `post_process_chunk` -> `update_ids`?
- how does the approach hold end to end, should we prove the actual approach works as intended through a small test with "real" LLM calls?

src/neo4j_graphrag/generation/prompts.py
ali-sedaghatbaf: For our use case (i.e. supporting semi-structured data), the existing graph would include 1 or 2 nodes. Not much impact on the prompt size

src/neo4j_graphrag/experimental/components/entity_relation_extractor.py
ali-sedaghatbaf: No, don't think so.

PR #534, opened 2026-05-28
feat: add LiteParseLoader (LiteParse optional extra)

jexp:
## Summary

- Adds `LiteParseLoader` in a new `neo4j_graphrag.experimental.components.liteparse_loader` module as an alternative PDF/document loader backed by [LiteParse](https://github.com/run-llama/liteparse) — a local, zero-cloud Rust parser with optional Tesseract OCR support
- `data_loader.py` is unchanged — `LiteParseLoader` lives in its own module alongside the optional-dependency import guard
- New `liteparse` optional extra: `pip install "neo4j-graphrag[liteparse]"`
- Drop-in replacement for `PdfLoader`; pass as `file_loader=LiteParseLoader()` to `SimpleKGPipeline`

## Changes

- `src/neo4j_graphrag/experimental/components/liteparse_loader.py` — new module with `LiteParseLoader`
- `pyproject.toml` — `liteparse = ["liteparse>=2.0.0,<3.0.0"]` optional extra + mypy override
- `CHANGELOG.md` — entry under `## Next`
- `examples/customize/build_graph/components/loaders/liteparse_loader.py` — usage example
- `tests/unit/experimental/components/test_liteparse_loader.py` — 9 unit tests (fully mocked)
- `tests/unit/experimental/components/test_liteparse_loader_integration.py` — 5 integration tests (real liteparse; auto-skipped when absent)

## Usage

```python
from neo4j_graphrag.experimental.components.liteparse_loader import LiteParseLoader
from neo4j_graphrag.experimental.pipeline.kg_builder import SimpleKGPipeline

pipeline = SimpleKGPipeline(
    llm=...,
    driver=...,
    embedder=...,
    file_loader=LiteParseLoader(ocr_enabled=True),  # or LiteParseLoader() for text PDFs
    from_file=True,
)
await pipeline.run_async(file_path="document.pdf")
```

## Test plan

- [ ] `uv run pytest tests/unit/experimental/components/test_liteparse_loader.py` — 9 unit tests pass without liteparse installed
- [ ] `pip install "neo4j-graphrag[liteparse]" && uv run pytest tests/unit/experimental/components/test_liteparse_loader_integration.py` — 5 integration tests pass with liteparse installed
- [ ] Existing `test_data_loader.py` tests unaffected

PR #548, opened 2026-06-24
feat: HierarchicalTextSplitter — header-aware chunking with verbless sentence filtering

jexp:
## Summary

Adds `HierarchicalTextSplitter`, a new `TextSplitter` based on the chunking strategy from [Min et al., arXiv:2507.03226](https://arxiv.org/abs/2507.03226) — *Towards Practical GraphRAG: Efficient KG Construction and Hybrid Retrieval at Scale*.

Splits documents at section headers first, then recursively at a configurable character limit with overlap, and optionally drops verb-free sentences as a cheap pre-filter before chunking.

Related: the companion PR #547 (`SpacyEntityRelationExtractor`) uses this splitter in its example pipeline.

## Header detection strategies

| Strategy | Description | Best for |
|----------|-------------|----------|
| `"markdown"` (default) | Lines starting with `#` | MarkdownLoader, `LiteParseLoader(output_format="markdown")` |
| `"capitalization"` | Short ALL_CAPS or Title Case lines without terminal punctuation | Plain-text loaders |
| `"blank_line"` | Short lines surrounded by blank lines | Plain-text loaders |
| `"spacy_verbless"` | SpaCy-detected verbless sentences | Mixed-format documents |

## Constructor

```python
HierarchicalTextSplitter(
    max_chunk_size: int = 2048,
    chunk_overlap: int = 200,
    header_strategy: str = "markdown",  # "markdown" | "capitalization" | "blank_line" | "spacy_verbless"
    model: str = "en_core_web_sm",      # only loaded for spacy_verbless or drop_verbless_sentences=True
    drop_verbless_sentences: bool = True,
)
```

## Usage

```python
from neo4j_graphrag.experimental.components.text_splitters import HierarchicalTextSplitter

splitter = HierarchicalTextSplitter(header_strategy="markdown", max_chunk_size=2048, chunk_overlap=200)
chunks = await splitter.run(text)
```

Pairs naturally with `LiteParseLoader(output_format="markdown")` from PR #534.

## Test plan

- [ ] `uv run pytest tests/unit/experimental/components/text_splitters/test_hierarchical_splitter.py -v` (20 unit tests, spacy mocked — no model download)
- [ ] `python -m spacy download en_core_web_sm && uv run pytest tests/unit/experimental/components/text_splitters/test_hierarchical_splitter_integration.py -v` (2 integration tests)
- [ ] `uv run python -c "from neo4j_graphrag.experimental.components.text_splitters import HierarchicalTextSplitter; print('ok')"`

### DIFF
--- a/src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
+++ b/src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
@@ -1,0 +1,452 @@
+#  Copyright (c) "Neo4j"
+#  Neo4j Sweden AB [https://neo4j.com]
+#  #
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  #
+#      https://www.apache.org/licenses/LICENSE-2.0
+#  #
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Hierarchical text splitter that detects section boundaries before chunking.
+
+Supported header strategies:
+
+* ``"markdown"`` — lines starting with ``#`` (ATX Markdown headers).
+* ``"capitalization"`` — short Title Case or ALL_CAPS lines without terminal
+  punctuation (appropriate for plain-text output from loaders like
+  LiteParseLoader).
+* ``"blank_line"`` — short lines that are surrounded by blank lines on both
+  sides (a common plain-text section marker).
+* ``"spacy_verbless"`` — SpaCy-parsed sentences that are short, contain no
+  verb, and are followed by a longer sentence.
+
+All strategies produce a list of *sections* (contiguous text blocks).  Each
+section is then emitted as a single chunk when it fits within ``max_chunk_size``,
+or recursively split with ``chunk_overlap`` when it is larger.
+
+Optionally, when ``drop_verbless_sentences=True`` (default), SpaCy is used to
+remove sentences with no verb token from every emitted chunk.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Any, Optional
+
+from pydantic import validate_call
+
+from neo4j_graphrag.experimental.components.text_splitters.base import TextSplitter
+from neo4j_graphrag.experimental.components.types import TextChunk, TextChunks
+
+# ---------------------------------------------------------------------------
+# Header-detection helpers
+# ---------------------------------------------------------------------------
+
+_MARKDOWN_HEADER_RE = re.compile(r"^#{1,6}\s+\S", re.MULTILINE)
+
+# Title Case: starts with capital, no terminal punctuation at end of line,
+# short (≤ 80 chars), and at least half the words are Title-cased.
+_TITLECASE_RE = re.compile(r"^[A-Z][^\n]{0,78}$")
+_TERMINAL_PUNCT_RE = re.compile(r"[.!?;,]$")
+
+# ALL_CAPS line: all uppercase letters / spaces / digits
+_ALLCAPS_RE = re.compile(r"^[A-Z0-9][A-Z0-9 \t\-:]+$")
+
+# Blank-line boundary: a line that is non-empty, short, and preceded/followed
+# by blank lines (handled at section-split level, not a single regex).
+_SHORT_LINE_MAX = 80
+
+
+def _is_title_case(line: str) -> bool:
+    """Return True when a line looks like a Title Case heading."""
+    if not _TITLECASE_RE.match(line):
+        return False
+    if _TERMINAL_PUNCT_RE.search(line):
+        return False
+    words = line.split()
+    if len(words) == 0:
+        return False
+    capitalised = sum(1 for w in words if w and w[0].isupper())
+    return capitalised / len(words) >= 0.5
+
+
+def _is_allcaps(line: str) -> bool:
+    """Return True when a line is ALL CAPS without terminal punctuation."""
+    if not _ALLCAPS_RE.match(line):
+        return False
+    return not _TERMINAL_PUNCT_RE.search(line)
+
+
+def _split_at_markdown_headers(text: str) -> list[str]:
+    """Split *text* into sections at Markdown ATX header lines (``#``, ``##``, …)."""
+    # We split before each header line so that the header stays with its section.
+    lines = text.splitlines(keepends=True)
+    sections: list[str] = []
+    current: list[str] = []
+    for line in lines:
+        if re.match(r"^#{1,6}\s+\S", line) and current:
+            sections.append("".join(current))
+            current = [line]
+        else:
+            current.append(line)
+    if current:
+        sections.append("".join(current))
+    return [s for s in sections if s.strip()]
+
+
+def _split_at_capitalization(text: str) -> list[str]:
+    """Split *text* into sections at Title Case or ALL_CAPS heading lines."""
+    lines = text.splitlines(keepends=True)
+    sections: list[str] = []
+    current: list[str] = []
+    for line in lines:
+        stripped = line.rstrip("\r\n")
+        is_header = (
+            len(stripped) <= _SHORT_LINE_MAX
+            and stripped.strip()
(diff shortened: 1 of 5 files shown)


src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
```diff
@@ -0,0 +1,452 @@
...
+    """
+
+    @validate_call
+    def __init__(
+        self,
+        max_chunk_size: int = 2048,
+        chunk_overlap: int = 200,
+        header_strategy: str = "markdown",
+        model: str = "en_core_web_sm",
+        drop_verbless_sentences: bool = True,
```
matteomedioli, line 346: With header_strategy="spacy_verbless", the header this strategy just detected gets deleted by the default verb filter (drop_verbless_sentences=True):

```python
splitter = HierarchicalTextSplitter(header_strategy="spacy_verbless")  # drop_verbless_sentences=True by default
await splitter.run("Introduction. This section covers the basics. Conclusion. This wraps up.")
```
**Step 1**:  `_split_at_spacy_verbless` finds "Introduction." and "Conclusion." (short, no verb, followed by long sentence) → uses them as section boundaries.
**Step 2**: sections become chunks.
**Step 3**: verb filter runs on the final chunks, sees "Introduction." has no verb → deletes it. Same for "Conclusion."
**Result**: 2 chunks, no "Introduction", no "Conclusion".

Maybe we should skip verb filter on text just used as a section boundary, or warn when both options are combined together.

src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
```diff
@@ -0,0 +1,452 @@
...
+
+        chunks.append(chunk)
+
+        # Advance by step, ensuring we always make progress.
+        next_start = start + step
+        if next_start <= start:
+            next_start = start + 1
+        start = next_start
+
+    return chunks
```
matteomedioli, line 255: Nit: `chunk_overlap` width can drift a bit.

Whitespace-snapping (so words don't get chopped) shifts where each chunk ends, but the next chunk's start is still fixed `(start + (max_size - overlap))`, so the shared region between two chunks isn't always exactly `chunk_overlap` chars — just close to it. Nothing is lost, content still overlaps either way.

```python
text = "The cat sat on the warm mat today."
splitter = HierarchicalTextSplitter(max_chunk_size=15, chunk_overlap=5, header_strategy="markdown")
```
```
# chunk 0: 'The cat sat on '
# chunk 1: 't on the warm '
# chunk 2: 'arm mat today.'
# gap 0->1: tail='t on ' == head='t on '   -> matches
# gap 1->2: tail='warm ' != head='arm m'   -> off by one, drifted
```

Nit, not blocking at all, maybe might be worth a quick docstring note so it's clear the overlap is "approximate" rather than exact.

src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
```diff
@@ -0,0 +1,452 @@
...
+    A sentence is considered *verbless* when it contains no token whose
+    part-of-speech tag is ``VERB`` or ``AUX``.
+    """
+    doc = nlp(text)
+    kept: list[str] = []
+    for sent in doc.sents:
+        has_verb = any(tok.pos_ in ("VERB", "AUX") for tok in sent)
+        if has_verb:
+            kept.append(sent.text)
+    return " ".join(kept)
```
matteomedioli, line 275: Nit: text reformatted even when nothing is dropped
```python
text = "# Section\nThe team shipped the release.\nThey tested it thoroughly.\n"
HierarchicalTextSplitter().run(text)  # every sentence has a verb
```

**Step 1**: verb filter runs anyway (default True, unconditional).
**Step 2**: splits into sentences, keeps all (all have verbs).
**Step 3**: rejoins with `" ".join(...)`.
**Result**: output != input -> the newline between sentences becomes a single space, even though nothing was filtered.

Possible solution: return the original text unchanged when `len(kept) == len(list(doc.sents)) `

src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
```diff
@@ -0,0 +1,452 @@
...
+            sections.append("".join(current))
+            current = [line]
+        else:
+            current.append(line)
+    if current:
+        sections.append("".join(current))
+    return [s for s in sections if s.strip()]
+
+
+def _split_at_capitalization(text: str) -> list[str]:
```
matteomedioli, line 102: Nit, very edge case:
```python
text = "This code selects records.\nSELECT SINGLE F1 FROM VBBS INTO WA\nWHERE ID = X\nDone."
HierarchicalTextSplitter(header_strategy="capitalization").run(text)
```
Both code lines get misread as headers here: 
`SELECT SINGLE F1 FROM VBBS INTO WA`
`WHERE ID = X `
Splits the snippet line by line instead of keeping it together. low risk, just flagging.

src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
```diff
@@ -0,0 +1,452 @@
...
+from pydantic import validate_call
+
+from neo4j_graphrag.experimental.components.text_splitters.base import TextSplitter
+from neo4j_graphrag.experimental.components.types import TextChunk, TextChunks
+
+# ---------------------------------------------------------------------------
+# Header-detection helpers
+# ---------------------------------------------------------------------------
+
+_MARKDOWN_HEADER_RE = re.compile(r"^#{1,6}\s+\S", re.MULTILINE)
```
matteomedioli, line 50: defined, never referenced

src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
```diff
@@ -0,0 +1,452 @@
...
+        """Load a SpaCy model, raising a clear error when SpaCy is missing."""
+        try:
+            import spacy  # noqa: PLC0415
+        except ImportError as exc:
+            raise ImportError(
+                "SpaCy is required for this configuration of HierarchicalTextSplitter. "
+                "Install it with: pip install 'neo4j-graphrag[nlp]'"
+            ) from exc
+        try:
+            return spacy.load(model)
```
matteomedioli, line 383: Nit: loads NER/lemmatizer/etc even though only sentence boundaries + POS tags are used. We can consider `spacy.load(model, disable=["ner", "lemmatizer", "attribute_ruler"])`.

src/neo4j_graphrag/experimental/components/text_splitters/hierarchical_splitter.py
```diff
@@ -0,0 +1,452 @@
...
+        sections = self._detect_sections(text)
+        raw_chunks: list[str] = []
+        for section in sections:
+            raw_chunks.extend(self._chunk_section(section))
+
+        chunks: list[TextChunk] = []
+        for raw in raw_chunks:
+            filtered = self._filter_verbless(raw)
+            if filtered.strip():
+                chunks.append(TextChunk(text=filtered, index=len(chunks)))
```
matteomedioli, line 450: Nit: calls `_filter_verbless` once per chunk in a Python loop, which calls `nlp(text)` (line 269) each time: spaCy processes one chunk at a time instead of as a batch.

Possible other approach: 
```python
# spacy batch instead of one nlp() call per chunk:
docs = list(nlp.pipe(raw_chunks))
for raw, doc in zip(raw_chunks, docs):
    kept = [s.text for s in doc.sents if any(t.pos_ in ("VERB", "AUX") for t in s)]
    filtered = " ".join(kept) if kept else raw
    if filtered.strip():
        chunks.append(TextChunk(text=filtered, index=len(chunks)))
```

matteomedioli (approved):
Great work on this, @jexp! thanks for putting it together! I spent some time testing it locally and left some minor comments/nits along the way. None of them are blocking, but I'd value your take on a couple of them when you get a chance. Thanks!

tests/unit/experimental/components/text_splitters/test_hierarchical_splitter.py
```diff
@@ -0,0 +1,524 @@
...
+
+SpaCy is not downloaded in these tests.  Where the splitter would normally
+load a model (`drop_verbless_sentences=True` or `header_strategy="spacy_verbless"`),
+the tests patch `spacy.load` with a lightweight fake nlp object so that no
+network access or model installation is required.
+"""
+
+from __future__ import annotations
+
+from typing import Any
```
matteomedioli, line 25: unused, remove for linter

PR #530, opened 2026-05-19
feat(llm): support structured output in AnthropicLLM via response_format

SAY-5:
# Description

Closes #493. `AnthropicLLM.invoke`/`ainvoke` previously raised `NotImplementedError` for `response_format`; this adds structured output support so `AnthropicLLM` works alongside `OpenAILLM` and `VertexAILLM`. A Pydantic model is converted to a JSON schema and a passthrough dict is also accepted, both forwarded as Anthropic's `output_config`; behavior is unchanged when `response_format` is omitted. The issue is labelled `help wanted` and `good first issue` by the maintainers, and the request is "to add structured output support for `AnthropicLLM` alongside `OpenAILLM` and `VertexLLM` providers".

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [ ] Project configuration change

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [ ] Documentation has been updated
- [x] Unit tests have been updated
- [ ] E2E tests have been updated
- [ ] Examples have been updated
- [ ] New files have copyright header
- [ ] CLA (https://neo4j.com/developer/cla/) has been signed
- [x] CHANGELOG.md updated if appropriate

PR #551, opened 2026-06-26
feat(llm): add structured output support to AnthropicLLM

1HazyOne707:
## Summary

Closes #493

Implements structured output for `AnthropicLLM` via Anthropic's tool-forcing pattern, bringing it to parity with `OpenAILLM` and `VertexAILLM`.

## How it works

Anthropic doesn't expose a `response_format` parameter directly. Instead, structured output is achieved by:
1. Defining a single tool whose `input_schema` is the desired JSON schema
2. Forcing the model to call it via `tool_choice={"type": "tool", "name": <name>}`
3. Extracting the `input` field from the `tool_use` response block as a JSON string

This mirrors the approach used in Anthropic's own SDK documentation and is consistent with how `OpenAILLM` converts Pydantic models to JSON schema.

### `anthropic_llm.py`
- Add `supports_structured_output: bool = True` to `AnthropicLLM`
- Add `_build_tool_choice()` — converts a Pydantic `BaseModel` subclass or raw dict schema into Anthropic tool definition + `tool_choice`
- Add `_extract_tool_result()` — extracts JSON string from the `tool_use` block in the response
- Replace `NotImplementedError` in `__invoke_v2` and `__ainvoke_v2` with the tool-forcing implementation

### `tests/unit/llm/test_anthropic_llm.py`
- Replace `test_anthropic_invoke_v2_with_response_format_raises_error` with structured output coverage:
  - Pydantic model (sync)
  - Raw dict schema (sync)
  - Missing tool_use block raises `LLMGenerationError`
  - Pydantic model (async)
  - `supports_structured_output` flag


src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -54,6 +54,7 @@
 
 # pylint: disable=redefined-builtin, arguments-differ, raise-missing-from, no-else-return, import-outside-toplevel
 class AnthropicLLM(LLMBase):
+    supports_structured_output: bool = True
```
NathalieCharbel: This is inserted above the class docstring, can you please move it below?

tests/unit/llm/test_anthropic_llm.py
```diff
@@ -413,24 +413,126 @@ def test_anthropic_llm_invoke_v2_empty_response_error(mock_anthropic: Mock) -> N
...
+
+    assert response.content == json.dumps({"value": "async structured result"})
+    call_kwargs = mock_model.messages.create.call_args[1]
+    assert call_kwargs["tool_choice"] == {"type": "tool", "name": "TestModel"}
+
+
+def test_anthropic_llm_supports_structured_output(mock_anthropic: Mock) -> None:  # noqa: ARG001
+    """Test that AnthropicLLM advertises structured output support."""
+    llm = AnthropicLLM(model_name="claude-3-opus")
+    assert llm.supports_structured_output is True
```
NathalieCharbel, line 525: Can you please update:
- docs/source/user_guide_rag.rst: add an Anthropic subsection alongside OpenAI/VertexAI, and update the line stating other providers raise `NotImplementedError` for `response_format`?
- CHANGELOG.md

Can you also add:
- anthropic_llm_structured_output.py to examples/customize/llms/ for parity with the openai/vertexai examples?

NathalieCharbel (changes requested):
Thanks @1HazyOne707 for taking this up and for your contribution!

The tool-forcing path is valid for Structured Output. However, this is a bit different from Anthropic's guidance and our design for existing providers where Structured Output is supported.

Anthropic now ships a first-class structured-output API via `output_config` with `type: "json_schema"`, which uses constrained decoding to guarantee schema-conforming output (https://platform.claude.com/docs/en/build-with-claude/structured-outputs). Tool-forcing (defining a tool + tool_choice) is the older workaround from before that feature existed.
Also, for consistency with our other providers: `OpenAILLM` and `VertexAILLM` both implement structured output through the provider's native feature (OpenAI's `response_format` `json_schema`; Vertex's `GenerationConfig` `response_schema`), not a tool/function fallback - even though OpenAI offers one. Using `output_config` for Anthropic keeps all three providers on the same "native SO" pattern; tool-forcing would make Anthropic the outlier.
Could you switch the implementation to build an `output_config` payload - i.e. `{"format": {"type": "json_schema", "schema": <schema>}}` passed to `messages.create` - and read the result back from `response.content[0].text` (which stays a JSON string, so `Model.model_validate_json(...)` still works, consistent with the other providers)? That removes the tool-channel coupling entirely and matches Anthropic's recommendation. Please note that PR #530 is just a concrete example of that approach. 

Smaller items regardless of approach are noted inline below.

1HazyOne707:
Thanks for the detailed review — addressed everything:

**Core change**: replaced the tool-forcing implementation with Anthropic's native `output_config` API (json_schema format), matching the pattern from #530 and bringing AnthropicLLM in line with how OpenAILLM/VertexAILLM implement structured output. `_build_tool_choice`/`_extract_tool_result` are gone, replaced by a single `_build_output_config` static method. Both `__invoke_v2` and `__ainvoke_v2` now read the result the same way as unstructured responses (`response.content[0].text`), no special tool_use extraction needed.

**Docstring placement**: moved `supports_structured_output` below the class docstring.

**Docs/changelog/example**: `docs/source/user_guide_rag.rst` has a new Anthropic Structured Output subsection alongside OpenAI/VertexAI, the stale 'other providers raise NotImplementedError' line no longer includes Anthropic, CHANGELOG.md has an entry (including a note on the supports_structured_output default becoming True), and `examples/customize/llms/anthropic_llm_structured_output.py` shows both the Pydantic and raw JSON schema paths.

Tests updated accordingly — Pydantic model case, dict case, no-response-format-omits-output_config case, and an async case, all verifying output_config is built/passed correctly. CLA and semgrep checks passing.

1HazyOne707:
Fixed the strict mypy failures on the Python 3.14 job — 48 errors down to 0. Two issues: the `system` param's sentinel type didn't match the current SDK stubs, and the response-content extraction assumed every block was a `TextBlock` without narrowing for the wider union type the SDK now returns. Verified locally with `uv run mypy` (clean) and the full unit suite (24/24 passing).

NathalieCharbel:
Hi @1HazyOne707. Thank you for addressing my comments and for your prompt responses.
I also did some more refactoring and lint fixes. 
One important thing is to ensure that `SchemaFromTextExtractor` and `LLMEntityRelationExtractor` are working with anthropic models (old and new ones) especially that the path using Structured Output (SO) is now enabled by default there. From running a simple example I noticed that there are some errors when using SO after the changes here, I am suspecting things related to having a strict json schema. I am investigating now the issue. Once we resolve this, we will be ready to go 🚀

NathalieCharbel:
@1HazyOne707 I just added a fix for the structured-output kg extraction that was failing on Anthropic with 
`400 - additionalProperties: object is not supported`.
Anthropic's constrained decoding only accepts closed schemas and can't represent open maps like `Neo4jNode.properties: dict[str, PropertyValue]` (Anthropic-specific). Forcing `additionalProperties: false` would just drop all property values.

What [1c6e3c4](https://github.com/neo4j/neo4j-graphrag-python/pull/551/commits/1c6e3c4bf19c157dfd986968015a80c703c6bcf1) does (all in `anthropic_llm.py`): rewrites open maps into closed `[{key, value}]` arrays before the call, then converts them back to maps in the response, so the returned content stays identical to what `Neo4jGraph` expects. No public types or other providers touched.

Note: this is an intentional, self-contained intermediate fix. Proper cross-provider strict-JSON-schema handling is planned as a separate PR so we don't have to hold this one.

NathalieCharbel (approved):
LGTM now 🎉

Thanks @1HazyOne707 again for your contribution.

PR #554, opened 2026-07-05
Use uniqueness constraint for KG writer temp ids

### LINKED ISSUE #159: Use Constraints Instead of Plain Range Index for Node keys/ids
Regarding stuff like this in kg writer: https://github.com/neo4j/neo4j-graphrag-python/blob/main/src/neo4j_graphrag/experimental/components/kg_writer.py#L103

For unique node ids/keys used in loading, you want to use a Constraint: Either [Key](https://neo4j.com/docs/cypher-manual/current/constraints/#node-key) or [Uniqueness](https://neo4j.com/docs/cypher-manual/current/constraints/#unique-node-property) constraint. Key is like uniqueness but also doesn't allow Nulls and is Enterprise only.

dylothx:
# Description

Fixes [#159](https://github.com/neo4j/neo4j-graphrag-python/issues/159) 

Replaces the KG writer’s temporary `__KGBuilder__.__tmp_internal_id` range index with a uniqueness constraint.

`Neo4jWriter` uses `__tmp_internal_id` to match nodes when creating relationships during graph writes. A plain index improves lookup performance, but does not prevent duplicate temporary ids. The new uniqueness constraint preserves indexed lookup behavior while also enforcing that temporary ids are unique during the write process.

This uses a uniqueness constraint rather than a node key constraint because `__tmp_internal_id` is removed during writer cleanup.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/components/kg_writer.py
+++ b/src/neo4j_graphrag/components/kg_writer.py
@@ -215,9 +215,18 @@ class Neo4jWriter(KGWriter):
         self.is_version_5_24_or_above = is_version_5_24_or_above(version_tuple)
 
     def _db_setup(self) -> None:
-        self.driver.execute_query("""
-        CREATE INDEX __entity__tmp_internal_id IF NOT EXISTS FOR (n:__KGBuilder__) ON (n.__tmp_internal_id)
-        """)
+        self.driver.execute_query(
+            "DROP INDEX __entity__tmp_internal_id IF EXISTS",
+            database_=self.neo4j_database,
+        )
+        self.driver.execute_query(
+            """
+            CREATE CONSTRAINT __entity__tmp_internal_id_unique IF NOT EXISTS
+            FOR (n:__KGBuilder__)
+            REQUIRE n.__tmp_internal_id IS UNIQUE
+            """,
+            database_=self.neo4j_database,
+        )
 
     @staticmethod
     def _nodes_to_rows(
--- a/src/neo4j_graphrag/components/kg_writer.py
+++ b/src/neo4j_graphrag/components/kg_writer.py
@@ -289,7 +289,7 @@ class Neo4jWriter(KGWriter):
             support_variable_scope_clause=self.is_version_5_23_or_above,
             batch_size=self.batch_size,
         )
-        with self.driver.session() as session:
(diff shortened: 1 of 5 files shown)


AmirLayegh (commented):
Thanks for the contribution! I left one comment that may be worth addressing before merging.

Also, would this fix be worth adding to `CHANGELOG.md` under `Fixed`? It changes the handling of duplicate temporary node IDs and fixes #159.

Could you also please sign the Contributor License Agreement as advised [here](https://neo4j.com/developer/contributing-code/#sign-cla)?

src/neo4j_graphrag/components/kg_writer.py
```diff
@@ -215,9 +215,18 @@ def __init__(
         self.is_version_5_24_or_above = is_version_5_24_or_above(version_tuple)
 
     def _db_setup(self) -> None:
-        self.driver.execute_query("""
-        CREATE INDEX __entity__tmp_internal_id IF NOT EXISTS FOR (n:__KGBuilder__) ON (n.__tmp_internal_id)
-        """)
+        self.driver.execute_query(
+            "DROP INDEX __entity__tmp_internal_id IF EXISTS",
+            database_=self.neo4j_database,
```
AmirLayegh, line 220: Should `_db_cleaning()` target `self.neo4j_database` as well? `_db_setup()`, `_upsert_nodes()` and _upsert_relationships() all explicitly use it, but `_db_cleaning()` opens `self.driver.session()` without specifying a database, which would use the user's home database.
If `neo4j_database` is different from the home database, the temporary IDs would be written to the target DB but cleaned from the home DB. With this uniqueness constraint, a subsequent write with the same temporary IDs will fail.

dylothx: Hey! Yes, you're right, I have pushed a fix to open the cleanup session with database=self.neo4j_database (Also rebased onto latest main to clear merge conflict). Thanks!

AmirLayegh (approved):
Nice work! LGTM.

PR #562, opened 2026-07-16
fix(llm): route http_client to the matching sync/async Anthropic client

matteomedioli:


matteomedioli:
Accidentally closed by a branch rename (no content lost, no review activity on this PR). Reopened as #565.

PR #565, opened 2026-07-16
fix(llm): route http_client to the matching sync/async Anthropic client

matteomedioli:
## Stack

# Description

`AnthropicLLM` builds two SDK clients — one sync, one async — from the same `**kwargs`. Most settings are safe to share between them, but `http_client` is not: the sync client needs an `httpx.Client` and the async client needs an `httpx.AsyncClient`. Today whichever one you pass is forwarded to **both** clients, so one of them always receives the wrong type. In practice there is no way to inject a custom HTTP client into `AnthropicLLM` at all.

This PR fixes that:

- `http_client` now goes only to the client whose type it matches.
- Passing anything else emits a warning and is ignored (both clients fall back to their defaults) — the same behavior `OpenAILLM` already has.
- Every other kwarg is still shared between both clients, unchanged.

No new parameters, no signature changes. If you never pass `http_client`, nothing changes for you.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,6 +97,7 @@ ENVIRONMENT MANAGEMENT:
 - `WeakKeyDictionary` works for caching per-driver since `neo4j.Driver` is hashable
 - Neo4j 2026 CREATE VECTOR INDEX syntax: WITH clause must come BEFORE OPTIONS, not after
 - E2E tests for SEARCH clause: use `docker compose -f tests/e2e/docker-compose.neo4j2026.yml up -d`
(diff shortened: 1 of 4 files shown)


stellasia (approved):
LGTM!

PR #566, opened 2026-07-16
feat(llm): extract BaseAnthropicLLM, add base_url, export extension points

matteomedioli:
## Stack

# Description

Today, using `AnthropicLLM`'s logic against a different endpoint — a proxy, an internal gateway, a self-hosted compatible API — means forking the whole class. This PR makes it a small subclass instead, using the same pattern OpenAI already has (`BaseOpenAILLM` → `OpenAILLM` / `AzureOpenAILLM`).

- **New `BaseAnthropicLLM`** holds everything that doesn't depend on how the connection is made: message building, schema conversion, response parsing, structured output. `AnthropicLLM` becomes a thin subclass that only creates the two SDK clients. Logic moved as-is — no behavior change (full unit suite passes unchanged).
- **`AnthropicLLM` and `OpenAILLM` both gain an explicit `base_url` parameter** to point at a custom endpoint. It already worked when passed through `**kwargs`; the explicit parameter makes it visible in the signature, docstring, and API docs, and keeps the two classes symmetric.
- **`BaseAnthropicLLM`, `BaseOpenAILLM`, and `split_http_client_kwargs` are now exported** from `neo4j_graphrag.llm` as supported extension points (`BaseOpenAILLM` existed but was never exported).
- **New docs page** (`llm_extensibility.rst`) explains the `base_url` / `http_client` contract, with a working subclass example.

Notes for reviewers — deliberate decisions, happy to change any of them:

- Like the existing `BaseOpenAILLM`, the new base class has no abstract methods, so it can technically be instantiated (and fails on first call if no client was set). Kept consistent with the OpenAI base; making `client`/`async_client` abstract properties across both is a possible follow-up.
- Once #571/#572 land, the docs page should be extended to cover the Gemini and VertexAI bases as well.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

- [x] CHANGELOG.md updated


docs/source/llm_extensibility.rst
```diff
@@ -0,0 +1,89 @@
+.. _llm-extensibility:
```
stellasia: Not for this PR, but we have a long section about the different LLM providers that could better fit in a separate file like this one. Wondering if we could rename it to something more generic so that we can move content here later on?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -50,8 +51,11 @@
 )
 
 if TYPE_CHECKING:
-    from anthropic import Omit
+    from anthropic import AsyncAnthropic, Anthropic, Omit
     from anthropic.types.message_param import MessageParam
+else:
```
stellasia: Out of curiosity, why do we need this else block?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -530,3 +515,55 @@ def get_messages_v2(
...
+        )
+        llm.invoke("Who is the mother of Paul Atreides?")
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        model_params: Optional[dict[str, Any]] = None,
+        rate_limit_handler: Optional[RateLimitHandler] = None,
+        base_url: Optional[str] = None,
```
stellasia, line 555: Can we have both `base_url` and `http_client` parameters?

src/neo4j_graphrag/llm/anthropic_llm.py
```diff
@@ -168,35 +172,19 @@ def _restore_open_maps(value: Any, schema: dict[str, Any], defs: dict[str, Any])
 
 
 # pylint: disable=redefined-builtin, arguments-differ, raise-missing-from, no-else-return, import-outside-toplevel
-class AnthropicLLM(LLMBase):
```
stellasia, line 171: What's the advantage of this split? Is the plan to have more subclasses depending on the `BaseAnthropicLLM` class?

matteomedioli: Yes, that's the intent. The base holds everything Anthropic-shaped (message building, the strict JSON-schema, response parsing, structured-output support), a subclass only assigns client/async_client. That enables calling Claude through other routes like direct SDK or VertexAI.

docs/source/llm_extensibility.rst
matteomedioli: renamed to `lm.rst` with a top-level "LLMs" title, the current content is now an "Extending LLMs" section, so the provider section from other docs can move here later. Kept the llm-extensibility label so existing cross-references still resolve. tested with `cd docs && make html`

src/neo4j_graphrag/llm/anthropic_llm.py
matteomedioli: Removed it, not needed. Autodoc is actually better without it: it renders `client: Anthropic` instead of `client: Any.`

src/neo4j_graphrag/llm/anthropic_llm.py
matteomedioli: Yes, and if you prefer, I can expose `http_client` as an explicit parameter (currently it's `**kwargs`; small, non-breaking change).

I verified this e2e with the real anthropic SDK against a mocked Anthropic endpoint:

- `base_url` reaches both internal SDK clients: `invoke` (sync) and `ainvoke` (async) both hit the custom endpoint.
- Passing `httpx.Client` → the custom client is used by the sync SDK client only, i.e. by `invoke()`. The async client doesn't get it (wrong type): it falls back to the SDK's default transport, but still points at `base_url` — so async calls keep working, just without the custom transport. Same logic mirrored for `httpx.AsyncClient`.

```python
# works — but only invoke() uses the proxy; ainvoke() falls back to the SDK default transport (async)
llm_sync = AnthropicLLM(model_name="...", base_url="https://my-sync-endpoint", http_client=httpx.Client(proxy="..."))

# works — but only ainvoke() uses the proxy; invoke() falls back to the SDK default transport (sync)
llm_async = AnthropicLLM(model_name="...", base_url="https://my-async-endpoint", http_client=httpx.AsyncClient(proxy="..."))

llm_sync.invoke()
llm_async.ainvoke()

# Edge cases: still work, silently ignoring the provided client
llm_sync.ainvoke()
llm_async.invoke()
```

For the edge cases we could add a call-time warning, but I'd rather not: the cleaner long term fix is specializing the sync/async interfaces as discussed (e.g. separate sync/async classes), which removes the situation entirely instead of warning about it. Happy to open a follow-up for that.

One footgun I did guard: the SDK ignores a base_url set on the httpx client itself, so I added a UserWarning at construction for it and st. Is that enough, or is it somethingwe'd rather prevent?

```python
# ignored — httpx base_url never used, ic.com
AnthropicLLM(model_name="...", http_client=httpx.Client(base_url="https://my-endpoint"))
```

stellasia: Ok, I see - another argument to split sync and async classes, this is quite unclear as it is now.

stellasia (approved):
Thanks for the answers!

PR #567, opened 2026-07-16
refactor(llm): extract shared http_client sync/async split helper

matteomedioli:
## Stack

# Description

The routing logic fixed in #565 — decide whether `http_client` belongs to the sync or the async SDK client — existed as three identical copies: in `AnthropicLLM`, `OpenAILLM`, and `AzureOpenAILLM`. Any new class constructing its own clients would have needed a fourth copy.

This PR moves it into one shared function, `split_http_client_kwargs`. It takes the constructor kwargs and returns two dicts — one for the sync client, one for the async client — with `http_client` placed in whichever one it matches. All three classes now call it.

Pure refactor, no behavior change. The function has its own unit tests (routing in both directions, invalid type warns and drops, input dict is never modified). Custom subclasses that build their own SDK clients can call it too, so they inherit the fix instead of copying the logic — it becomes a documented public export in #566, where the extension docs live.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/src/neo4j_graphrag/llm/utils.py
+++ b/src/neo4j_graphrag/llm/utils.py
@@ -113,8 +113,10 @@ def split_http_client_kwargs(
     elif httpx is not None and isinstance(http_client, httpx.AsyncClient):
         async_kwargs["http_client"] = http_client
     elif http_client is not None:
+        # stacklevel=3 attributes the warning to the caller of the LLM
+        # constructor, not to the constructor's own call into this helper.
         warnings.warn(
             f"Invalid http_client type (got {type(http_client)}, expected httpx.Client or httpx.AsyncClient). Using default client.",
-            stacklevel=2,
+            stacklevel=3,
         )
     return sync_kwargs, async_kwargs
--- a/src/neo4j_graphrag/llm/utils.py
+++ b/src/neo4j_graphrag/llm/utils.py
@@ -14,12 +14,17 @@
 #  limitations under the License.
 from __future__ import annotations
 import warnings
-from typing import Union, Optional
+from typing import Any, Union, Optional
 
 from pydantic import TypeAdapter
 
 from neo4j_graphrag.message_history import MessageHistory
 from neo4j_graphrag.types import LLMMessage
+
+try:
+    import httpx
+except ImportError:
+    httpx = None  # type: ignore[assignment]
 
 
 def system_instruction_from_messages(messages: list[LLMMessage]) -> str | None:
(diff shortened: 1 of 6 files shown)


tests/unit/llm/test_llm_utils.py
```diff
@@ -63,3 +65,58 @@ def test_legacy_inputs_prompt_as_message_history() -> None:
...
+def test_split_http_client_kwargs_no_http_client() -> None:
+    sync_kwargs, async_kwargs = split_http_client_kwargs({"api_key": "sk-test"})
+    assert sync_kwargs == {"api_key": "sk-test"}
+    assert async_kwargs == {"api_key": "sk-test"}
+    assert "http_client" not in sync_kwargs
+    assert "http_client" not in async_kwargs
+
+
+def test_split_http_client_kwargs_routes_sync_client() -> None:
+    client = httpx.Client()
```
stellasia: Nit: I think we could also use a fixture to deal with client.close, something like:

```
@fixture(scope="function")
def httpx_sync_client():
    client = httpx.Client()
    yield client
    client.close()
```

tests/unit/llm/test_llm_utils.py
```diff
@@ -63,3 +65,58 @@ def test_legacy_inputs_prompt_as_message_history() -> None:
...
+        assert sync_kwargs["http_client"] is client
+        assert sync_kwargs["api_key"] == "sk-test"
+        assert "http_client" not in async_kwargs
+        assert async_kwargs["api_key"] == "sk-test"
+    finally:
+        client.close()
+
+
+def test_split_http_client_kwargs_routes_async_client() -> None:
+    client = httpx.AsyncClient()
```
stellasia: If we also need to close the client, we could use an async test?

src/neo4j_graphrag/llm/utils.py
```diff
@@ -14,13 +14,18 @@
...
 import warnings
-from typing import Union, Optional
+from typing import Any, Union, Optional
 
 from pydantic import TypeAdapter
 
 from neo4j_graphrag.message_history import MessageHistory
 from neo4j_graphrag.types import LLMMessage
 
+try:
```
stellasia, line 24: We can maybe make `httpx` a required dependency, wdyt?

stellasia (approved):
Looks good. Left some minor comments but can be merged as it is.

src/neo4j_graphrag/llm/utils.py
matteomedioli: I'd keep it optional: the helper only needs `httpx` for `isinstance` checks, and every SDK extra that can reach this code path (anthropic/openai/cohere/mistralai/ollama) already brings `httpx` transitively. Making it a core dependency would add an HTTP stack to the base install without changing behavior. Happy to revisit if we ever use `httpx` directly in core.

PR #571, opened 2026-07-17
feat(llm): extract BaseGeminiLLM, add base_url support

matteomedioli:
## Stack

# Description

Same split as #566, applied to Gemini.

- **New `BaseGeminiLLM`** holds all the shared logic: message building, config and schema building, response parsing, tool handling. **`GeminiLLM`** becomes a thin subclass whose only job is creating the `genai.Client`.
- `BaseGeminiLLM` is exported from `neo4j_graphrag.llm` as a supported extension point. A subclass reaching a custom endpoint just builds its own client — everything else is inherited. A test proves the contract: a minimal subclass that only assigns a client runs `invoke()` end to end.
- **New explicit `base_url` parameter on `GeminiLLM`**, mirroring #566. The `google-genai` SDK has no top-level `base_url` argument, so the value is applied through `http_options`; when `http_options` is also passed (dict or `types.HttpOptions`), only its `base_url` field is overridden. All three merge paths are tested, plus the negative path (no `base_url` → nothing passed).
- **`aclose()` now closes the `genai.Client`** (sync and `aio` surfaces), so `with GeminiLLM(...)` / `async with` actually release resources like the other providers.
- Docs: `llm.rst` extensibility page extended with a Gemini section (single-client model, `http_options` instead of `http_client`), `api.rst` cross-references, `base_url` example added.

Behavior notes:

- The extraction itself is behavior-neutral (commits 1–2); `base_url` and `aclose` are small additive features on top.
- An earlier version of this PR also enabled the `supports_structured_output` flag for Gemini. That silently changes pipeline defaults for existing users, so it moved to its own PR (#573) with a proper breaking-change note.
- `GeminiLLM`'s parent class changes from the two interface classes to `LLMBase`, which extends both — every existing `isinstance` check still passes, no deprecation warning fires, and the class gains working context-manager support (`with GeminiLLM(...) as llm:`) like the other providers.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

- [x] CHANGELOG.md updated


src/neo4j_graphrag/llm/google_genai_llm.py
```diff
@@ -275,6 +275,10 @@ async def ainvoke_with_tools(
         except Exception as e:
             raise LLMGenerationError(f"Error calling GeminiLLM with tools: {e}") from e
 
+    async def aclose(self) -> None:
+        self.client.close()
```
stellasia, line 279: Weird that we need both. Do they also have both a sync and an async connection internally?

stellasia (approved):

src/neo4j_graphrag/llm/google_genai_llm.py
matteomedioli: Seems so:
- Client.close() and AsyncClient.aclose()
https://github.com/googleapis/python-genai/blob/v1.62.0/google/genai/client.py

- BaseApiClient.close() / aclose():
https://github.com/googleapis/python-genai/blob/v1.62.0/google/genai/_api_client.py

I run a quick minimal test:
```python
import asyncio
from google import genai

async def main() -> None:
    client = genai.Client(api_key="fake")
    api = client._api_client
    client.close()
    print("after close():       sync closed:", api._httpx_client.is_closed,
          "| async closed:", api._async_httpx_client.is_closed)  # True | False
    await client.aio.aclose()
    print("after aio.aclose():  sync closed:", api._httpx_client.is_closed,
          "| async closed:", api._async_httpx_client.is_closed)  # True | True
asyncio.run(main())
```

PR #575, opened 2026-07-22
Save embedding_model_name in the Chunk node

stellasia:
# Description

### What 

Save the embedding model name in the `Chunk` node.

### How

- Add a `model` (+ `dimensions`) attribute to all `Embedder` classes
- Add properties / warnings for backward compatible changes -> only breaking change for users using a custom Embedder

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Breaking change
- [ ] Documentation update

## Complexity

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


CHANGELOG.md
```diff
@@ -5,10 +5,12 @@
 ### Added
 
 - `AnthropicLLM` now supports structured output via the `response_format` argument, accepting a Pydantic model or an Anthropic `output_config` dict, alongside `OpenAILLM` and `VertexAILLM`.
+- Chunk node metadata now contains `embedding_model_name` and `embedding_dimensions
```
williedoran-neo4j, line 9: Might be misreading this — the changelog says chunk metadata now holds `embedding_model_name` and `embedding_dimensions`, but I can only find `embedding_model_name` being set in `components/embedder.py`. Was `embedding_dimensions` meant to land too? If so it looks like a one-liner and would be genuinely handy; if not, just checking the wording before we ship it. (Also the entry's missing a closing backtick on `embedding_dimensions`.)

src/neo4j_graphrag/components/embedder.py
williedoran-neo4j, line 61: Quick question — _async_embed_chunk sets embedding_model_name but _embed_chunk doesn't. I might be missing something, but run() looks like it always goes async and I couldn't find any callers of the sync method, so I'm guessing this doesn't matter in practice. 
 Curious whether the intent is to keep them in sync anyway, or the sync path is on its way out.

src/neo4j_graphrag/components/embedder.py
```diff
@@ -83,6 +83,7 @@ async def _async_embed_chunk(
             embedding = await self._embedder.async_embed_query(text_chunk.text)
             metadata = text_chunk.metadata if text_chunk.metadata else {}
             metadata["embedding"] = embedding
+            metadata["embedding_model_name"] = getattr(self._embedder, "model", None)
```
williedoran-neo4j, line 86: Small one — since `getattr(embedder, "model", None)` defaults to `None`, a custom embedder with no `model` would store `embedding_model_name: None` rather than omit the key. Curious if that's deliberate (predictable schema) or if it'd be cleaner to skip the key when there's no model. Either's fine by me.

src/neo4j_graphrag/embeddings/bedrock.py
```diff
@@ -82,15 +83,21 @@ def __init__(
                 "Could not import boto3 python client. "
                 'Please install it with `pip install "neo4j-graphrag[bedrock]"`.'
             )
-        super().__init__(rate_limit_handler)
-        self.model_id = model_id
-        self.dimensions = dimensions
+        super().__init__(model_id, dimensions, rate_limit_handler)
```
williedoran-neo4j, line 86: Am I right that Bedrock is staying on `model_id` for now, with `model` exposed as a deprecated read-only property? If so, the changelog's "subclasses now require a `model` argument" note doesn't quite cover it. Not sure whether keeping Bedrock on `model_id` is intentional or it's meant to move across like the others — whichever way you want to go, I'm happy to accommodate.

PR #578, opened 2026-08-03
fix(examples): make OpenAI examples work with gpt-5

williedoran-neo4j:
# Description

Follow-up to #482, which moved the examples and user guides from `gpt-4o` to `gpt-5`. gpt-5 constrains three parameters the examples relied on, so 17 files stopped working. The failures are immediate rather than intermittent, so they were simply never run — nothing in CI executes anything under `examples/`.

Verified against the live OpenAI API:

| parameter on gpt-5 | result |
| --- | --- |
| `max_tokens` | HTTP 400 — *"Unsupported parameter: use `max_completion_tokens` instead"* |
| `temperature` (any value other than 1) | HTTP 400 — *"does not support 0 with this model. Only the default (1) value is supported"* |
| `max_completion_tokens` set too low | HTTP 200 but **empty content** |

The third is the least obvious. Reasoning tokens are billed against `max_completion_tokens`, so a small budget is consumed entirely by reasoning and the response comes back as an empty string. Since `OpenAILLM` discards `finish_reason`, this surfaces as `LLMGenerationError: LLM response is not valid JSON`, which points at JSON parsing rather than at the token budget.

Two variations worth noting: `temperature` also fails when passed as a keyword argument to `invoke()`, not only via `model_params`; and LangChain's `ChatOpenAI` always sends a temperature of its own (default 0.7), so a gpt-5 model fails there even when the argument is omitted.

## What changed

Fixed per example intent rather than uniformly, since the examples want different things:

- **Where output stability matters** — text2cypher (the generated Cypher is executed directly, with no retry), tool calls, and structured output — moved to `gpt-4.1` / `gpt-4.1-mini` and kept `temperature=0`.
- **Where answers are free text** (RAG answer generation) — stayed on `gpt-5` and dropped `temperature`.
- **Extraction pipelines** — `max_completion_tokens: 16000` plus `reasoning_effort: "low"`. Measured on `examples/kg_builder.py`, `"low"` roughly halves the cost per run (~$0.24 → ~$0.12) with no measurable difference in extraction quality; run-to-run variance exceeded the difference between the effort settings.

Each change carries a short comment explaining the model choice, so a future bulk model update doesn't silently reintroduce the same problem.

Also included:

- The equivalent snippets in `user_guide_rag.rst` and `user_guide_kg_builder.rst`, plus a note documenting the reasoning-model constraints. The Anthropic snippets are deliberately unchanged — `max_tokens` is required for `AnthropicLLM`.
- **The two `README.md` snippets**, which carried the same failing combination and are the most-read code in the project.
- **Three class docstrings that autodoc renders into the API reference** — `components/schema.py` and two in `entity_relation_extractor.py`. One of them also passed `"response_format": {"type": "object"}`, which is not a valid OpenAI value.
- `examples/kg_builder.py` now resolves its PDF relative to the file rather than the working directory, matching the other examples, and its stale path is corrected.

## Tooling split out

An earlier version of this PR also added `scripts/check_examples.py` and two pre-commit hooks. Those are now **#598**, so this one is a plain bug fix and can be reviewed as such. The checker is opt-in there — manual stage, live run — rather than something every contributor pays for on commit.

For the record, run against `main` those static rules report **22 problems across 17 files**, which is what this PR fixes.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

Every fixed example was run against the live API rather than only reviewed. Full local run: `ruff format --check`, `ruff check`, `mypy --strict`, the unit suite, and the docs build. Coverage is unaffected — nothing under `examples/` is measured.

Not verified: two of the fixed examples can't be run end-to-end for reasons unrelated to this change — `graphrag_with_neo4j_message_history.py` writes message-history nodes to the read-only `recommendations` demo database, and `custom_prompt.py` needs a vector index that no setup script creates. In both cases the gpt-5 error is resolved and the remaining failure is pre-existing. Vector-retriever examples against `demo.neo4jlabs.com` also currently fail on an unrelated Cypher dialect issue, so those were verified via the procedure-based path.

# Checklist


oskarhane (approved):
LGTM. Checked out and reviewed the full diff locally.

What I verified:

- **No leftovers.** Every remaining `temperature` in `examples/`, `docs/source/`, `README.md` and `src/` now sits on a `gpt-4.1`/`gpt-4.1-mini` model; every remaining `max_tokens` is in an `AnthropicLLM` context, where it's required. Nothing was missed.
- **The path fix is real.** `examples/pipeline/` doesn't exist — the PDF lives in `examples/data/`, so `kg_builder.py` was broken independently of the model issue. Resolving via `Path(__file__).parent` matches the other examples.
- **`"type": "object"` → `"type": "json_object"`** in the `entity_relation_extractor.py` docstring is a genuine correctness fix; `object` is not a valid OpenAI value, and autodoc renders that block into the API reference.
- `ruff format --check` and `ruff check` clean; all 11 CI checks passing.

The per-example split is the right call rather than a uniform substitution — free-text RAG answers tolerate sampling variation, but text2cypher executes its output directly and the tool-selection examples assert which tool gets picked, so those genuinely need `temperature=0` and therefore a non-reasoning model. Documenting the reasoning-model constraints once in `user_guide_rag.rst` and leaving a rationale comment at each call site is what stops the next bulk model bump from reintroducing this.

Worth calling out for anyone reading later: the empty-content failure mode is the nasty one. Reasoning tokens billing against `max_completion_tokens` means a too-small budget returns HTTP 200 with an empty string, and since `OpenAILLM` discards `finish_reason` it surfaces as `LLMGenerationError: LLM response is not valid JSON` — pointing at JSON parsing rather than the token budget. Surfacing `finish_reason` might be worth a follow-up issue, but it's out of scope here.

Approving — merge is yours.

PR #592, opened 2026-08-10
fix(llm): repair CohereLLM and MistralAILLM, and refresh retired model references

williedoran-neo4j:
Linear: [GENKGB-1858](https://linear.app/neo4j/issue/GENKGB-1858/coherellm-and-mistralaillm-are-unusable-and-five-providers-examples)

Found by running every non-OpenAI example against the live provider APIs. **5 of 26 provider examples failed, none of them for want of credentials** — they're library defects and retired models.

## Library defects

**`CohereLLM` cannot be constructed at all.**

```
AttributeError: No core found in _dynamic_imports for module name -> cohere
```

It reads `cohere.core.api_error.ApiError` as an attribute chain, but the top-level `cohere` package resolves attributes lazily and doesn't list `core`. The module is present and importable — only the attribute access fails — so the fix is one line. Verified on both the pinned 5.20.1 and the 7.0.8 renovate proposes. Both Cohere examples were dead on arrival for any new user.

**`CohereEmbeddings` omits `input_type`**, which current Cohere models require: `invalid request: valid input_type must be provided with the provided model`.

Cohere's embeddings are asymmetric, and this library exposes a single `embed_query()` used for both sides — retrievers embed a query, `TextChunkEmbedder` embeds document chunks — so no single value is right everywhere. Defaulted to `search_document`, what the indexing path needs, because that's the expensive side to get wrong: a mismatch there means re-embedding the corpus.

Two details worth being precise about:

- **A retriever cannot override this per call.** `Embedder.embed_query` is abstract as `(self, text: str)` with no `**kwargs`, and all eight call sites — the vector and hybrid retrievers, the three external stores, and `TextChunkEmbedder` — pass the text alone. The remedy is a second instance, `CohereEmbeddings(input_type="search_query")`, for the retrieval side. The docstring, the CHANGELOG and `user_guide_rag.rst` all say that rather than promising an override that does not exist.
- **The new argument is keyword-only.** Added positionally it would have displaced `rate_limit_handler`, so an existing `CohereEmbeddings(model, handler)` call would silently bind the handler to `input_type` — dropping it and passing a `RateLimitHandler` to the API. Keyword-only keeps the `(model, rate_limit_handler, **kwargs)` order the other embedders share.

**`MistralAILLM` cannot be closed** — `AttributeError: 'Mistral' object has no attribute 'close'` on any use as a context manager, because the SDK replaced `close()`/`aclose()` with the context-manager protocol.

Fixing that meant taking the `mistralai` 2.x upgrade rather than patching around it, and **merging renovate's pending bump untouched would have broken MistralAI outright rather than fixing it**: in 2.x the top-level package became a namespace with no exports, so `from mistralai import Mistral` fails, everything moved under `mistralai.client`, and `Messages` was renamed `ChatCompletionRequestMessage`.

Two notes on that bump:

- Renovate proposes `mistralai>=2.8.0,<2.9.0`, which **cannot resolve** — `[tool.uv] exclude-newer` caps the index at 2026-07-23 and 2.8.0 shipped on the 28th. Pinned `>=2.7.1`, same v2 layout. The same collision applies to the pending cohere v7 and pinecone-client v6 bumps; worth addressing separately rather than here.
- v2's stricter typing surfaced a latent bug — a choice can arrive with no message, for a filtered or truncated completion — which all four response-parsing sites assumed otherwise.

## Retired models

Each verified against the live provider before and after:

| example | was | now |
| --- | --- | --- |
| `anthropic_llm.py` | `claude-3-opus-20240229` (404) | `claude-sonnet-4-5` |
| `cohere_llm.py` | `command-r` (removed 2025-09-15) | `command-a-03-2025` |
| `google_genai_llm.py` | `gemini-2.5-flash` ("no longer available to new users") | `gemini-flash-latest` |
| `bedrock_llm.py` | `…claude-sonnet-4-20250514-v1:0` ("marked as Legacy") | `…claude-haiku-4-5-20251001-v1:0` |
| `vertexai_llm.py`, `vertexai_tool_calls.py` | `gemini-2.0-flash-001` (404) | `gemini-2.5-flash` |

**Two library defaults had rotted the same way**, so calling either class with no arguments failed for everyone:

- `VertexAILLM`'s default, `gemini-1.5-flash-001`, 404s. This resolves item 2 of GENKGB-1825; that ticket's SDK-migration and model-standardisation questions stay open.
- `BedrockLLM`'s default was the same sonnet-4 id Bedrock now refuses as Legacy.

The retired references left in `docs/source/user_guide_rag.rst` and `docs/source/llm.rst` are refreshed too — a branch about retired models should not have skipped the user guide.

`GeminiLLM`'s default is left alone deliberately: `gemini-2.0-flash` can't be called by a free-tier AI Studio key (quota `limit: 0`), but paid keys are fine, so that's a quota boundary rather than a retired model.

## Two examples that broke themselves

`anthropic_llm.py` and `google_genai_llm.py` each built a *second* client pointed at `https://my-custom-endpoint.example.com` to demonstrate `base_url`, then called it. That host doesn't resolve, so both died on DNS right after their real call had already succeeded — the feature being demonstrated was what broke the example. Now an optional `--base-url`, exercised only when supplied and documented in each example's own docstring.

`custom_embeddings.py` passed `dimensions=` to a class taking `dimension=`, so the stray keyword fell through `**kwargs` into `Embedder.__init__()` and raised `TypeError`.

## Why CI didn't catch any of it

- The e2e suite mocks the LLMs and uses local embedders, so no test constructs a real provider client.
- **`tests/unit/llm/test_cohere_llm.py` imports `cohere.core` at module level** — which populates `core` on the parent package, exactly the condition the broken code depended on. The test suite created the situation that hid the bug. The regression test for it runs in a subprocess, and I confirmed it fails against the old code before keeping it.
- Model references are plain strings: nothing type-checks them and no job calls the provider, so five providers rotted simultaneously and silently.

## Breaking changes

Three, all marked in `CHANGELOG.md` under `### Changed`:

- The `mistralai` extra now requires 2.x — users pinning 1.x must upgrade.
- `VertexAILLM`'s default model changed, so callers relying on the old default reach a different model. Pass `model_name` to pin one.
- `BedrockLLM`'s default model changed, for the same reason and with the same remedy.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [x] Breaking change
- [x] Documentation update

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

Every fix confirmed against the live provider, not just unit tests: Cohere LLM and embeddings, MistralAI sync and async invoke plus both context managers, `VertexAILLM()` with no arguments, and each replaced model. Regression tests added for the Cohere import, the three `input_type` paths, the Mistral close protocol, the Vertex default pinned against its own docstring, and the `CohereEmbeddings` argument order — the last confirmed to fail if `input_type` is moved back to a positional parameter.

Green on `ruff check`, `ruff format --check`, `mypy --strict`, and 1232 unit tests.

# Checklist


src/neo4j_graphrag/llm/mistralai_llm.py
```diff
@@ -160,7 +165,10 @@ def __invoke_v1(
             content: str = ""
             usage = None
             if response and response.choices:
-                possible_content = response.choices[0].message.content
+                # mistralai 2.x types `message` as optional, so a choice can
+                # arrive without one - a filtered or truncated completion.
+                message = response.choices[0].message
+                possible_content = message.content if message else None
                 if isinstance(possible_content, str):
                     content = possible_content
```
oskarhane: This block is now 4x identical, comment included: `__invoke_v1` (168-173), `__invoke_v2` (216-221), `__ainvoke_v1` (266-271), `__ainvoke_v2` (317-322). The `response.usage` -> `LLMUsage` block right below it is identical in all four too.

Could collapse to one static helper, no instance state needed:

```python
@staticmethod
def _parse_response(response: Any) -> tuple[str, Optional[LLMUsage]]:
    content = ""
    usage = None
    if response and response.choices:
        # mistralai 2.x types `message` as optional, so a choice can
        # arrive without one - a filtered or truncated completion.
        message = response.choices[0].message
        possible_content = message.content if message else None
        if isinstance(possible_content, str):
            content = possible_content
    if response and response.usage:
        usage = LLMUsage(
            request_tokens=response.usage.prompt_tokens,
            response_tokens=response.usage.completion_tokens,
            total_tokens=response.usage.total_tokens,
        )
    return content, usage
```

Each of the four then becomes `content, usage = self._parse_response(response)`. Drops ~20 lines and the 4 copies of the comment.

Nice-to-have, not blocking.

oskarhane (approved):
LGTM.

Good call on the subprocess test for `CohereLLM` construction — a mock-based test would have sailed straight past that one. Same for pinning the VertexAI default against the docstring; that's the right response to a default that fails silently. The migration reasoning is in comments where it needs to be.

One inline note on duplication in `mistralai_llm.py`. Nice-to-have, not blocking — your call whether it's worth it here.

williedoran-neo4j:
Thanks @oskarhane — took the refactor, and rebased onto current `main`.

`_parse_response` is now a single static helper with four call sites, 32 lines lighter, exactly as suggested. No behaviour change; the four paths differ only in how they call the SDK.

Worth flagging one thing, because it is a good argument for the tests that exist here: my first attempt inserted the helper between `__invoke_v1` and its `@rate_limit_handler_decorator`, so the helper silently absorbed the decorator and `__invoke_v1` lost it. `ruff` and `mypy` were both clean on that version — `test_mistralai_invoke_v2_rate_limit_handler_called` and its siblings are what caught it. Fixed, and the decorators are back on their own methods.

Green on `ruff`, `mypy --strict` and 1241 unit tests.

PR #595, opened 2026-08-11
build(compose): put the vector stores behind a profile and add healthchecks

williedoran-neo4j:
First of the PRs replacing #591, which was too large to review as one change.

`tests/e2e/docker-compose.yml` starts Neo4j, Weaviate, its vectorizer and Qdrant unconditionally. Anyone who only wants Neo4j — which is most of the examples, and much of local development — pays for four containers and a model download.

## What changed

- The vector stores move behind a **`vectordb` profile**. The default is Neo4j with APOC; `--profile vectordb` adds the rest.
- **Healthchecks**, so `up --wait` blocks until the services actually answer rather than merely start. CI currently polls readiness by hand for want of them.
- **Pinecone Local**, an in-memory emulator that ignores API keys, so the Pinecone examples no longer need a hosted account.

The one service without a healthcheck is `pinecone-local`: its image ships no shell and no binary a healthcheck could run. I verified that by probing the image rather than assuming it, and the file header says so.

The `t2v-transformers` healthcheck uses `python3` because that image has neither `wget` nor `curl` — also verified by running it; its ready endpoint answers 204, which `urlopen` treats as success.

## Callers updated

No workflow references this file — only docs do — so the blast radius is documentation:

- `README.md` and `docs/source/index.rst` (the e2e instructions now pass `--profile vectordb`)
- the Qdrant and Weaviate per-store example READMEs

**Teardown needs `--profile '*'`.** `docker compose down` only removes services in the active profiles, so running it bare after starting `--profile vectordb` leaves four containers up and fails to remove the network with *"Resource is still in use"*. `--profile '*'` matches every profile and tears down whatever you started. The docs say so.

## Verification

Run against the real thing, not just `config`-validated:

- `up -d --wait` on the default profile: **neo4j alone, healthy**, exit 0.
- `up -d --wait --profile vectordb`: **all five healthy in ~11s**, exit 0. `pinecone-local` reports no `(healthy)` marker, as expected for a container with no healthcheck.
- Each service probed functionally rather than trusted: Neo4j over bolt with **174 APOC procedures** resolved, Weaviate `/v1/.well-known/ready` 200, Qdrant 200, Pinecone Local 200.
- **The full e2e suite against this stack: 101 passed, 1 skipped** (excluding the SEARCH-clause tests, which use the separate Neo4j 2026 compose file).
- `--profile '*' down` leaves zero containers and removes the network.

The Neo4j healthcheck was the one real risk — it was written against `neo4j:5-enterprise` and is applied here to this file's incumbent `neo4j:enterprise`. Confirmed `wget` is present and the check passes on that image.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [x] E2E tests
- [x] Manual tests

# Checklist

- [x] E2E tests have been updated (run against the new stack; no test changes needed)

### DIFF
--- a/tests/e2e/docker-compose.yml
+++ b/tests/e2e/docker-compose.yml
@@ -1,6 +1,39 @@
 ---
+# Local services for the e2e tests and for running the examples.
+#
+#   docker compose -f tests/e2e/docker-compose.yml up -d --wait
+#       Neo4j with APOC. Enough for the e2e suite's Neo4j tests and for most
+#       examples.
+#
+#   docker compose -f tests/e2e/docker-compose.yml --profile vectordb up -d --wait
+#       Adds Weaviate, Qdrant and Pinecone Local. Needed by the external
+#       retriever tests and the examples under
+#       examples/customize/retrievers/external/.
+#
+# Every service except pinecone-local declares a healthcheck, so `up --wait`
+# blocks until they actually answer rather than merely start. Pinecone Local
+# ships no shell and no binary a healthcheck could run, so `--wait` returns as
+# soon as its container is up; give it a moment before using it.
 services:
+  neo4j:
+    image: neo4j:enterprise
+    ports:
+      - 7687:7687
+      - 7474:7474
+    environment:
+      NEO4J_AUTH: neo4j/password
+      NEO4J_ACCEPT_LICENSE_AGREEMENT: "yes"
+      NEO4J_PLUGINS: "[\"apoc\"]"
+      NEO4J_server_memory_heap_max__size: 6G
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://localhost:7474 || exit 1"]
+      interval: 5s
+      timeout: 5s
+      retries: 30
+      start_period: 20s
+
   weaviate:
+    profiles: [vectordb]
     command:
       - --host
       - 0.0.0.0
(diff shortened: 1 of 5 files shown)


AmirLayegh (approved):
Nice work!
Left a couple of nits below but nothing blocking.

tests/e2e/docker-compose.yml
```diff
@@ -1,6 +1,39 @@
...
+    ports:
+      - 7687:7687
+      - 7474:7474
+    environment:
+      NEO4J_AUTH: neo4j/password
+      NEO4J_ACCEPT_LICENSE_AGREEMENT: "yes"
+      NEO4J_PLUGINS: "[\"apoc\"]"
+      NEO4J_server_memory_heap_max__size: 6G
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://localhost:7474 || exit 1"]
```
AmirLayegh: Nit: this checks HTTP (7474), but the e2e suite connects over bolt via 7687. They usually come up together, but a bolt-based check would validate what actually matters.

tests/e2e/docker-compose.yml
```diff
@@ -21,21 +54,56 @@ services:
...
+      retries: 60
+      start_period: 30s
+
   qdrant:
+    profiles: [vectordb]
     image: qdrant/qdrant
     ports:
       - 6333:6333
+    healthcheck:
+      test: ["CMD-SHELL", "bash -c ':> /dev/tcp/127.0.0.1/6333' || exit 1"]
```
AmirLayegh: Nit: could we use `http://localhost:6333/healthz` instead, as [qdrant provides health endpoint](https://qdrant.tech/documentation/ops-monitoring/monitoring/)? It also matches the pattern used for Weaviate.

tests/e2e/docker-compose.yml
williedoran-neo4j: Good catch — switched to `cypher-shell -u neo4j -p password 'RETURN 1'`, which ships in the image and defaults to bolt (7687), so it now checks the same protocol the e2e suite actually uses. Verified live: it returns a row with the real credentials and fails with a wrong password.

tests/e2e/docker-compose.yml
williedoran-neo4j: Checked the actual image rather than assuming — it has no `wget`/`curl`/`nc` (open upstream: qdrant/qdrant#3491), so a literal wget-based check would fail on every run. Kept your intent — check `/healthz`, not just that the port is open — by sending a raw HTTP request over `/dev/tcp` inside an explicit `bash -c` (CMD-SHELL runs via /bin/sh, which is dash here and can't do /dev/tcp at all). Verified live with a real `docker compose up --wait`, plus a negative control pointing the probe at a dead port to confirm it actually fails.

PR #596, opened 2026-08-11
docs(examples): document what the examples need in order to run

williedoran-neo4j:
Follow-up to #595, now that the services are easy to start.

`examples/README.md` is a pure link index: no mention of API keys, extras, Docker, or the fact that 17 examples talk to a remote demo database. What setup knowledge exists is spread across the root README's Tests section and three per-store READMEs.

## Documentation

**`examples/SETUP.md`** — what you need in order to run the examples: credentials, the services, which providers are free or have a local equivalent, and the traps that cost the most time. Running from the repo root; the three examples that need `PYTHONPATH=examples/data`; the two that write to a **read-only** demo database and so fail with `Forbidden` as written.

**`examples/.env.example`** — a credential template, with `.env` variants gitignored and the template itself kept tracked.

Every count in SETUP.md was checked against the code rather than estimated — "17 examples use the demo database", "14 call `load_dotenv()`", "three need PYTHONPATH", the two empty splitter files.

## Not documentation — three Ollama examples change behaviour

Flagging this explicitly, because it is code in an otherwise-docs PR.

**They could not be run as shipped.** `ollama_llm.py` and `ollama_embeddings.py` carried a literal `model_name="<model_name>"` placeholder that had to be edited before the example would do anything. Documenting *"first edit this file"* would have been documenting a defect, so the model became a command-line argument instead:

```bash
ollama pull llama3.2
python examples/customize/llms/ollama_llm.py llama3.2   # or omit it for the default
```

Which model you have is a property of your machine, not of the example, which is why it is an argument rather than a new hardcoded constant. Defaults are `llama3.2` for the LLMs and `nomic-embed-text` for the embedder, so all three now run with no arguments after a single `ollama pull`.

**One is a real default change worth calling out:** `ollama_tool_calls.py` was not a placeholder — it hardcoded `mistral:latest`. It now defaults to `llama3.2`, so that one `ollama pull` covers all three examples rather than two models being needed. Both support tool calling; the docstring now says to check the model's Ollama page before substituting one that may not.

That file's run instructions also pointed at `examples/tool_calls/ollama_tool_calls.py`, a path that does not exist. Corrected.

## Packaging

`python-dotenv` and `requests` are now declared in the `examples` extra. Fourteen examples call `load_dotenv()` and `tools_retriever_example.py` calls a web API with `requests`, but neither was declared, so both resolved only transitively — working by accident.

## Also

The Weaviate and Pinecone READMEs gave `python -m` a slash path, which cannot work; `-m` takes a dotted module path. Fixed rather than documented.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [ ] E2E tests
- [x] Manual tests

All three Ollama examples confirmed to parse their arguments and print help. Every command quoted in SETUP.md run as written, and every numeric claim in it re-derived from the code. Green on ruff, mypy --strict and the unit suite.

# Checklist

- [ ] New files have copyright header (Markdown and .env template only)

### DIFF
--- a/examples/SETUP.md
+++ b/examples/SETUP.md
@@ -129,6 +129,6 @@ These are properties of the examples, not of your setup.
 - `customize/llms/llm_with_neo4j_message_history.py` and
   `question_answering/graphrag_with_neo4j_message_history.py` write message history to the
   **read-only** demo database, so they fail with `Forbidden` as written.
-- `customize/build_graph/components/splitters/langhchain_splitter.py` and `llamaindex_splitter.py`
-  are empty files, though `README.md` links to them.
-- `customize/embeddings/azure_openai_embeddings.py` reads no environment variables at all.
+- `customize/embeddings/azure_openai_embeddings.py` reads no environment variables. Its endpoint,
+  key and API version are placeholders in the file itself (`api_key="<my key>"`), so it has to be
+  edited before it will run — and that edit must not be committed.
--- a/examples/SETUP.md
+++ b/examples/SETUP.md
@@ -1,0 +1,134 @@
+# Running the examples
+
+[README.md](README.md) indexes what the examples demonstrate. This file covers what you need
+installed and configured to actually run them.
+
+## Quick start
+
+```bash
+uv sync --all-extras --group dev
+cp examples/.env.example .env          # then add an OPENAI_API_KEY
+docker compose -f tests/e2e/docker-compose.yml up -d --wait
+set -a; source .env; set +a
+python examples/question_answering/graphrag.py
+```
+
+That covers about two thirds of the examples. The rest need another provider's key, a vector
+store, or a local runtime — see below.
+
+## Credentials
+
+**Never commit a credential to this repository.** Keys belong in `.env` at the repo root, which is
+gitignored. `examples/.env.example` is a template of placeholders and is the only env file that is
+tracked.
+
+Only 14 examples call `load_dotenv()`. The rest read the environment directly, so export the file
+before running one:
+
+```bash
+set -a; source .env; set +a
+```
+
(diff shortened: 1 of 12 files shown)


AmirLayegh (approved):
LGTM!
Left a few comments that might be worth addressing before merging.

examples/SETUP.md
```diff
@@ -0,0 +1,134 @@
...
+
+These are properties of the examples, not of your setup.
+
+- `customize/retrievers/text2cypher_custom_prompt.py` declares `(:User)-[:REVIEWED]->(:Movie)` in
+  its schema, which the `recommendations` database does not have.
+- `customize/llms/llm_with_neo4j_message_history.py` and
+  `question_answering/graphrag_with_neo4j_message_history.py` write message history to the
+  **read-only** demo database, so they fail with `Forbidden` as written.
+- `customize/build_graph/components/splitters/langhchain_splitter.py` and `llamaindex_splitter.py`
+  are empty files, though `README.md` links to them.
```
AmirLayegh: Can we drop the README link here?

examples/SETUP.md
```diff
@@ -0,0 +1,134 @@
...
+- `customize/embeddings/azure_openai_embeddings.py` reads no environment variables at all.
```
AmirLayegh: But `azure_openai_embeddings.py` has `api_key="<my key>"` and `api_version="<update version"`.

williedoran-neo4j: fixed

examples/SETUP.md
williedoran-neo4j: fixed up

PR #597, opened 2026-08-11
feat(scripts): model what each example under examples/ requires

williedoran-neo4j:
First of four PRs. The order matches how the tools get used: **work out what each example needs (this) → report what is missing (#599) → walk through fixing it (#602) → run what the machine supports (#598)**.

Nothing knows what an example needs in order to run, so a contributor finds out by running it and reading the traceback. This adds one source of truth the rest of the tooling shares.

## Where requirements come from

- **Packages**, from an AST import scan that maps *library symbols* to extras. An example imports `OpenAILLM`, not `openai`, so the dependency is invisible from its imports alone. The substring scan this replaces treated any file containing the word "google" as a Vertex example and could not tell `OpenAIEmbeddings` from `AzureOpenAIEmbeddings`.
- **Env vars and datastores**, read from `os.getenv` calls and connection URIs in the source.
- **Everything else** — APOC, a pre-existing index, outbound internet, or a component named only in a YAML config — declared per path in `SERVICE_RULES`, because it leaves no trace in the source.

Stdlib-only by design: it has to run before `uv sync`, which is exactly when somebody most needs to be told what to install.

## Keeping the tables honest

The tables are hand-maintained, and a half-finished edit produced a plausible wrong answer rather than an error. Each of these is a one-line test:

- **`SYMBOL_EXTRAS` and `SYMBOL_PROVIDERS` are two halves of one fact.** Their values legitimately differ — `AzureOpenAILLM` needs the `openai` extra but talks to the `azure` provider — so they cannot be merged, but their keys must not drift. A symbol in one and not the other means the doctor installs a package it never asks for a key for, or the reverse.
- **Every service an example needs has a label.** `service_available()` returns `True` for anything it cannot probe, so a typo'd service name would otherwise report as satisfied.
- **`SERVICE_RULES` matches are pinned by count.** Asserting only that a pattern matches *something* protects the nine rules matching exactly one file; the other ten match several, so renaming one of them leaves the pattern non-empty and the requirement silently lost. The count also catches a glob widened until it sweeps in a file that does not want the requirement. Update it deliberately when `examples/` gains or loses a file — that is the moment to confirm the rule still applies.

`SERVICE_RULES` specs are a `TypedDict` rather than `dict[str, object]`. Read back through `spec.get()`, a mistyped key (`"service"`) or a scalar where a list belongs both silently did nothing. mypy already runs over `scripts/` in CI and now rejects both.

## The CI fixes have moved to #610

This PR previously carried two commits fixing a venv-cache failure in CI. At
reviewer request they are now #610, which is based directly on `main` and is
independent of this stack — so it can merge first and on its own.

## Scope

Deliberately just the model and its tests. The model has no consumer until #599 — which is what keeps this small enough to check by reading.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

29 tests over the scan, the per-path rules and the table invariants. Each new invariant was confirmed to fail when the thing it guards is broken — dropping a symbol from one table, and simulating a rename against the rule counts — rather than merely passing today.

`scripts/` sits outside the coverage gate (`[tool.coverage.run] source = ["src"]`), so these tests are additive and cannot move the 90% number. They need one line of wiring, `pythonpath = ["scripts"]`.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,6 +103,9 @@ packages = ["src/neo4j_graphrag"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
+# scripts/ is not a package and is not installed. Put it on the path under the
+# same top-level module names mypy already resolves it under.
+pythonpath = ["scripts"]
 filterwarnings = [
     "",
 ]
--- a/scripts/example_requirements.py
+++ b/scripts/example_requirements.py
@@ -1,0 +1,523 @@
+#  Copyright (c) "Neo4j"
+#  Neo4j Sweden AB [https://neo4j.com]
+#  #
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  #
+#      https://www.apache.org/licenses/LICENSE-2.0
+#  #
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""What each file under ``examples/`` needs in order to run.
+
+One source of truth for the tooling around the examples: what to report as
+missing, what to install, and what is safe to run.
+
+Requirements come from three places, in decreasing order of how well they can be
+inferred:
+
+1. **Packages.** An example rarely imports a provider SDK directly - it imports
+   ``OpenAILLM`` from ``neo4j_graphrag.llm`` and the SDK dependency hides behind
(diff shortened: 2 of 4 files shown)


williedoran-neo4j:
Split the CI commits out as requested — they are now #610, based on `main` and independent of this stack, so it can go in first without waiting on anything here.

The two commits moved unchanged (verified byte-identical via `git patch-id --stable`), and the three workflow files on every branch in this stack are now identical to `main`. Rebased #599, #602 and #598 on top; while doing so I found each of them was carrying its own duplicate copy of the same two CI commits, which are now gone too.

This PR is back to just the model and its tests: 2 commits, `scripts/`, `tests/` and one line of `pyproject.toml`.

PR #600, opened 2026-08-11
Fix e2e tests by waiting for indexes to come online

AmirLayegh:
# Description

Fixes a flaky e2e test: `test_graphrag_v2_e2e.py::test_graphrag_v2_happy_path` intermittently fails with
`Neo.ClientError.Schema.IndexNotFound` — `51N63: Index 'vector-index-name' is not ready yet.`

**Root cause:** `populate_neo4j` in `tests/e2e/utils.py` inserts the Question nodes first, then creates the
vector index over them. `CREATE VECTOR INDEX` returns as soon as the schema transaction commits, but
population runs in the background, so the index is still in `POPULATING` state. The first query against it —
the first test in the module — fails. Nothing waited for the index to reach `ONLINE`. The same latent race
exists in `test_graphrag_e2e.py`, which uses an identical fixture and has just been getting lucky.

**Fix:** added a shared `await_index_online()` helper to `tests/e2e/utils.py` that polls
`SHOW INDEXES ... WHERE name = $name` until the index reports `ONLINE`, raising after a 60s timeout. It is
now called at every point an e2e fixture creates an index:

- `populate_neo4j` (the failing path)
- `setup_neo4j_for_retrieval` - vector + fulltext indexes
- `setup_neo4j_for_kg_construction`

`test_search_clause_e2e.py` already contained this exact wait loop inline; it now uses the shared helper, so
the polling query is unchanged from code already passing in CI.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [ ] Unit tests
- [x] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/tests/e2e/utils.py
+++ b/tests/e2e/utils.py
@@ -17,12 +17,39 @@ from __future__ import annotations
 import hashlib
 import json
 import os.path
+import time
 from typing import Any, Literal
 
 import neo4j
 from neo4j_graphrag.indexes import create_vector_index, drop_index_if_exists
 
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+def await_index_online(
+    neo4j_driver: neo4j.Driver, index_name: str, timeout: int = 60
+) -> None:
+    """Block until ``index_name`` reaches the ONLINE state.
+
+    Index population is asynchronous: creating an index over existing data
+    returns before the index is queryable, and querying it while it is still
+    POPULATING fails with ``Neo.ClientError.Schema.IndexNotFound`` (51N63).
+    Any fixture that creates an index must wait for it before running queries.
+    """
+    deadline = time.monotonic() + timeout
+    while True:
+        records, _, _ = neo4j_driver.execute_query(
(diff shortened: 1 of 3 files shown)


tests/e2e/utils.py
```diff
@@ -17,13 +17,40 @@
...
 
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 
+
+def await_index_online(
+    neo4j_driver: neo4j.Driver, index_name: str, timeout: int = 60
+) -> None:
+    """Block until ``index_name`` reaches the ONLINE state.
+
+    Index population is asynchronous: creating an index over existing data
```
williedoran-neo4j: minor nit: 
maybe move the doc string  to a comment explanation to the call site. This function is not specific to the index population anymore

AmirLayegh: Agree - moved the doc string to comment.

williedoran-neo4j (approved):

PR #605, opened 2026-08-16
fix: recognize SEARCH clause support on Aura 5.27+

### LINKED ISSUE #601: [BUG]: VectorCypherRetriever uses deprecated db.index.vector.queryNodes on Aura 5.27 even though SEARCH works
### Before You Report a Bug, Please Confirm You Have Done The Following...

- [x] I have updated to the latest version of the packages.
- [x] I have searched for both [existing issues](https://github.com/neo4j/neo4j-graphrag-python/issues) and [closed issues](https://github.com/neo4j/neo4j-graphrag-python/issues?q=is%3Aissue+is%3Aclosed) and found none that matched my issue.

### neo4j-graphrag-python's version

1.18.0

### Python version

3.13.12

### Operating System

Windows

### Dependencies

annotated-types==0.8.0
anyio==4.14.2
certifi==2026.7.22
charset-normalizer==3.4.9
colorama==0.4.6
distro==1.9.0
fsspec==2026.7.0
h11==0.16.0
httpcore==1.0.9
httpcore2==2.10.0
httpx==0.28.1
httpx2==2.10.0
idna==3.18
jiter==0.16.0
json_repair==0.63.0
jsonpatch==1.33
jsonpointer==3.1.1
langchain==1.3.15
langchain-core==1.5.4
langchain-openai==1.4.3
langchain-protocol==0.0.18
langgraph==1.2.11
langgraph-checkpoint==4.2.0
langgraph-prebuilt==1.1.0
langgraph-sdk==0.4.2
langsmith==0.10.18
markdown-it-py==4.2.0
mdurl==0.1.2
neo4j==6.2.0
neo4j-graphrag==1.18.0
numpy==2.5.2
openai==2.54.0
orjson==3.11.9
ormsgpack==1.12.2
packaging==26.3
pydantic==2.13.4
pydantic_core==2.46.4
Pygments==2.20.0
pypdf==6.15.0
python-dotenv==1.2.2
pytz==2026.3.post1
PyYAML==6.0.3
regex==2026.7.19
requests==2.34.2
requests-toolbelt==1.0.0
rich==15.0.0
scipy==1.18.0
sniffio==1.3.1
tenacity==9.1.4
tiktoken==0.13.0
tqdm==4.70.0
truststore==0.10.4
types-PyYAML==6.0.12.20260724
typing-inspection==0.4.3
typing_extensions==4.16.0
urllib3==2.7.0
uuid_utils==0.17.0
websockets==15.0.1
xxhash==3.8.1
zstandard==0.25.0

### Reproducible example

```Python
// Create graph

DROP INDEX chunkEmbedding IF EXISTS;
MATCH (c:Chunk) DETACH DELETE c;

CREATE (c:Chunk {
text: "minimal repro chunk",
embedding: [0.1, 0.2, 0.3]
});

CALL db.index.vector.createNodeIndex(
"chunkEmbedding",
"Chunk",
"embedding",
3,
"cosine"
);

# Python code
import os
from dotenv import load_dotenv
from neo4j import GraphDatabase
from neo4j_graphrag.retrievers import VectorCypherRetriever

load_dotenv()

driver = GraphDatabase.driver(
    os.getenv("NEO4J_URI"),
    auth=(os.getenv("NEO4J_USERNAME"), os.getenv("NEO4J_PASSWORD")),
)

retriever = VectorCypherRetriever(
    driver=driver,
    index_name="chunkEmbedding",
    retrieval_query="RETURN node.text AS text, score",
    neo4j_database=os.getenv("NEO4J_DATABASE"),
)

result = retriever.search(
    query_vector=[0.1, 0.2, 0.3],
    top_k=1,
)
```

### Relevant Log Output

Runtime warning from server:

Received notification from DBMS server: warn: feature deprecated with replacement.
db.index.vector.queryNodes is deprecated. It is replaced by SEARCH.

Generated query starts with:

CALL db.index.vector.queryNodes($vector_index_name, $top_k * $effective_search_ratio, $query_vector)

### Expected Result

When SEARCH is available on the connected database, VectorCypherRetriever should use SEARCH instead of db.index.vector.queryNodes, or at least avoid triggering a deprecation warning on normal retrieval paths.

### What happened instead?

The retriever uses db.index.vector.queryNodes and Aura returns a deprecation warning on each retrieval query.

### Additional Info

Connected DB reports:

- Neo4j Kernel
- 5.27-aura
- enterprise

This suggests the fallback/version gating used by the retriever may be too strict for Aura 5.27, or not aligned with actual SEARCH availability on Aura.

Aryan-Pardeshi:
## Problem

On Aura instances reporting `5.27-aura`, `VectorCypherRetriever` (and the other vector retrievers) still route through `db.index.vector.queryNodes` and trigger its deprecation warning on every retrieval (#601), even though those instances support the Cypher 25 `SEARCH` clause.

## Root cause

`supports_search_clause` compared every server against `(2026, 1, 0)`. Aura keeps a `5.x` version scheme and pins `dbms.components()` to `5.27.0` for clients that parse SemVer, regardless of the underlying engine. The previous `>= (5, 27, 0)` check therefore incorrectly matched year-based Aura strings such as `2025.x-aura`.

## Change

`supports_search_clause` now uses an exact `5.27.0` match for Aura instances and retains the existing year-based comparison for self-managed servers.

## Testing

- `TestSupportsSearchClause`: removed the obsolete `5.28-aura` case, added the `2025.x-aura` regression case, and retained the `5.27-aura` positive case.
- `uv run python -m pytest tests/unit/utils/test_version_utils.py -q`: 40 passed.
- `uv run ruff check` and `uv run ruff format --check`: passed.
- `uv run mypy`: no issues found in 343 source files.
- The full unit suite reaches the same unrelated, pre-existing `test_pipeline_draw` failure observed on clean `main` on Windows; the focused tests, lint, and mypy checks pass.

Closes #601

### DIFF
--- a/src/neo4j_graphrag/utils/version_utils.py
+++ b/src/neo4j_graphrag/utils/version_utils.py
@@ -156,20 +156,27 @@ def clear_version_cache() -> None:
 def supports_search_clause(
     driver: neo4j.Driver, database: Optional[str] = None
 ) -> bool:
-    """Check if the Neo4j server supports the SEARCH clause (>= 2026.01).
+    """Check if the Neo4j server supports the SEARCH clause.
+
+    Self-managed servers report year-based versions and support SEARCH from
+    2026.01. Aura instances on the March 2026 release line and later GA'd the
+    same Cypher 25 SEARCH clause while still reporting a 5.x version string
+    (e.g. "5.27-aura"), so Aura 5.27+ counts as SEARCH-capable. An over-eager
+    guess is safe: retrievers catch the resulting ClientError and fall back to
+    the procedure-based path.
 
     Uses cached version detection. On connection errors, returns False
     so callers fall back to the procedure-based path.
 
     Args:
-        driver: Neo4j Python driver instance.
-        database: Optional database name.
+        driver (neo4j.Driver): The Neo4j Python driver instance.
+        database (str, optional): The name of the Neo4j database.
 
     Returns:
         True if SEARCH clause is supported, False otherwise.
     """
     try:
-        version_tuple, _, _ = get_version_cached(driver, database)
+        version_tuple, is_aura, _ = get_version_cached(driver, database)
     except Exception:
         logger.debug(
             "Failed to detect Neo4j version for SEARCH clause support, "
@@ -177,6 +184,8 @@ def supports_search_clause(
             exc_info=True,
         )
         return False
+    if is_aura:
+        return version_tuple >= (5, 27, 0)
     return version_tuple >= (2026, 1, 0)
 
 
--- a/tests/unit/utils/test_version_utils.py
+++ b/tests/unit/utils/test_version_utils.py
@@ -179,6 +179,13 @@ class TestSupportsSearchClause:
             ("2025.12.0", False),
             ("5.23.0", False),
             ("5.26.0", False),
+            # Aura kept the 5.x version scheme after on-prem moved to
+            # year-based versions; the Aura March release line (5.27-aura+)
+            # GA'd the Cypher 25 SEARCH clause for vector queries (#601).
+            ("5.27-aura", True),
+            ("5.28-aura", True),
+            ("5.26.0-aura", False),
+            ("5.27.0", False),
         ],
     )
     def test_version_check(


src/neo4j_graphrag/utils/version_utils.py
```diff
@@ -156,27 +156,36 @@ def clear_version_cache() -> None:
...
-        version_tuple, _, _ = get_version_cached(driver, database)
+        version_tuple, is_aura, _ = get_version_cached(driver, database)
     except Exception:
         logger.debug(
             "Failed to detect Neo4j version for SEARCH clause support, "
             "falling back to procedure path.",
             exc_info=True,
         )
         return False
+    if is_aura:
```
NathalieCharbel: Aura reports CalVer strings too, so every Aura CalVer version below `2026.01` would evaluate to `True` (e.g., `(2025, 1, 0) >= (5, 27, 0)`) and now reports `SEARCH` support which is not the correct behaviour.
so let's scope the override to the SemVer scheme

```suggestion
    if is_aura and version_tuple[0] == 5:
```

src/neo4j_graphrag/utils/version_utils.py
```diff
@@ -156,27 +156,36 @@ def clear_version_cache() -> None:
...
+        return version_tuple >= (5, 27, 0)
```
NathalieCharbel, line 188: I am afraid it passes for `2025.x` too. In fact, Aura didn't keep a `5.x` line. Since the March 2025 release it pins `dbms.components()` to `5.27.0` for clients that parse SemVer - a frozen constant that every Aura instance reports regardless of the engine it runs.

A safer check would be:
```
if is_aura and version_tuple == (5, 27, 0):
   return True
```

tests/unit/utils/test_version_utils.py
```diff
@@ -179,6 +179,13 @@ def setup_method(self) -> None:
             ("2025.12.0", False),
             ("5.23.0", False),
             ("5.26.0", False),
+            # Aura kept the 5.x version scheme after on-prem moved to
+            # year-based versions; the Aura March release line (5.27-aura+)
+            # GA'd the Cypher 25 SEARCH clause for vector queries (#601).
+            ("5.27-aura", True),
+            ("5.28-aura", True),
```
NathalieCharbel, line 186: I think this check for `5.28-aura` can be removed

NathalieCharbel (changes requested):
Thank you @Aryan-Pardeshi for the fix!

A few comments that I left below worth to be addressed before we merge it. 
In general, the proposed fix is one direction. I am afraid the version check for aura is not reliable for checking features support as stated in one of my comments. A more reliable check would be a `CYPHER 25 EXPLAIN` of the `SEARCH` query either returning a plan or not, but this can be done in a follow up PR. 

Also, can you add a line in CHANGELOG about the fix?

PR #610, opened 2026-08-17
ci: stop a half-installed venv cache from failing later runs

williedoran-neo4j:
## What

Two fixes to how CI caches the virtualenv. Independent of everything else — it
touches only `.github/workflows/`, and applies to `main` on its own.

Split out of #597 at reviewer request; the commits are unchanged.

## Why

A cache entry saved from an interrupted install is restored on every later run
with the same key, and the `uv sync` that would repair it was skipped precisely
because the key hit. Seen on #596, which changed `uv.lock` and so minted a fresh
key: four e2e jobs failed with `ImportError: cannot import name
'WeaviateNeo4jRetriever'`, and the same for Pinecone and Qdrant. The restored
`.venv` was missing the vector-store extras.

It is nastier than its cause suggests — it presents as a flaky test in code that
did not change, it survives re-runs, and it persists until the cache expires or
somebody deletes the entry by hand. Any PR touching `uv.lock` mints the new key
and so is first to hit it, which lands it on whoever is least likely to suspect
CI. This one cost a while to diagnose on #596; the entry had to be removed with
`gh cache delete`.

## The two changes

1. **Install unconditionally.** `uv sync` no longer skipped on an exact cache
   hit. It takes a few seconds when the environment is already correct, which is
   cheap next to the above.

2. **Never publish a half-installed venv.** `actions/cache` saves in a post step
   that runs even when the job was cancelled or failed. Replaced with
   `actions/cache/restore` + `actions/cache/save`, where the save is conditional
   on a cache miss — so a cancelled job cannot mint a bad entry, and an existing
   good entry is not overwritten.

Same pinned action SHA, split across the two entry points.

Applied to all three workflows: `pr.yaml`, `pr-e2e-tests.yaml`,
`scheduled-e2e-tests.yaml`.

## Verification

CI on this PR exercises the changed `pr.yaml` directly. Locally: `ruff check`,
`ruff format --check`, `mypy .`, and `pytest tests/unit` (1241 passed) all pass
on this branch standalone.

### DIFF
--- a/.github/workflows/pr-e2e-tests.yaml
+++ b/.github/workflows/pr-e2e-tests.yaml
@@ -89,19 +89,29 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Set up uv
         uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
-      - name: Load cached venv
+      # restore/save rather than the combined cache action, so the save can be
+      # made conditional. The combined action saves in a post step that also runs
+      # when the job was cancelled or failed, which publishes a half-installed
+      # .venv under a key derived from uv.lock alone. Every later run with that
+      # lock then restores it, and an exact key hit suppresses the save, so the
+      # bad entry is never replaced - it has to be deleted by hand.
+      - name: Restore cached venv
         id: cached-venv
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: .venv
           key: ${{ runner.os }}-venv-${{ matrix.python-version }}-${{ hashFiles('**/uv.lock') }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-e2e-tests.yaml
```diff
@@ -89,15 +89,29 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Set up uv
         uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
-      - name: Load cached venv
+      # restore/save rather than the combined cache action, so the save can be
+      # made conditional. The combined action saves in a post step that also runs
+      # when the job was cancelled or failed, which publishes a half-installed
```
NathalieCharbel: is this describing the old behaviour since we now save only on success

NathalieCharbel (approved):
LGTM

Just a minor suggestion to remove comments explaining existing behaviour as it may be confusing with the actual code.

.github/workflows/pr-e2e-tests.yaml
williedoran-neo4j: good catch. will update

PR #618, opened 2026-08-24
Add lazily-evaluated pipeline DSL and local interpreter

jonnylaw:
# Description

Adds a lazily-evaluated dataflow pipeline DSL (`neo4j_graphrag.pipeline`) plus a local interpreter to run it.

A `Pipeline` is **pure data** — building one produces an operator graph and executes nothing. Evaluation is deferred to an `Interpreter`, triggered by `.collect()`, `.to_sink()`, iteration, or by passing an interpreter explicitly. This separation means the same pipeline definition can later be run by alternative interpreters (batched, distributed, instrumented) without touching the definition site.

```python
from neo4j_graphrag.pipeline import Pipeline

results = (
    Pipeline.from_source(my_source)
    .map(transform)
    .flat_map(expand)
    .map_async_chunked(async_transform)
    .reduce(zero=identity, combine=merge)
    .collect()
)
```

## What's included

- **`pipeline.py`** — the `Pipeline` / `ResultPipeline` builder API.
- **`operators.py`** — the operator graph: `Map`, `FlatMap`, `Filter`, `Take`, `TakeWhile`, `Skip`, `Grouped`, `ReduceOp`, `MapAsyncChunked`, the `Try*` error-capturing variants, the `*Ok` variants that operate inside a `Result`, plus `Tee` / `Partition` for fan-out.
- **`interpreter.py`** — `Interpreter` protocol and `LocalInterpreter`, which evaluates a graph in-process with bounded-concurrency chunking for the async operators.
- **`result.py`** — `Result` / `Ok` / `Err`, so a failing element becomes a value in the stream rather than an exception that tears down the whole run.
- **`source.py`** / **`sink.py`** — the `Source` and `Sink` boundary protocols.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

Purely additive — a new subpackage with no changes to existing modules, so nothing currently importable changes behaviour.

## Complexity

Complexity: High — new subsystem, ~1,500 lines of implementation and a fair amount of design surface in the operator set.

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

`tests/unit/pipeline/test_pipeline.py` (~870 lines) covers the builder, each operator, laziness/short-circuiting, error propagation through the `Try*` and `*Ok` variants, and the tee/partition fan-out paths.

# Checklist

- [ ] Documentation has been updated — module and method docstrings are in place, but there's no `docs/` entry yet
- [ ] E2E tests have been updated — not applicable, no I/O in this PR
- [ ] Examples have been updated — a `SimpleKGBuilder` built on this DSL is the planned follow-up PR

## Note on review order

This is the first of two changes. The follow-up rewrites `SimpleKGBuilder` on top of this DSL and adds a worked example; it's parked on `pipeline-dsl-example` until this lands. Reviewing the DSL on its own first is the intent — happy to open the second as a draft now if seeing the consumer helps judge the API.

### DIFF
--- a/src/neo4j_graphrag/pipeline/interpreter.py
+++ b/src/neo4j_graphrag/pipeline/interpreter.py
@@ -76,6 +76,8 @@ from neo4j_graphrag.pipeline.operators import (
     TryMapOkAsyncChunked,
 )
 from neo4j_graphrag.pipeline.result import Err, Ok
+from neo4j_graphrag.pipeline.sink import Sink
+from neo4j_graphrag.pipeline.source import Source
 
 if TYPE_CHECKING:
     from neo4j_graphrag.pipeline.pipeline import Pipeline, ResultPipeline
@@ -98,8 +100,10 @@ class Interpreter(ABC):
     def evaluate(self, pipeline: Pipeline[Any] | ResultPipeline[Any]) -> Iterator[Any]:
         """Evaluate *pipeline* and return its output stream.
 
-        If the pipeline ends in a sink, the stream is consumed, written to
-        the sink, and an empty iterator is returned.
+        Implementations should not execute any user code before the
+        returned stream is consumed.  A pipeline ending in a sink returns a
+        stream that yields nothing but writes to the sink as it is drained,
+        so callers must exhaust it — :meth:`Pipeline.to_sink` does this.
         """
 
 
@@ -277,6 +281,34 @@ def _try_flat_map_ok(
                 yield Err(exception=e)
 
 
+def _read_source(source: Source[Any]) -> Iterator[Any]:
+    """Read *source*, deferred until the stream is consumed.
+
+    ``read()`` may open a file or a connection, so it must not run while
+    the generator chain is merely being built.
+    """
+    yield from source.read()
+
+
+def _reduce_lazy(
+    stream: Iterator[Any], zero: Any, combine: Callable[[Any, Any], Any]
+) -> Iterator[Any]:
+    """Fold *stream* into a single element, deferred until consumed.
+
+    The fold itself is not incremental — it drains the upstream — but it
+    does so on first ``next()`` rather than when the chain is built, so
+    ``evaluate()`` stays free of side effects.
+    """
+    yield _reduce(combine, stream, zero)
+
+
+def _to_sink(stream: Iterator[Any], sink: Sink[Any]) -> Iterator[Any]:
+    """Write every element of *stream* to *sink*, yielding nothing."""
+    for item in stream:
+        sink.write(item)
+    yield from ()
+
+
 def _on_error(stream: Iterator[Any], handler: Callable[[Err], None]) -> Iterator[Any]:
     for item in stream:
         if isinstance(item, Err):
@@ -293,10 +325,12 @@ def _on_error(stream: Iterator[Any], handler: Callable[[Err], None]) -> Iterator
 class LocalInterpreter(Interpreter):
     """Evaluate a pipeline in the current process with lazy generators.
 
-    Folds the operator list into a single generator chain.  Apart from
-    ``ReduceOp`` (which folds its upstream eagerly) and the blocking
-    chunked-async operators, no work happens until the returned stream is
-    consumed.
+    Folds the operator list into a single generator chain.  Building the
+    chain runs no user code: every operator — including ``ReduceOp`` and
+    ``SinkOp``, which both drain their upstream — defers its work until the
+    returned stream is consumed.  The chunked-async operators are the one
+    exception to incremental evaluation: each chunk blocks in
+    ``asyncio.run`` while it is being produced.
     """
 
     def evaluate(self, pipeline: Pipeline[Any] | ResultPipeline[Any]) -> Iterator[Any]:
@@ -304,7 +338,7 @@ class LocalInterpreter(Interpreter):
         for op in pipeline.pipeline_operators:
             match op:
                 case SourceOp(source=source):
-                    stream = iter(source.read())
+                    stream = _read_source(source)
                 case Map(func=func):
                     stream = map(func, stream)
                 case FlatMap(func=func):
@@ -320,7 +354,7 @@ class LocalInterpreter(Interpreter):
                 case Grouped(size=size):
                     stream = _batched(stream, size)
                 case ReduceOp(zero=zero, combine=combine):
-                    stream = iter([_reduce(combine, stream, zero)])
+                    stream = _reduce_lazy(stream, zero, combine)
                 case MapAsyncChunked(func=func, map_batch_size=batch_size):
                     stream = _iter_chunked_async(
                         stream, batch_size, _make_map_chunk_coro(func)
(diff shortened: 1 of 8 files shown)


src/neo4j_graphrag/pipeline/pipeline.py
```diff
@@ -0,0 +1,527 @@
...
+    """Adapter wrapping any ``Iterable`` to satisfy the ``Source`` protocol."""
+
+    def __init__(self, iterable: Iterable[Any]) -> None:
+        self._iterable = iterable
+
+    def read(self) -> Iterable[Any]:
+        return self._iterable
+
+
+class Pipeline(Generic[T]):
```
williedoran-neo4j, line 107: Good to know experimental is going away. Until it actually does, we've got two `Pipeline`s with the same name and different semantics. Does the experimental one get deprecated in this PR, or is it already dark enough that nobody imports both under one roof?

src/neo4j_graphrag/pipeline/pipeline.py
```diff
@@ -0,0 +1,527 @@
...
+        """
+        if n < 0:
+            raise ValueError(f"n must be >= 0, got {n!r}")
+        return self._wrap(ops.Skip(prev=self._tail, n=n))
+
+    def filter(self, predicate: Callable[[T], bool]) -> Pipeline[T]:
+        """Keep only elements for which *predicate* is ``True``."""
+        return self._wrap(ops.Filter(prev=self._tail, predicate=predicate))
+
+    def tee(self, n: int = 2) -> tuple[Pipeline[T], ...]:
```
williedoran-neo4j: After both branches drain, `tee` returns `[]` on a second collect, and `partition` replays cached results instead of re-running. A plain chain re-runs fine. Is that deliberate, or just a side effect of the shared state token?

src/neo4j_graphrag/pipeline/operators.py
```diff
@@ -0,0 +1,341 @@
...
+
+    handler: Callable[[Err], None]
+
+
+# ---------------------------------------------------------------------------
+# Branching
+# ---------------------------------------------------------------------------
+
+
+class TeeState:
```
williedoran-neo4j: `TeeState`/`PartitionState` are the only non-pure-data bits and they won't serialize across processes. I'm trying to square that with the "alternative interpreters" pitch — is there a story for these, or is that a later problem?

src/neo4j_graphrag/pipeline/sink.py
```diff
@@ -0,0 +1,50 @@
...
+
+        class InMemorySink:
+            def __init__(self) -> None:
+                self.received: list[Any] = []
+
+            def write(self, element: Any) -> None:
+                self.received.append(element)
+    """
+
+    def write(self, element: T_contra) -> None:
```
williedoran-neo4j, line 57: A failed `write()` kills the whole run (no `Ok`/`Err`), and `to_sink()` executes eagerly in `evaluate()`, which breaks the lazy rule. Was that a deliberate choice to keep the terminal simple?

src/neo4j_graphrag/pipeline/interpreter.py
```diff
@@ -0,0 +1,436 @@
...
+    before moving to the next chunk.
+
+    Raises:
+        ValueError: If *map_batch_size* < 1.  (Builders validate eagerly;
+            this is a backstop for hand-built operator graphs.)
+    """
+    if map_batch_size < 1:
+        raise ValueError(f"map_batch_size must be >= 1, got {map_batch_size!r}")
+    for chunk in _batched(stream, map_batch_size):
+        yield from asyncio.run(make_chunk_coro(chunk))
```
williedoran-neo4j, line 136: Fine for v1 as long as the fix stays a single-loop `AsyncInterpreter` and nobody tweaks the loop juggling.

tests/unit/pipeline/test_pipeline.py
```diff
@@ -0,0 +1,870 @@
...
+    def test_tee_after_map(self) -> None:
+        b1, b2 = Pipeline([1, 2]).map(lambda x: x * 5).tee()
+        assert b1.collect() == [5, 10]
+        assert b2.collect() == [5, 10]
+
+    def test_n_less_than_two_raises(self) -> None:
+        with pytest.raises(ValueError, match="n must be >= 2"):
+            Pipeline([1]).tee(1)
+
+    def test_upstream_is_read_once_across_branches(self) -> None:
```
williedoran-neo4j: tee/partition re-run and `SinkOp` are untested — exactly where I hit unexpected behavior. I'd add a couple to pin the re-run semantics down.

src/neo4j_graphrag/pipeline/pipeline.py
```diff
@@ -0,0 +1,527 @@
...
+            zero: The identity / initial accumulator value.
+            combine: An associative binary callable ``(A, T) → A``.
+        """
+        return self._wrap(ops.ReduceOp(prev=self._tail, zero=zero, combine=combine))
+
+    # ------------------------------------------------------------------
+    # Async operators (blocking under LocalInterpreter)
+    # ------------------------------------------------------------------
+
+    def map_async_chunked(
```
williedoran-neo4j, line 234: `safe × ok × flat` is growing. Before this becomes public API, does collapsing to one method with an `error=` flag make sense? If it's noise, drop it.

src/neo4j_graphrag/pipeline/sink.py
jonnylaw: I made `to_sink` and `reduce` lazy

src/neo4j_graphrag/pipeline/operators.py
jonnylaw: Decided to remove these for now as they're not needed for the pipeline

src/neo4j_graphrag/pipeline/pipeline.py
jonnylaw: I think we're good - we need the full namespace to import it and this is unused in the lib until I add the example in the follow-up PR.

src/neo4j_graphrag/pipeline/pipeline.py
jonnylaw: I added `flatten_ok`, then `map_safe(...).flatten_ok()` is equivalent to `flat_map_safe`.

williedoran-neo4j:
This is a great update to the library. Adding a lazily-evaluated, interpreter-backed pipeline gives us a real foundation for the future,  not just a nicer API, but a definition/evaluation split that leaves the door open to scale this out to genuinely large volumes. 

I dont know if this is the right place for this... please point me to where i can contribute these ideas @stellasia 

A few things I'd love to see us start dreaming about from here (not blockers, just directions):
  - AsyncInterpreter: one event loop across the whole chain, so loop-bound clients (httpx.AsyncClient) span the run instead of per-chunk asyncio.run.
  - Pluggable executor under map_async_chunked, swap the chunk dispatch (event loop → process pool → actor cluster) without touching the operator graph. Most of a KG build is embarrassingly-parallel map, and those stages should scale by "add cores," not "add a cluster."
- Keyed reduce / group-by. ReduceOp is an unkeyed fold today, but entity resolution is a keyed reduction. That's the operator gap that matters once resolution outgrows a single node,  a measured threshold, not a default.

williedoran-neo4j (approved):
my comments were resolved, and added a few nuggets to start thinking about where and how we can use these changes to scale out workflows

PR #621, opened 2026-09-03
fix(llm): populate LLMUsage from Gemini's usage_metadata

matteomedioli:
## What & Why

`BaseGeminiLLM`/`GeminiLLM` built `LLMResponse` from `response.text` alone on all four invoke paths (sync/async, v1/v2), silently discarding `response.usage_metadata` — every gemini call lost token-usage tracking. Every other base class (`BaseAnthropicLLM`/`BaseOpenAILLM`/`BaseVertexAILLM`) already populates `usage`.

Found while migrating kg-builder's gemini route onto `BaseGeminiLLM` (a downstream consumer that tracks token usage on every LLM call for billing/observability) — a naive subclass would have silently regressed usage tracking with no test failure to catch it.

## How

- Added `BaseGeminiLLM._parse_content_response()`, mapping `usage_metadata.{prompt,candidates,total}_token_count` to `LLMUsage`. Used by all four invoke paths (`__invoke_v1`/`__invoke_v2`, sync and async). `usage` is `None` when the API doesn't report `usage_metadata`, matching `LLMUsage`'s optional-fields contract.
- 3 new regression tests: usage on the v1 (str input) path, usage on the v2 (message-list) path, and `usage is None` when `usage_metadata` is absent.

## Testing

```
uv run pytest tests/unit/llm/test_google_genai_llm.py -q
```
26 passed. mypy/ruff clean. Full `tests/unit/llm/` suite unaffected (pre-existing unrelated mistralai env failures only).


AmirLayegh (approved):
LGTM! 🚀

PR #622, opened 2026-09-03
feat(llm): GeminiLLM declares supports_structured_output = True

matteomedioli:
## What & Why

`GeminiLLM` already supports structured output via `response_schema`/`response_mime_type`, but never declared the `supports_structured_output` capability flag. Pipeline components that check the flag (`SimpleKGPipeline`, `SchemaFromTextExtractor`, `LLMEntityRelationExtractor`) saw `False` and silently fell back to prompt-based JSON instead of schema-enforced structured output.

Reimplements the intent of #573 (closed without merging, and now badly stale — 193 files of drift) fresh on current `main`.

**Breaking for pipeline users:** the components above now enable structured output by default with `GeminiLLM`. To keep the previous behavior, construct them with `use_structured_output=False`. Same shape as the existing `AnthropicLLM.supports_structured_output` breaking-change entry in 1.19.0's CHANGELOG.

## How

- Set `supports_structured_output: bool = True` on `BaseGeminiLLM` (same place `BaseVertexAILLM`/`AnthropicLLM` declare it).
- Added a regression test and a CHANGELOG entry.

## Testing

```
uv run pytest tests/unit/ -q --ignore=tests/unit/llm/test_mistralai_llm.py
```
1329 passed — no fallout from the flag flip elsewhere in the suite. mypy/ruff clean.


AmirLayegh (approved):
Nice work! 🚀

tests/unit/llm/test_google_genai_llm.py
```diff
@@ -208,6 +208,16 @@ def test_gemini_llm_is_base_gemini_llm_subclass() -> None:
...
 
 
+def test_gemini_llm_supports_structured_output(
+    mock_genai: Tuple[MagicMock, MagicMock],
+) -> None:
+    """GeminiLLM already supports response_schema/response_mime_type; the
+    capability flag must say so, or SimpleKGPipeline/SchemaFromTextExtractor/
+    LLMEntityRelationExtractor silently fall back to prompt-based JSON."""
+    llm = GeminiLLM("gemini-2.0-flash")
+    assert llm.supports_structured_output is True
```
williedoran-neo4j, line 218: the most LLM test ever... set a default... test the default is default :D

matteomedioli:
postponed for the next release.

PR #624, opened 2026-09-10
Convert `GeoPoint` to a compatible format for parquet file output

adamnsch:
# Description

Fixes [GENKGB-2040](https://linear.app/neo4j/issue/GENKGB-2040/parquet-write-fails-on-graphrag-geopoint-properties): Parquet writes aborted with `ArrowInvalid` when a node or relationship still had a GraphRAG `GeoPoint` at write time. PyArrow cannot infer a type from the pydantic object.

`Neo4jGraphParquetFormatter.format_parquet` now converts `GeoPoint` (and dumped `{latitude, longitude, height}` dicts) to WKT `POINT(longitude latitude height)` before type inference. POINT columns are tagged in Arrow field metadata. Writer file metadata keeps the Parquet source type as `STRING` and sets `target_type: POINT` for Aura import-spec.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Complexity: Low

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [ ] Manual tests

# Checklist

### DIFF
--- a/src/neo4j_graphrag/components/parquet_formatter.py
+++ b/src/neo4j_graphrag/components/parquet_formatter.py
@@ -35,6 +35,7 @@ from neo4j_graphrag.components.schema import (
     GraphSchema,
 )
 from neo4j_graphrag.components.types import (
+    GeoPoint,
     LexicalGraphConfig,
     Neo4jGraph,
     Neo4jNode,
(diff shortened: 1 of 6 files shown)


src/neo4j_graphrag/components/types.py
```diff
@@ -44,6 +44,14 @@ class GeoPoint(BaseModel):
     longitude: float
     height: float
 
+    def to_wkt(self) -> str:
+        """Serialize as WKT ``POINT(longitude latitude height)``.
+
+        Axis order is longitude then latitude (x then y), then height.
+        Height is always included so 3D values round-trip losslessly.
+        """
+        return f"POINT({self.longitude} {self.latitude} {self.height})"
```
matteomedioli, line 53: nit: WKT output is POINT(x y z), standard 3D form is POINT Z(x y z). Tested with shapely, parses fine either way, no data loss. just spec conformance.

src/neo4j_graphrag/components/parquet_formatter.py
```diff
@@ -696,6 +714,26 @@ def _normalize_column_types(rows: list[dict[str, Any]]) -> None:
...
+                if value is None:
+                    continue
+                wkt = _geopoint_like_to_wkt(value)
+                if wkt is not None:
+                    row[key] = wkt
+                    saw_geopoint.add(key)
+                else:
+                    saw_other.add(key)
+        return saw_geopoint - saw_other
+
```
matteomedioli, line 736: nit: a column mixing a geo point with an unrelated dict falls back to plain-text stringification, so output could look confusing in that edge case. No crash.

matteomedioli (approved):
Nice work! LGTM ❇️

src/neo4j_graphrag/components/types.py
adamnsch: Didn't know about `POINT Z`! Good catch - done :)

src/neo4j_graphrag/components/parquet_formatter.py
adamnsch: Good point (no pun intended 😄)! I've changed it so that we log a warning now if this happens, and serialize to JSON strings instead of just dumping the `str(dict)` we should hopefully look better :)

PR #626, opened 2026-09-13
refactor(llm): unify invoke/ainvoke dispatch across providers

matteomedioli:
# Description

Refactors `LLMBase.invoke`/`ainvoke` from an abstract dispatcher every subclass had to reimplement into a concrete dispatcher: it branches on input type (`str` vs `list[LLMMessage]`) and delegates to four narrow abstract hooks (`_invoke_v1`, `_invoke_v2`, `_ainvoke_v1`, `_ainvoke_v2`). All eight providers (`AnthropicLLM`, `OpenAILLM`, `GeminiLLM`, `CohereLLM`, `MistralAILLM`, `OllamaLLM`, `BedrockLLM`, `VertexAILLM`) migrated onto this contract, each now sharing one request builder and response parser per input version, plus a pair of `_call_sync`/`_call_async` transport hooks that differ only in whether the SDK call is awaited. This removes the duplicated sync/async dispatch and request/response logic that previously lived four times per provider.

Along the way, this fixes bugs the duplication had let drift out of sync:
- `BedrockLLM` now inherits from `LLMBase` (previously `LLMInterface`/`LLMInterfaceV2` directly), aligning its hierarchy with every other provider. Its async path no longer stacks a sync-side rate-limit retry inside the thread-pool-executed call on top of its own async retry layer.
- `OllamaLLM`'s async path now spreads `model_params` into the `client.chat` call the same way the sync path always did, instead of passing it verbatim as the `options` value — this fixed a double-nested `options` bug when `model_params` carried a sibling key (e.g. `{"options": {...}, "format": "json"}`).
- `CohereLLM`'s v2 error paths now raise `LLMGenerationError(e) from e` consistently, preserving the original SDK exception (previously a generic `"Error calling cohere"` message on v2 only).
- `BedrockLLM` and `GeminiLLM` now wrap request-building errors (message/schema validation, before the SDK call) into `LLMGenerationError` too, so callers get one consistent error type for the whole `invoke`/`ainvoke` call instead of transport errors being wrapped while build-time errors leak raw.

**Breaking (targets 2.0):** a subclass overriding `invoke`/`ainvoke` directly (a supported extension point for `BaseAnthropicLLM`/`BaseOpenAILLM`/`BaseGeminiLLM`) must migrate to overriding the four hooks instead. The public `invoke`/`ainvoke`/`invoke_with_tools`/`ainvoke_with_tools` calling contract for callers is unchanged.

Note: `GeminiLLM`'s usage-tracking fix (`response.usage_metadata` -> `LLMResponse.usage`) landed separately in #621 and is already on `main`; this branch is rebased on top of it, not re-introducing it.

## Type of Change
- [x] Refactoring (no functional changes) — plus incidental bug fixes surfaced by the consolidation (see above)

## Complexity

Complexity: Medium

## How Has This Been Tested?
- [x] Unit tests

# Checklist


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+- (**breaking**) `BedrockLLM` now inherits from `LLMBase` (previously `LLMInterface`/`LLMInterfaceV2` directly), aligning it with every other provider's class hierarchy. Its async path no longer stacks a sync-side rate-limit retry inside the thread-pool-executed call on top of its own async retry layer, so only the async layer applies. The public `invoke`/`ainvoke`/`invoke_with_tools`/`ainvoke_with_tools` contract is unchanged.
```
jonnylaw: The contract has changed - I think because we've fixed a bug on ainvoke_v2 to now pass the kwargs along.

PR #627, opened 2026-09-17
Add pipeline stage observer and example

jonnylaw:
# Description

The pipeline `StageObserver` can be used to track values going through the pipeline without adding logging, metric collection, tracing etc. directly to the transformation functions. This makes it easy to uniformly manage observability from a collection of `StageObserver` classes.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Complexity

Medium

Complexity:

## How Has This Been Tested?
- [x] Unit tests
- [ ] E2E tests
- [x] Manual tests

# Checklist


src/neo4j_graphrag/pipeline/observers.py
adamnsch, line 1: Just a thought: I wonder if it can be too spammy do log for every item? Could we add something so that it only accumulates and logs every n seconds? Maybe it's something for each implementation though - not the `StageObserver` as it's good if that's very general and as simple as possible i think

src/neo4j_graphrag/pipeline/operators.py
```diff
@@ -84,6 +84,11 @@ class Operator:
 
     prev: Operator | None
 
+    @property
+    def name(self) -> str:
+        """Human-readable stage name for logs and observers (the class name)."""
+        return type(self).__name__
```
adamnsch: Is this specific enough? What if we have multiple `map` in our pipeline and so on?

src/neo4j_graphrag/pipeline/observers.py
```diff
@@ -0,0 +1,123 @@
...
+        while True:
+            try:
+                item = next(iterator)
+            except StopIteration:
+                return
+            except Exception as e:
+                self.on_error(op, e)
+                raise
+            self.before(op, item)
+            yield item
```
adamnsch, line 144: Sinks don't yield anything right? Will they not get this logging then?

src/neo4j_graphrag/pipeline/observers.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __init__(self, log: logging.Logger | None = None) -> None:
+        self._log = log if log is not None else logger
+
+    def wrap(self, op: Operator, stream: Iterator[Any]) -> Iterator[Any]:
+        self._log.info("Stage %s: starting", op.name)
+        count = 0
+        for item in super().wrap(op, stream):
+            count += 1
+            yield item
```
adamnsch, line 181: It feels a bit scary that these observers could technically modify the `item`. We could consider something like:
```
from copy import copy

item = next(iterator)
self.before(op, copy(item))
yield item
self.after(op, copy(item))
```
but it kind of feels overkill. Maybe just documenting this would be good though?

adamnsch: We could also pass some kind of view/metadata of item instead, like `type(item).__name__` but it's limiting of course

src/neo4j_graphrag/pipeline/observers.py
```diff
@@ -0,0 +1,123 @@
...
+from neo4j_graphrag.pipeline.operators import Operator
+
+__all__ = ["StageObserver", "LoggingStageObserver"]
+
+logger = logging.getLogger(__name__)
+
+_T = TypeVar("_T")
+
+
+class StageObserver(ABC, Generic[_T]):
```
adamnsch: Nitpick: this has no abstract methods. Should eg. `before` and `after` be abstract?

src/neo4j_graphrag/pipeline/observers.py
```diff
@@ -0,0 +1,123 @@
...
+    def wrap(self, op: Operator, stream: Iterator[Any]) -> Iterator[Any]:
+        self._log.info("Stage %s: starting", op.name)
+        count = 0
+        for item in super().wrap(op, stream):
+            count += 1
+            yield item
+        self._log.info("Stage %s: finished, %d item(s)", op.name, count)
+
+    def before(self, op: Operator, item: Any) -> None:
+        self._log.debug("Stage %s: emitting %r", op.name, item)
```
adamnsch: I would what kind of implementations we have for `item.__str__`

src/neo4j_graphrag/pipeline/observers.py
jonnylaw: Per-item logging is `DEBUG` only — at the default `INFO` you get two lines per stage (start, and finish with the item count), so the spam is opt-in. I agree with your instinct to keep `StageObserver` general, so I have left accumulation to implementations: a throttling observer is a few lines on top of the hooks, whereas time-based buffering in the base class would make every observer harder to test deterministically.

What I did add is `max_repr` on `LoggingStageObserver` (see the thread on `before`), which bounds the size of each `DEBUG` line.

src/neo4j_graphrag/pipeline/operators.py
jonnylaw: `name` is now, in order of precedence:

1. the `label` you passed, if any;
2. otherwise `ClassName[position](func_name)`, e.g. `Map[1](embed)` / `Map[2](normalise)` / `Filter[3](is_even)`.

The position comes from walking the `prev` chain, so it is unique by construction and three `map` calls stay apart; the function name is dropped for lambdas, where it carries no information. And every operator method now takes `label=` for when you want something meaningful:

```python
Pipeline.from_source(src).map(embed, label="embed").filter(keep, label="dedupe")
```

`name` is a `cached_property`, since observers read it once per item per stage.

src/neo4j_graphrag/pipeline/observers.py
jonnylaw: Fixed by wrapping the sink's *input* stream instead of its (empty) output: `before`/`after` now bracket each `sink.write`, and the stage reports the number of items written. `evaluate()`'s contract is unchanged a sink-terminated pipeline still yields nothing.

That needed one small API addition to be reachable: `to_sink` has to drain its own stream, so it now accepts an optional `interpreter`:

```python
pipeline.to_sink(sink, label="write", interpreter=LocalInterpreter(observers=[observer]))
```

src/neo4j_graphrag/pipeline/observers.py
jonnylaw: Agreed on both counts. `copy()` is shallow, so it would guard against the naive `item.x = ...` while giving false confidence about anything nested, and it costs a copy per item per stage boundary in a DSL whose point is bounded memory. Passing only metadata would remove the hazard but kill the debugging use case the example notebook is built around.

So: documented instead. `StageObserver` now carries an `.. important::` note that hooks receive the live item, must treat it as read-only (mutating it changes what later stages see), and should not retain it past the hook — derive what you need and let the item go.

src/neo4j_graphrag/pipeline/observers.py
jonnylaw: You are right, the `ABC` was decorative — nothing stopped instantiation. Dropped it; `StageObserver` is now a plain `Generic[_T]` base with no-op hooks, which matches how observers are actually written (override only the hook you care about — a metrics observer often wants just `before`). Making `before`/`after` abstract would force two methods on every implementation.

src/neo4j_graphrag/pipeline/observers.py
jonnylaw: It's true - I just added this as an example really but we may as well make it more robust by trimming the output.

`LoggingStageObserver` now takes `max_repr` (default 200, `None` for the full `repr`). The item is handed to the logger wrapped in a small lazy object whose `__repr__` renders and clips, so nothing is built unless the `DEBUG` record is actually emitted — there is a test asserting an expensive `__repr__` is never called when `DEBUG` is off.

src/neo4j_graphrag/pipeline/observers.py
```diff
@@ -0,0 +1,179 @@
...
+        """Wrap *op*'s output stream, invoking the hooks around each item."""
+        iterator = iter(stream)
+        while True:
+            try:
+                item = next(iterator)
+            except StopIteration:
+                return
+            except Exception as e:
+                self.on_error(op, e)
+                raise
```
adamnsch: If there's an error in the wrapped `next()` here, do we still need to call this `self.on_error`? The observer itself has not failed in any way

src/neo4j_graphrag/pipeline/observers.py
adamnsch: Nice! But does this also mean that `sink.write(item)` is not in the wrapper, so `self.on_error` is not called if it raises?

adamnsch (approved):
LGTM 🚀 

Just have some questions for my understanding

src/neo4j_graphrag/pipeline/observers.py
jonnylaw: Yes - not ideal. I'll fix that one before merging!

jonnylaw: Exactly right, and it did — thanks, that was a regression. Wrapping the sink's input put `sink.write` downstream of the observer, so a failing write propagated out of the consumer while the wrapper generator sat suspended at its `yield`; `on_error` never ran (it fired before this change, when the sink's empty output was wrapped).

Fixed by moving the write *inside* the observed stream instead of on either side of it:

```python
def _write_through(stream, sink):
    for item in stream:
        sink.write(item)
        yield item
```

and, for the `SinkOp` case:

```python
stream = _drain(self._observe(op, _write_through(stream, op.sink)))
```

So the write now happens in the `next()` call the observer itself makes: items are reported by `before`/`after`, a raising `sink.write` surfaces inside `wrap` and reaches `on_error`, and `_drain` hides the passed-through items again so a sink pipeline still yields nothing to the caller. Write cost lands in the window before `before`, which is what the `StageObserver` docstring already promises ("the elapsed time since the previous item ... is the time `op` spent computing this item") — for a sink, computing *is* writing.

Regression test added:

```
with pytest.raises(RuntimeError, match="write failed: 2"):
    Pipeline([1, 2, 3]).to_sink(sink, label="capture",
                                interpreter=LocalInterpreter(observers=[observer]))

assert sink.received == [1]
assert events[:2] == [("before", 1), ("after", 1)]
assert events[2][0] == "error"   # and nothing for the failed item
```

src/neo4j_graphrag/pipeline/observers.py
jonnylaw: I realised from this comment that this only handles a fatal `Exception` for the whole pipeline - however it should handle partial failure where a stage returns `Err`. This meant I had to move the logic out of `wrap` and into the interpreter directly. This still separates observability from the pipeline logic at the expensive of a more complex interpreter.

